PR #32, opened 2015-11-13
Content-Disposition filename also gives a UTF-8 encoded Filename

meg1502:
the "filename*=UTF-8''" notation will be honored by modern browsers and ignored by older browsers, which will use the "filename" value as the fallback.

http://greenbytes.de/tech/tc2231/#attfnboth

### DIFF
--- a/src/main/java/sirius/web/http/Response.java
+++ b/src/main/java/sirius/web/http/Response.java
@@ -837,7 +837,7 @@ public class Response {
         addHeaderIfNotExists("Content-Disposition",
                              (download ? "attachment;" : "inline;") + "filename=\"" + name.replaceAll(
                                      "[^A-Za-z0-9\\-_\\.]",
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/web/http/Response.java
```diff
@@ -837,7 +837,7 @@ private void setContentDisposition(String name, boolean download) {
         addHeaderIfNotExists("Content-Disposition",
                              (download ? "attachment;" : "inline;") + "filename=\"" + name.replaceAll(
                                      "[^A-Za-z0-9\\-_\\.]",
-                                     "_") + "\"");
+                                     "_") + ";" + "filename*=UTF-8''" + Strings.urlEncode(name));
```
andyHa: kann man vereinfachen zu "; filename...."

PR #37, opened 2016-02-03
Change danger modal to work with Bootstrap 3

sbi-scireum:
Change danger modal to work with Bootstrap 3


sbi-scireum:
bildschirmfoto 2016-02-03 um 16 00 45

PR #65, opened 2016-02-29
Add adminonly option to singleselect template part

meg1502:


src/main/resources/default/view/parts/singleSelect.html
```diff
@@ -5,10 +5,10 @@
 *  Copyright by scireum GmbH
 *  http://www.scireum.de - info@scireum.de
 *@
-@args String span, String name, String labelKey, String helpKey, boolean optional
+@args String span, String name, String labelKey, String helpKey, boolean optional, boolean adminonly
```
andyHa: additionalClass as String using "admin-link" as parameter would be more re-usable

andyHa: please add the appropriate admin-link class to one of the scss files...

PR #75, opened 2016-03-08
Submit an editForm after successful file upload if page contains one

sbi-scireum:
Submit an editForm after successful file upload if page contains one

### DIFF
--- a/src/main/resources/default/view/wondergem/template-script.html
+++ b/src/main/resources/default/view/wondergem/template-script.html
@@ -289,7 +289,13 @@
                 }
                 if (responseJSON.refresh) {
                     setTimeout(function () {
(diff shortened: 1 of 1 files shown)


src/main/resources/default/view/wondergem/template-script.html
```diff
@@ -289,7 +289,13 @@
                 }
                 if (responseJSON.refresh) {
                     setTimeout(function () {
-                        window.location.reload()
+                        var editForm = $('#editForm');
+                        if (editForm.length) {
```
andyHa: > 0 ?

andyHa: &gt;

PR #81, opened 2016-04-04
jho/tableColorized

jahoa:
Additional class table-colorize 
Background color of first column indicates state; row title also has state color to make sure state color and link color match.

### DIFF
--- a/src/main/resources/default/assets/wondergem/stylesheets/tables.scss
+++ b/src/main/resources/default/assets/wondergem/stylesheets/tables.scss
@@ -87,7 +87,7 @@ tfoot {
       background-color: #b6b7bb;
     }
     &.success > td:first-child {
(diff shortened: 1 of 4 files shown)


src/main/resources/default/assets/wondergem/stylesheets/tables.scss
```diff
@@ -98,21 +81,60 @@ tfoot {
...
 
   tbody > tr {
+
+    // Set colors for first column
+    &.default > td:first-child {
+      background-color: #b6b7bb;
+    }
     &.success > td:first-child {
-      background-color: #5eb526;
+      background-color: $oxomi;
```
andyHa: nui des muss $green sein....in sirius-web gibts "oxomi" net

PR #126, opened 2016-10-10
aha/HorrorShow

andyHa:


pom.xml
```diff
@@ -9,21 +9,20 @@
...
             <groupId>com.scireum</groupId>
             <artifactId>sirius-kernel</artifactId>
-            <version>3.0.1</version>
+            <version>3.1</version>
         </dependency>
         <dependency>
             <groupId>com.scireum</groupId>
             <artifactId>sirius-kernel</artifactId>
-            <version>[3.0.1,)</version>
+            <version>[3.1,)</version>
```
sabieber: Please use a specific version here as this can cause tests to fail after a newer version of sirius-kernel was released

PR #133, opened 2016-11-04
Adds keyboard shortcuts to the pagination template part

sabieber:
This allows the user to browse through the pageinated view by pressing Shift+Left or Shift+Right.
The corresponding event handlers are only added if there is more or less content in the provided Page object.


src/main/resources/default/view/parts/pagination.html
```diff
@@ -34,6 +34,31 @@
...
+                        @if (page.hasLess()) {
+                            $(document).keyup(function (e) {
+                                if (e.shiftKey && (e.which === 37)) {
+                                    @if (Strings.isFilled(type)) {
+                                        var location = '@prefix/@baseURL?@page.createPrevPageQueryString()&pageType=@type';
+                                    } else {
+                                        var location = '@prefix/@baseURL?@page.createPrevPageQueryString()';
+                                    }
+                                    window.location.replace(location);
+                                }
```
andyHa: e.preventDefault() / bubbleblubb / false returnen / browser event handling scheisse machen, damit das event nicht weitergeht?

PR #170, opened 2017-02-20
Add Paramter to template part deleteLink, to make it an admin-link

welzfabian:


pom.xml
```diff
@@ -9,7 +9,7 @@
         <version>3.8.3</version>
     </parent>
     <artifactId>sirius-web</artifactId>
-    <version>6.6</version>
+    <version>6.6.1</version>
```
andyHa: 6.7 not a bug fix

PR #176, opened 2017-03-02
Modifies the UserContextHelper so that one can see the actual messages if an assertion fails

D3nnisH:
image
... but which is the content of the damn messages that screwed up my test


pom.xml
```diff
@@ -9,7 +9,7 @@
         <version>3.8.3</version>
     </parent>
     <artifactId>sirius-web</artifactId>
-    <version>6.8.2</version>
+    <version>6.8.3</version>
```
sabieber: Please upgrade it to 6.8.4

src/test/java/sirius/web/security/UserContextHelper.groovy
```diff
@@ -6,67 +6,57 @@
...
      */
-    public static boolean expectNoErrorMessages(TestResponse response) {
-        for (Message msg : getUserContext(response).getMessages()) {
-            if (msg.getType() == Message.ERROR) {
-                return false;
-            }
-        }
-        return true;
+    public static void expectNoErrorMessages(TestResponse response) {
+        assert !getUserContext(response).getMessages().any { msg -> msg.getType() == Message.ERROR }
```
sabieber: I think `getUserContext(response).getMessages().all { msg -> msg.getType() != Message.ERROR }` would be easier to read

PR #183, opened 2017-03-17
Adds a fallback for composites based on the ScopeInfo directly instead of the inner scope object

andyHa:
Adds a fallback for composites based on the ScopeInfo directly instead of the inner scope object

This is required by SellSite which attaches some components to the ScopeInfo, which is
not changed when the inner scope object (Shop) is refreshed


src/main/java/sirius/web/security/ScopeInfo.java
```diff
@@ -172,8 +172,8 @@ public String getLang() {
     @Override
     public boolean is(@Nonnull Class<?> type) {
         Transformable userObject = getScopeObject(Transformable.class);
-        if (userObject != null) {
-            return userObject.is(type);
+        if (userObject != null&& userObject.is(type)) {
```
yzzo: codeformat

PR #189, opened 2017-03-28
Adds support for multiselect to facets

Tobsucht:


src/main/java/sirius/web/controller/Facet.java
```diff
@@ -151,7 +158,19 @@ public boolean hasItems() {
      * @return the facet itself for fluent method calls
      */
     public Facet withValue(String value) {
-        this.value = value;
+        this.values.add(value);
```
D3nnisH: shouldn't the value list be cleared before?

Tobsucht: right ... :o

PR #198, opened 2017-04-11
Improves the SmartLineBasedProcessor to support magic import

D3nnisH:
* makes it possible to process multiple columns with the same name
* adds a method for returning the computed column-mapping


src/main/java/sirius/web/templates/NamedRowProcessor.java
```diff
@@ -28,5 +27,5 @@
      * @param lineNumber the line number which is currently being processed.
      * @param row        the data of the row / line to process
      */
-    void handleRow(int lineNumber, Map<String, Value> row);
+    void handleRow(int lineNumber, ListMultimap<String, Value> row);
```
andyHa: wouldn't it even be sexier to pass a "SmartRow" (or the like) interface here, which has getValue(String) -> Value and getValues(String) -> List<Value> to hide the implementation details.

This interface could be implemented in-place in SmartLineBaseProcessor

D3nnisH: `ListMultimap` is already an interface. And I want to use Guava's MultiMap, which I cannot if I use an own interface

andyHa: but retrieving a single value for a column requires a complex lookup like  l = get(X); if (!l.isEmpty()) { val = l.get(0); }else{val = Value.EMPTY; }

which could be hidden in a getValue method. the 2nd getValues simply forwards to map.get ...

PR #200, opened 2017-04-13
Fixes a translation for select2

D3nnisH:
The text for "The results could not be loaded" wasn't translated

merge before #198


src/main/resources/web_de.properties
```diff
@@ -31,6 +31,7 @@ template.html.select2.selection=Ihre Auswahl...
 template.html.select2.loading=Lade Daten...
 template.html.select2.searching=Suche...
 template.html.select2.noMatches=Keine Suchtreffer...
+template.html.select2.errorLoading=Die Suche ist fehlgeschlagen.
```
meg1502: DE text doesnt match EN text

src/main/resources/web_en.properties
```diff
@@ -74,6 +74,7 @@ template.html.placeholder=Enter a search text...
 template.html.select2.loading=Loading data...
 template.html.select2.maximumSelected.prefix=You can maximum
 template.html.select2.maximumSelected.suffix=Select elements
```
meg1502: this prefix / suffix translation doesnt make much sense

meg1502: 'you can select a maximum of' + 'elements'

PR #201, opened 2017-04-18
Refactoring to sirius-kernel 6.0

andyHa:


src/main/java/sirius/web/security/ScopeInfo.java
```diff
@@ -65,7 +66,7 @@
     private Function<ScopeInfo, Object> scopeSupplier;
     private Map<Class<?>, Object> helpersByType = Maps.newConcurrentMap();
     private Map<String, Object> helpersByName = Maps.newConcurrentMap();
-    private Config config;
+    private ExtendedSettings config;
```
sabieber: Could we rename it to settings everywhere?

PR #208, opened 2017-05-09
Modified CheckBoxTemplate

PhilHup:
In case you want to display raw label or help Strings and not have them translated there are two new parameters in the checkbox template.
Bumps Versionnumber.


src/main/resources/default/view/parts/checkbox.html
```diff
@@ -1,11 +1,21 @@
...
+@args String span, String name, boolean value, String labelKey, String helpKey, boolean readonly, String label, String help
 <div class="col-md-@(span == null || span == "" ? 6 : span)">
     <div class="form-group checkbox">
         <label>
             <input type="checkbox" name="@name" value="true" @if (value) { checked="checked" } @if (readonly) { disabled="disabled" } />
-            @i18n(labelKey)
+            @if (Strings.isFilled(labelKey)) {
+                @i18n(labelKey)
+            } else {
+                @if (Strings.isFilled(label)) {
```
sabieber: ich glaube die if's kann man sich sparen. vgl textfield.html

PR #215, opened 2017-05-24
User agent detection

fkrone:
From now on it is possible to determine if the current request was created by a deskop client or a mobile client. It is possible to distinguish between phones and tablets for mobile requests. Furthermore it is also possible to determine if a request was made by an Android device or an iOS device.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -9,7 +9,7 @@
         <version>3.8.3</version>
     </parent>
     <artifactId>sirius-web</artifactId>
-    <version>8.6.2</version>
+    <version>8.7</version>
     <name>SIRIUS web</name>
     <description>Provides a modern and scalable web server as SIRIUS module</description>
 
--- a/pom.xml
+++ b/pom.xml
@@ -9,7 +9,7 @@
         <version>3.8.3</version>
     </parent>
     <artifactId>sirius-web</artifactId>
-    <version>8.7</version>
+    <version>8.8</version>
     <name>SIRIUS web</name>
     <description>Provides a modern and scalable web server as SIRIUS module</description>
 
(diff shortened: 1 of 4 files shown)


src/main/java/sirius/web/http/WebContext.java
```diff
@@ -1873,6 +1879,19 @@ public String getDynamicAssetToken() {
         return Product.getProduct().getUniqueVersionString();
     }
 
+    /**
+     * Returns {@link UserAgent} for easy access to the user agent used for this request. Also it provides access to
+     * some assumtions based on the user agent e.g. which device was used.
```
meg1502: typo (assumptions)

src/main/java/sirius/web/http/session/UserAgent.java
```diff
@@ -0,0 +1,142 @@
...
+    private boolean android = false;
+    private boolean iOS = false;
+
+    public UserAgent(String userAgentString) {
+        this.userAgentString = userAgentString;
+    }
+
+    protected void parseUserAgent() {
+        userAgentParsed = true;
+        if (Strings.isEmpty(userAgentString)) {
```
andyHa: why not call parseUserAgent in the constructor....it will be called eventually....

pom.xml
```diff
@@ -9,7 +9,7 @@
         <version>3.8.3</version>
     </parent>
     <artifactId>sirius-web</artifactId>
-    <version>8.6.4</version>
+    <version>8.7</version>
```
D3nnisH: 8.7 hat Welzi jetz schon in Beschlag genommen mit #219

src/main/java/sirius/web/http/WebContext.java
```diff
@@ -1873,6 +1879,19 @@ public String getDynamicAssetToken() {
...
 
+    /**
+     * Returns {@link UserAgent} for easy access to the user agent used for this request. Also it provides access to
+     * some assumptions based on the user agent e.g. which device was used.
+     *
+     * @return user agent wrapper object
+     */
+    public UserAgent getUserAgent() {
+        if (userAgent == null) {
+            userAgent = new UserAgent(getHeader("User-Agent"));
```
andyHa: HttpHeaders.Names.USER_AGENT o.ä.

PR #220, opened 2017-06-07
Attaches session even if some permission is missing

tbiScireum:
As long as the user is logged in.

Up to know it was not possible to call a route without the necessary permission, log in and save the login information into the session.
As soon as the user left the 403 Page (or reloaded) the login was lost.

### DIFF
--- a/src/main/java/sirius/web/controller/ControllerDispatcher.java
+++ b/src/main/java/sirius/web/controller/ControllerDispatcher.java
@@ -155,6 +155,12 @@ public class ControllerDispatcher implements WebDispatcher {
                 return;
             }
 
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/web/controller/ControllerDispatcher.java
```diff
@@ -155,6 +155,12 @@ private void performRouteInOwnThread(WebContext ctx, Route route, List<Object> p
                 return;
             }
 
+            // If a user authenticated during this call...bind to session!
+            UserContext userCtx = UserContext.get();
```
andyHa: why not user.isLoggedIn?

PR #254, opened 2017-07-14
Adds a facet template which takes care of the sizing if facets are pr…

tbiScireum:
…esent or not


src/main/resources/default/taglib/w/facet.html.pasta
```diff
@@ -0,0 +1,19 @@
+<i:arg type="Page" name="page" />
```
andyHa: facetSidebar oder so?

src/main/resources/default/taglib/w/facet.html.pasta
```diff
@@ -0,0 +1,19 @@
...
+
+<div class="row">
+    <i:if test="page.hasFacets()">
+        <div class="col-md-3">
+            <w:filterbox page="@page" baseURL="@baseURL" />
+        </div>
+    </i:if>
+
+
+    <div class="@if(page.hasFacets()) { col-md-9 } else { col-md-12 }">
```
andyHa: if_(

PR #257, opened 2017-07-18
Adds a page parameter to the deleteLink

tbiScireum:
Default is null and this will not change the behavior
If page is set the query string containing the active facets is appended to the delete url
Fixes also a test of a ported template

### DIFF
--- a/src/main/resources/default/taglib/w/deleteLink.html.pasta
+++ b/src/main/resources/default/taglib/w/deleteLink.html.pasta
@@ -1,7 +1,12 @@
 <i:arg type="String" name="url" />
 <i:arg type="boolean" name="adminOnly" default="false"/>
+<i:arg type="Page" name="page" default=""/>
 
 <i:pragma name="inline" value="true" />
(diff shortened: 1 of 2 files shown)


src/main/resources/default/taglib/w/deleteLink.html.pasta
```diff
@@ -1,7 +1,12 @@
...
 <i:arg type="boolean" name="adminOnly" default="false"/>
+<i:arg type="Page" name="page" default=""/>
 
 <i:pragma name="inline" value="true" />
 <i:pragma name="description" value="Renders a delete/dangerous link within a Wondergem template" />
 
-<i class="fa fa-trash"></i> <a class="link link-danger @if (adminOnly) {admin-link}" href="@url">@i18n("NLS.delete")</a>
+<i class="fa fa-trash"></i>
+<a class="link link-danger @if (adminOnly) {admin-link}"
+   href="@if(page != null) {@url?@page.createQueryString()} else {@url}">
```
andyHa: @url before if

PR #258, opened 2017-07-18
Give Api-User more controll over when to do Page-Jumps with Pagination

welzfabian:
By checking the result of "shouldDoPageJump"-function if it exists

### DIFF
--- a/src/main/resources/default/view/parts/pagination.html
+++ b/src/main/resources/default/view/parts/pagination.html
@@ -24,6 +24,11 @@
                 <script type="text/javascript">
                     $(document).ready(function() {
                         $("input:text.jump-to").keyup(function (e) {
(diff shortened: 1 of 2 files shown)


src/main/resources/default/view/parts/pagination.html
```diff
@@ -24,6 +24,11 @@
                 <script type="text/javascript">
                     $(document).ready(function() {
                         $("input:text.jump-to").keyup(function (e) {
+                            // api-users can define a shouldDoPageJump method, where they can controll whether to
```
meg1502: typo 'control*l*'

src/main/resources/default/view/parts/pagination.html
```diff
@@ -24,6 +24,11 @@
                 <script type="text/javascript">
```
andyHa: pastaaaaaaaa

PR #261, opened 2017-07-19
Adds the possibility to pass an title-string to the pageHeader part.

PhilHup:
Adds the possibility to pass an title-string to the pageHeader part.


src/main/resources/default/taglib/w/pageHeader.html.pasta
```diff
@@ -1,9 +1,10 @@
 <i:arg type="String" name="titleKey" default="" />
+<i:arg type="String" name="title" default="@i18n(titleKey)" />
 <i:pragma name="inline" value="true" />
 <i:pragma name="description" value="Provides a header for a Wondergem page"/>
 
 <i:if test="isFilled(titleKey)">
```
fkrone: Check here for title as well

PR #272, opened 2017-07-24
Merges some mobile taglib templates into the wondergem templates

tbiScireum:


src/main/resources/default/taglib/w/autocompleteSingleSelect.html.pasta
```diff
@@ -26,7 +27,7 @@
         </label>
     </i:if>
 
-    <select name="@name" class="form-control @user.signalFieldError(name)"
+    <select name="@name" id="@id" class="form-control @user.signalFieldError(name)"
```
andyHa: @if (isFilled(id) {} drumrum?

PR #273, opened 2017-07-24
Improves security checks.

andyHa:
Tagliatelle now happily compiles and invokes all .pasta files
as they're obviously meant to be templates.

However, for includes, only /assets is accepted, as this is the
directory(tree) to put files for inclusions anyway..

### DIFF
--- a/src/main/java/sirius/tagliatelle/Tagliatelle.java
+++ b/src/main/java/sirius/tagliatelle/Tagliatelle.java
@@ -258,7 +258,7 @@ public class Tagliatelle {
     }
 
     /**
-     * For security reasons we only include or invoke templates and resources in either /assets or /templates.
+     * For security reasons we only invoke templates that end with .pasta.
      * <p>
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/tagliatelle/Tagliatelle.java
```diff
@@ -258,20 +258,16 @@ private static boolean checkAutoboxing(Class<?> primitveType, Class<?> boxedType
...
      * @param path the path to check
      */
     public static void ensureProperTemplatePath(String path) {
-        String effectiveUri = path.startsWith("/") ? path : "/" + path;
-
-        if (!effectiveUri.startsWith("/templates") && !effectiveUri.startsWith("/assets") && !effectiveUri.startsWith(
-                "/taglib")) {
-            throw new IllegalArgumentException(
-                    "Tagliatelle templates must reside in /templates, /taglib or /assets. Invalid path: " + path);
+        if (path.endsWith(".pasta")) {
```
yzzo: andersrum?

PR #278, opened 2017-07-25
Implements some mail macros

tbiScireum:


src/main/java/sirius/tagliatelle/macros/EscapeXmlMacro.java
```diff
@@ -0,0 +1,76 @@
...
+
+import javax.annotation.Nonnull;
+import java.text.StringCharacterIterator;
+import java.util.List;
+
+/**
+ * Provides a xml escaping method.
+ */
+@Register
+public class EscapeXmlMacro implements Macro {
```
andyHa: mach tagliadelle audomadisch

PR #298, opened 2017-08-16
Revives i18n support for generated mails

sabieber:


src/main/java/sirius/web/mails/MailSender.java
```diff
@@ -317,6 +337,9 @@ public void send() {
         String tmpLang = NLS.getCurrentLang();
         try {
             try {
+                if (lang != null) {
+                    CallContext.getCurrent().setLang(lang);
```
andyHa: reset lang afterwards, plz

PR #299, opened 2017-08-21
Fix if-handler for static content

Tobsucht:
When using @if(){...} containing e.g. a javascript expression or function, the compiler couldn't find the correct ending
of the @if-block, as the first found closing "}" is used to detect the end of the if-block.

Example code:
@if (...) {
    if () {
        ...;
    }
}

### DIFF
--- a/src/main/java/sirius/tagliatelle/compiler/Compiler.java
+++ b/src/main/java/sirius/tagliatelle/compiler/Compiler.java
@@ -518,12 +518,20 @@ public class Compiler extends InputProcessor {
      */
     private String consumeStaticBlock() {
         StringBuilder sb = new StringBuilder();
+        Integer numberOfOpenBlocks = 0;
+
         while (!reader.current().isEndOfInput()) {
             if (isAtEscapedAt()) {
                 sb.append(reader.consume().getValue());
                 reader.consume().getValue();
             } else {
-                if (isAtPotentialEndOfStaticBlock()) {
+                if (reader.current().is('{')) {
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/tagliatelle/compiler/Compiler.java
```diff
@@ -518,12 +518,20 @@ public Expression parseExpression(boolean skipWhitespaces) {
      */
     private String consumeStaticBlock() {
         StringBuilder sb = new StringBuilder();
+        Integer numberOfOpenBlocks = 1;
```
D3nnisH: use `int` instead of `Integer` to prevent autoboxing/unboxing at `++`, `--` and `== 0`

src/main/java/sirius/tagliatelle/compiler/Compiler.java
```diff
@@ -518,12 +518,20 @@ public Expression parseExpression(boolean skipWhitespaces) {
...
                 reader.consume().getValue();
             } else {
-                if (isAtPotentialEndOfStaticBlock()) {
+                if (reader.current().is('{')) {
+                    numberOfOpenBlocks++;
+                } else if (reader.current().is('}')) {
+                    numberOfOpenBlocks--;
+                }
+
+                if (isAtPotentialEndOfStaticBlock(numberOfOpenBlocks)) {
```
andyHa: soll mer das net in zeile 532 packen ala if (numberOfOpenBlock == 0) { return sb.toString()) }

PR #307, opened 2017-09-11
Handling of Amount in ExcelExport

fkrone:
Amount can now be processed by ExcelExport and is handled as numeric. Futhermore, BigDecimal is also handled as numeric from now on.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -9,7 +9,7 @@
         <version>3.8.3</version>
     </parent>
     <artifactId>sirius-web</artifactId>
-    <version>9.5.1</version>
+    <version>9.6</version>
     <name>SIRIUS web</name>
     <description>Provides a modern and scalable web server as SIRIUS module</description>
 
--- a/pom.xml
+++ b/pom.xml
@@ -9,7 +9,7 @@
         <version>3.8.3</version>
     </parent>
     <artifactId>sirius-web</artifactId>
-    <version>9.5</version>
+    <version>9.5.1</version>
     <name>SIRIUS web</name>
     <description>Provides a modern and scalable web server as SIRIUS module</description>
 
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -9,7 +9,7 @@
         <version>3.8.3</version>
     </parent>
     <artifactId>sirius-web</artifactId>
-    <version>9.5</version>
+    <version>9.5.1</version>
```
meg1502: Shouldnt this be 9.6 ?

src/main/java/sirius/web/data/ExcelExport.java
```diff
@@ -303,7 +308,11 @@ private HSSFCellStyle getCellStyleForObject(Object data) {
         HSSFCellStyle style = normalStyle;
         if (data instanceof LocalDate || data instanceof LocalDateTime) {
             style = dateStyle;
```
meg1502: return datestyle directly

src/main/java/sirius/web/data/ExcelExport.java
```diff
@@ -303,7 +308,11 @@ private HSSFCellStyle getCellStyleForObject(Object data) {
...
-        } else if (data instanceof Integer || data instanceof Double || data instanceof Long) {
+        } else if (data instanceof Integer
+                   || data instanceof Double
+                   || data instanceof Long
+                   || data instanceof BigDecimal
+                   || (data instanceof Amount && ((Amount) data).isFilled())) {
             style = numeric;
```
meg1502: return numericStyle directly

src/main/java/sirius/web/data/ExcelExport.java
```diff
@@ -303,7 +308,11 @@ private HSSFCellStyle getCellStyleForObject(Object data) {
...
         }
         return style;
```
meg1502: return normalStyle directly

PR #321, opened 2017-09-21
Provides easy way to add the ace editor.

tbiScireum:
Also fixes a bug where the searchfield color and background-color are the same.


src/main/resources/default/templates/system/scope-config.html.pasta
```diff
@@ -32,12 +32,7 @@
             </w:navbox>
         </div>
         <div class="col-md-9">
-            <div id="editor" style="height: 450px">@contents</div>
-            <script type="text/javascript">
-                var editor = ace.edit("editor");
-                editor.setTheme("ace/theme/idle_fingers");
-                editor.setReadOnly(true);
-            </script>
+            <w:codeEditor mode="json">@contents</w:codeEditor>
```
fkrone: Set readonly

PR #326, opened 2017-10-04
POST for tests with InputStream and formData

FrizRobin:
Adds an additional method for controller tests. This POST method takes both an InputStream and formData so that files could be uploaded while providing additional parameters (e.g. filename).

### DIFF
--- a/src/test/java/sirius/web/http/TestRequest.java
+++ b/src/test/java/sirius/web/http/TestRequest.java
@@ -31,6 +31,7 @@ import io.netty.handler.codec.http.multipart.MemoryAttribute;
 import sirius.kernel.async.Barrier;
 import sirius.kernel.async.CallContext;
 import sirius.kernel.async.Promise;
+import sirius.kernel.commons.Context;
 import sirius.kernel.commons.Strings;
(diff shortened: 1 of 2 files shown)


src/test/java/sirius/web/http/TestRequest.java
```diff
@@ -228,6 +229,25 @@ public static TestRequest POST(String uri, JSONObject json) {
...
+     * Creates a mock request simulating a POST on the given uri and form data while sending the given data.
+     *
+     * @param uri      the relative uri to call
+     * @param resource resource the data to send to the server
+     * @param formData the parameters be included in the post request
+     * @return an instance used to further specify the request to send
+     */
+    public static TestRequest POST(String uri, InputStream resource, Context formData) {
+        TestRequest result = POST(uri, formData);
+        try {
```
andyHa: duplicate code. irgendein grund das nicht in installContent zu schieben?

andyHa: außerdem - wie soll das gehen? man kann kein form UND content posten - dafür braucht man einen multipart form

PR #331, opened 2017-10-18
Makes WebContext.getContent() reusable

tbiScireum:
The bug occured by calling isContentProbablyXML() followed by getContent().
The first method call would change the readerindex of the byteBuf object and getContent would always return an empty byte array


src/main/java/sirius/web/http/WebContext.java
```diff
@@ -1621,7 +1594,7 @@ public InputStream getContent() throws IOException {
         }
         //Backup the original size...
         contentSize = (long) content.getByteBuf().readableBytes();
```
andyHa: dann kann man den hack killen - den brauchts ja dann nimmer

PR #385, opened 2018-01-24
Autocomplete Update

tbiScireum:
Bugfixes and Documentation
- Polyfills for better IE support
- adds a spinner option which indicates the waiting for response status of the autocomplete service
- adds multiple autocomplete instances per page option (including a shared cache)
- fixes a bug related to the autocomplete.hide method
- moves sirius.keys into own template file

### DIFF
--- a/src/main/resources/default/templates/wondergem/scripts/autocomplete.html.pasta
+++ b/src/main/resources/default/templates/wondergem/scripts/autocomplete.html.pasta
@@ -1,34 +1,11 @@
 <script type="text/javascript">
     window.sirius = window.sirius || {};
 
-    sirius.keys = (function(){
-        // fix physical keyboard bug on iPad that returns keyCode 0 on keyup
-        var map = {
-            "Backspace": 8,
-            "Tab": 9,
-            "Enter": 13,
-            "Control": 17,
-            "Escape": 27,
-            " ": 32,
-            "ArrowUp": 38,
-            "ArrowDown": 40,
-            ",": 188
-        };
+    function asFunction(possibleFunction) {
+        return typeof possibleFunction === "function" ? possibleFunction : Function();
+    }
 
-        return {
-            KEY_ENTER: 13,
-            KEY_ARROW_UP: 38,
-            KEY_ARROW_DOWN: 40,
-
-            enableKeyUpIpadSupport: function(event){
-                if (event.keyCode === 0 && event.key) {
-                    event.keyCode = event.which = map[event.key];
-                }
-            }
-        }
-    }());
-
-    sirius.autocomplete = (function () {
+    sirius.createAutocomplete = (function () {
         var anchor = {
             name: "autocomplete-anchor",
             element: undefined,
@@ -154,12 +131,12 @@
             VISIBLE: 1,
             state: undefined,
 
-            init: function () {
-                var wrapper = "<div id='autocomplete-wrapper' class='well autocomplete-wrapper'></div>";
+            init: function (config) {
+                var wrapper = "<div class='well autocomplete-wrapper autocomplete-wrapper-js' {{#id}}id='{{id}}'{{/id}}></div>";
 
                 this.state = completions.HIDDEN;
 
-                this.element = $(wrapper);
+                this.element = $(Mustache.render(wrapper, config));
                 this.element.width(anchor.element.outerWidth());
                 this.element.offset({
                     top: anchor.element.offset().top + anchor.element.height(),
@@ -223,7 +200,6 @@
                 if (completions.state === completions.HIDDEN) {
                     return;
                 }
-                completions.state = completions.HIDDEN;
 
                 window.setTimeout(function () {
                     /*
@@ -275,7 +251,7 @@
             update: function () {
                 this.element.empty();
 
-                if (this.rows.length < 1){
+                if (this.rows.length < 1) {
                     completions.hideImmediate();
                     return;
                 }
@@ -311,7 +287,10 @@
                     return;
                 }
 
-                events.onSelect(selectedElement);
+                if (events.onSelect(selectedElement) !== true) {
+                    completions.hideImmediate();
+                }
+
             }
         };
 
@@ -404,10 +383,6 @@
             }
         };
 
-        function asFunction(possibleFunction) {
-            return typeof possibleFunction === "function" ? possibleFunction : Function();
-        }
-
         return {
             eventNames: function () {
                 return Object.keys(events);
@@ -421,6 +396,13 @@
                 }
             },
 
+            config: undefined,
+            init: function (config) {
+                if (config) {
+                    this.config = config;
+                }
+            },
+
             /**
              * Method to initialize and start the autocompleter.
              *
@@ -431,16 +413,17 @@
                     throw "The event afterLoad must be set.";
                 }
 
+                this.init(config);
 
-                service.init(config.service);
+                service.init(this.config.service);
 
-                input.name = config.inputField;
+                input.name = this.config.inputField;
 
                 input.init();
-                anchor.init(config.anchor);
-                completions.init();
+                anchor.init(this.config.anchor);
+                completions.init(this.config.completions || {});
 
-                completions.setTemplates(config.templates);
+                completions.setTemplates(this.config.templates);
 
                 input.element.focus();
                 events.onReady();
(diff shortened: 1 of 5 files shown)


src/main/resources/default/templates/wondergem/scripts/autocomplete.html.pasta
```diff
@@ -46,16 +30,43 @@
...
+         * Represents the input field of the autocomplete.
+         *
+         * Events on the input field like onChange are defined here.
+         *
+         * @@type {{name: string, element: undefined, spinnerElement: undefined, spinnerTemplate: string, showSpinner: showSpinner, hideSpinner: hideSpinner, onChange: onChange, onNonInputKeys: onNonInputKeys, init: init}}
+         */
         var input = {
             name: "autocomplete-input",
             element: undefined,
+            spinnerElement: undefined,
```
sabieber: Do you really want to set these to undefined? I feel like it would be more correct to set them to null

src/main/resources/default/templates/wondergem/scripts/autocomplete.html.pasta
```diff
@@ -46,16 +30,43 @@
...
+            spinnerTemplate: '<span class="autocomplete-spinner hide"><i class="fa fa-spinner fa-spin"></i></span>',
+
+            /**
+             * Shows a small spinner in the input field whenever the service loads data.
+             */
+            showSpinner: function () {
+                if (input.spinnerElement) {
```
sabieber: Would be a lot safer to check if it is a jQuery object (will actually have method removeClass):
`input.spinnerElement instanceof jQuery`

src/main/resources/default/templates/wondergem/scripts/autocomplete.html.pasta
```diff
@@ -46,16 +30,43 @@
...
+                }
+
+            },
 
             /**
-             * This method gets called each time the user input is changed
+             * Hides the spinner.
+             */
+            hideSpinner: function () {
+                if (input.spinnerElement) {
```
sabieber: same as above

src/main/resources/default/templates/wondergem/scripts/autocomplete.html.pasta
```diff
@@ -90,26 +105,49 @@
                     return true;
                 }
 
-                if (event.keyCode === sirius.keys.KEY_ARROW_UP && completions.selectedRow !== undefined) {
-                    var prevRow = completions.selectedRow.prevAll(".autocomplete-row-js").first();
+                if (event.keyCode === sirius.keys.KEY_ARROW_UP) {
+                    if (completions.selectedRow !== undefined) {
```
sabieber: Do you need to exclude "null" as a value here? if not you could write `if (completions.selectedRow) {` or add a simple helper function like `function exists(object) { return typeof object !== 'undefined' && object !== null; }`

src/main/resources/default/templates/wondergem/scripts/autocomplete.html.pasta
```diff
@@ -90,26 +105,49 @@
...
+            init: function (config) {
                 this.element = $("#" + this.name);
 
+                if (config && !config.noSpinner) {
+                    input.spinnerElement = $(input.spinnerTemplate);
+                    input.spinnerElement.insertAfter(input.element);
+
+                    // this is needed to display the spinner in the same row as the input field
+                    // the spinner holds the class .fa which defines the size to 14
+                    input.element.css("max-width", input.element.width() - 14);
```
sabieber: Would like it a lot more if this was solved with a css class

src/main/resources/default/templates/wondergem/scripts/autocomplete.html.pasta
```diff
@@ -152,19 +195,22 @@
 
             HIDDEN: 0,
             VISIBLE: 1,
+            SOON_TO_BE_HIDDEN: 2,
```
sabieber: `HIDING`?

src/main/resources/default/templates/wondergem/scripts/autocomplete.html.pasta
```diff
@@ -175,6 +221,19 @@
                 $("body").append(this.element);
             },
 
+            rePosition: function () {
```
sabieber: reposition is a single word so the camelcase seems strange

src/main/resources/default/templates/wondergem/scripts/autocomplete.html.pasta
```diff
@@ -197,8 +256,10 @@
             },
 
             unHoverRow: function ($row) {
```
sabieber: unhover is a single word so the camelcase seems strange

src/main/resources/default/templates/wondergem/scripts/autocomplete.html.pasta
```diff
@@ -197,8 +256,10 @@
...
+                if ($row) {
```
sabieber: jQuery instance check?

src/main/resources/default/templates/wondergem/scripts/autocomplete.html.pasta
```diff
@@ -311,7 +379,10 @@
                     return;
                 }
 
-                events.onSelect(selectedElement);
+                if (events.onSelect(selectedElement) !== true) {
```
sabieber: just `if (!events.onSelect(selectedElement)) {` should be fine

src/main/resources/default/templates/wondergem/scripts/autocomplete.html.pasta
```diff
@@ -353,10 +427,27 @@
...
+                input.showSpinner();
+
+                setTimeout(function () {
+                    if (service.lastLoadTriggered + service.delay <= Date.now()) {
+                        service.loadImmediate(value);
+                    }
+                }, service.delay)
+            },
+
+            loadImmediate: function (value) {
```
sabieber: `loadImmediately`?

src/main/resources/default/templates/wondergem/scripts/autocomplete.html.pasta
```diff
@@ -466,7 +583,17 @@
 
             focus: function () {
                 input.element.focus();
+            },
+
+            isVisible: function () {
+                return completions.isVisible();
+            },
+
+            rePosition: function(){
```
sabieber: reposition is a single word so the camelcase seems strange

src/main/resources/default/templates/wondergem/scripts/event-keys.html.pasta
```diff
@@ -0,0 +1,135 @@
+<script type="text/javascript">
+    window.sirius = window.sirius || {};
+
+    function asFunction(possibleFunction) {
```
sabieber: Same tipp as above

src/main/resources/default/templates/wondergem/scripts/event-keys.html.pasta
```diff
@@ -0,0 +1,135 @@
...
+            KEY_SEMICOLON: 186,
+            KEY_COMMA: 188,
+
+            eventNames: function () {
+                return Object.keys(events);
+            },
+
+            on: function (name, callback) {
+                if (this.eventNames().includes(name)) {
+                    events[name] = asFunction(callback);
```
sabieber: Wouldn't it be better to check if callback is a function and do nothing if not?

src/main/resources/default/templates/wondergem/scripts/event-keys.html.pasta
```diff
@@ -0,0 +1,135 @@
...
+            },
+
+            enableInternetExplorer: function (e) {
+                return e || event;
+            },
+
+            /**
+             * Helper method which will call all browser & event specific helper functions.
+             */
+            browserFixes: function (e) {
```
sabieber: `applyBrowserFixes`?

src/main/resources/default/templates/wondergem/scripts/event-keys.html.pasta
```diff
@@ -0,0 +1,135 @@
...
+            ",": 188
+        };
+
+        /**
+         * Holds information about all pressed keys.
+         *
+         * If a key is pressed pressedKeys[keyCode] is set to true, else false.
+         * If the key was never pressed at all pressedKeys[keyCode] is undefined
+         */
+        var pressedKeys = {};
```
sabieber: Is the false state used in any way or would it be better to just use an array?

src/main/resources/default/templates/wondergem/scripts/tokenfield.html.pasta
```diff
@@ -1,6 +1,18 @@
 <script>
     window.sirius = window.sirius || {};
 
+    function asFunction(possibleFunction) {
```
sabieber: Same tipp as above

PR #391, opened 2018-01-26
aha/LoopyLoops

andyHa:


D3nnisH:
could we add a method `@state.getN()` or alike to get `currentRow`?

PR #404, opened 2018-02-07
Fixes a bug where the click event is not triggered

tbiScireum:
The bug happend only on some mobile devices because clicking on a completions element would
trigger a scroll event of the page before the click event was triggered.
The following click event would hold the completions div as target instead of the selected row.


src/main/resources/default/templates/wondergem/scripts/autocomplete.html.pasta
```diff
@@ -213,10 +211,9 @@
                 this.element = $(Mustache.render(wrapper, config));
                 this.rePosition();
 
-                this.element.click(function (element) {
-                    completions.select($(element.target));
 
-                    completions.hideImmediate();
+                this.element.on("mouseup.sirius-autocomplete", ".autocomplete-row-js", function () {
```
sabieber: `this.element.find('.autocomplete-row-js').on('mouseup.sirius-autocomplete', function () {`

PR #410, opened 2018-02-14
Creates a helper method to find out if a TestResponse contains response data

D3nnisH:
Depending on the ResponseType, different fields needed to be checked for emptiness. The new helper method chooses the right field automatically.


src/test/java/sirius/web/http/TestResponse.java
```diff
@@ -205,6 +206,41 @@ public StructuredNode xmlContent() {
         }
     }
 
+    /**
+     * @return whether this response has a body that contains data
```
andyHa: Returns .... nommal als extra zeile - ist JavaDoc standard...

PR #414, opened 2018-02-19
Trims trailing empty cells in excel imports

welzfabian:
Necessary, because SmartLineBasedProcessor will throw an exception, if a "Data"-row has more columns than the "Headings"-row. And we don't want that to be caused by empty columns in an excel file.

### DIFF
--- a/src/main/java/sirius/web/data/XLSProcessor.java
+++ b/src/main/java/sirius/web/data/XLSProcessor.java
@@ -55,7 +55,7 @@ public class XLSProcessor implements LineBasedProcessor {
             current++;
             Row row = iter.next();
             short first = 0;
-            short last = row.getLastCellNum();
+            short last = getLastFilledCell(row);
             List<Object> values = Lists.newArrayList();
             for (int i = first; i <= last; i++) {
                 Cell cell = row.getCell(i);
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/web/data/XLSProcessor.java
```diff
@@ -67,6 +67,19 @@ public void run(RowProcessor rowProcessor) throws Exception {
         }
     }
 
+    private short getLastFilledCell(Row row) {
+        short lastFilled = -1;
+        Iterator<Cell> cellIterator = row.cellIterator();
+        for (short current = 0; cellIterator.hasNext(); current++) {
+            Cell cell = cellIterator.next();
+            Object cellValue = extractCellValue(cell);
+            if (cellValue != null && Strings.isFilled(cellValue)) {
```
tbiScireum: null check überflüssig

src/main/java/sirius/web/data/XLSProcessor.java
```diff
@@ -67,6 +67,19 @@ public void run(RowProcessor rowProcessor) throws Exception {
         }
     }
 
+    private short getLastFilledCell(Row row) {
+        short lastFilled = -1;
+        Iterator<Cell> cellIterator = row.cellIterator();
+        for (short current = 0; cellIterator.hasNext(); current++) {
```
tbiScireum:
```
while (cellIterator.hasNext()) {
lastFilled ++;
...
}
```

PR #416, opened 2018-02-20
Implements taglib to create a js variable holding the body of this tag & invokeMacro

tbiScireum:
It can be used to create javascript variables holding mustache templates without having to
write ' $line ' + around each line.
Enables codeformat as well.


src/main/resources/default/taglib/w/jsTemplate.html.pasta
```diff
@@ -0,0 +1,9 @@
+<i:arg type="String" name="variableName" />
+<i:arg type="String" name="content" default="@renderToString('body')" />
+
+<i:pragma name="description"
```
andyHa: inline maybe?
testcase?

PR #417, opened 2018-02-22
Routes requests using the unencoded URI

meg1502:
This is as per RFC 3986, Section 2.4 as otherwise data can be intrepted as subcomponent delimiters
Especially %2F was interpreted as "/" which made 'one/t%2Fwo' route to 'one/t/wo'

Routed methods will still get their params decoded

### DIFF
--- a/src/main/java/sirius/web/controller/ControllerDispatcher.java
+++ b/src/main/java/sirius/web/controller/ControllerDispatcher.java
@@ -80,7 +80,7 @@ public class ControllerDispatcher implements WebDispatcher {
     }
 
     private boolean route(final WebContext ctx, boolean preDispatch) {
-        String uri = ctx.getRequestedURI();
+        String uri = ctx.getRawRequestedURI();
         if (uri.endsWith("/") && !"/".equals(uri)) {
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/web/controller/Route.java
```diff
@@ -212,7 +212,13 @@ private static int compileRouteURI(Route result, String[] elements, StringBuilde
                 Object effectiveValue = Value.of(value).coerce(parameterTypes[idx - 1], null);
                 setAtPosition(result, idx, effectiveValue);
             } else if ("**".equals(expr.getFirst())) {
-                result.add(Arrays.asList(value.split("/")));
+                //we need to split the encoded values so we dont mistake data for the delimiter
+                ArrayList<String> values = new ArrayList<>(Arrays.asList(m.group(i).split("/")));
```
andyHa: Arrays.asStream().map(this::decode).collect(Lambdas.into(result))

und decode als neue methode für value -> URLDecoder.decode(value, Charsets.UTF_8.name()) (das brauchen wir ja noch iwo

PR #418, opened 2018-02-23
Date Frame Taglib

fkrone:
This taglib allows to show its content only for a certain time e.g. for information which is valid only for a certain time. Furthermore a macro for calling NLS.parseUserString was created for easy handling of the dateFrame taglib.

### DIFF
--- a/src/main/java/sirius/tagliatelle/macros/ParseUserStringMacro.java
+++ b/src/main/java/sirius/tagliatelle/macros/ParseUserStringMacro.java
@@ -1,0 +1,58 @@
+/*
+ * Made with all the love in the world
+ * by scireum in Remshalden, Germany
+ *
+ * Copyright by scireum GmbH
+ * http://www.scireum.de - info@scireum.de
+ */
+
+package sirius.tagliatelle.macros;
+
+import sirius.kernel.di.std.Register;
+import sirius.kernel.nls.NLS;
+import sirius.tagliatelle.Tagliatelle;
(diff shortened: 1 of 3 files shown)


src/main/resources/default/taglib/w/dateFrame.html.pasta
```diff
@@ -0,0 +1,12 @@
+<i:arg name="showFrom" type="String" default=""/>
+<i:arg name="showUntil" type="String" default=""/>
+
+<i:pragma name="inline" value="true" />
+<i:pragma name="description" value="Provides an easy taglib to show certain parts only in a limited date frame" />
+
+<i:local name="isShowFromMatched" value="@!isFilled(showFrom) || !today.isBefore(parseUserString(java.time.LocalDate.class, showFrom).as(java.time.LocalDate.class))"/>
```
andyHa: use magic macro
i:if not i:local

src/main/java/sirius/tagliatelle/macros/ParseUserStringMacro.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.List;
+
+/**
+ * Parses the given user string into the desired object.
+ * <p>
+ *
+ * @see NLS#parseUserString(Class, String)
+ */
+@Register
+public class ParseUserStringMacro implements Macro {
```
andyHa: replace with isBefore and isAfter with mad string parsing capabilities

PR #423, opened 2018-02-28
Implements second macro calling syntax

tbiScireum:
this allows us to call ___macroName instead of @macroName
Necessary is this to use tagliatelle macros within js.pasta files without destroying the code
with the intellij code format

### DIFF
--- a/src/main/java/sirius/tagliatelle/compiler/Compiler.java
+++ b/src/main/java/sirius/tagliatelle/compiler/Compiler.java
@@ -357,7 +357,7 @@ public class Compiler extends InputProcessor {
     }
 
     /**
-     * Processes an expression started with an {@literal @}.
+     * Processes an expression started with an {@literal @} or with "___"
      *
      * @param block the block to append the expression or the parsed emitters to
      * @return <tt>true</tt> if an expression was parsed, <tt>false</tt> otherwise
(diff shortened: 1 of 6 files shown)


src/main/java/sirius/tagliatelle/compiler/Compiler.java
```diff
@@ -345,14 +345,25 @@ private boolean processTag(TagHandler parentHandler, CompositeEmitter block, Con
...
 
+    /**
+     * Evaluates if the current char marks the start of an expression.
+     * <p>
+     * If the current char is '@' or this and the next 2 chars all equal to '_' the current char marks the start of an expression.
+     *
+     * @return <tt>true</tt> if the current char could mark an expression, <tt>false</tt> otherwise
+     */
+    private boolean isPotentialExpression() {
+        return reader.current().is('@') || (reader.current().is('_') && reader.next().is('_') && reader.next(2).is('_'));
```
andyHa: this would be 3?

src/main/java/sirius/tagliatelle/compiler/Compiler.java
```diff
@@ -345,14 +345,25 @@ private boolean processTag(TagHandler parentHandler, CompositeEmitter block, Con
...
+
     /**
      * Processes an expression started with an {@literal @}.
      *
      * @param block the block to append the expression or the parsed emitters to
      * @return <tt>true</tt> if an expression was parsed, <tt>false</tt> otherwise
      */
     private boolean processExpression(CompositeEmitter block) {
-        if (!reader.current().is('@')) {
+        if (!isPotentialExpression()) {
```
andyHa: fix javadoc above

tbiScireum: done

PR #428, opened 2018-03-06
Fixes rewriting the URI in the WebContext

jmuscireum:
The ControllerDispatcher uses the raw uri to determine the correct route for a request. Now this raw uri is also updated when withCustomURI or withCustomPath are used to rewrite the request.

### DIFF
--- a/src/main/java/sirius/web/http/WebContext.java
+++ b/src/main/java/sirius/web/http/WebContext.java
@@ -818,8 +818,7 @@ public class WebContext implements SubContext {
      */
     public String getRawRequestedURI() {
         if (rawRequestedURI == null && request != null) {
-            int pathEndPos = request.uri().indexOf('?');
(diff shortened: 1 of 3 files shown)


src/main/java/sirius/web/http/WebContext.java
```diff
@@ -818,8 +818,7 @@ public String getRequestedURI() {
      */
     public String getRawRequestedURI() {
         if (rawRequestedURI == null && request != null) {
-            int pathEndPos = request.uri().indexOf('?');
-            rawRequestedURI = pathEndPos < 0 ? request.uri() : request.uri().substring(0, pathEndPos);
+            rawRequestedURI = stripQueryFromURI(request.uri());
```
tbiScireum: tests for the changed logic?

jmuscireum: done

PR #430, opened 2018-03-08
Some additions to the usermanagment

welzfabian:


src/main/java/sirius/web/controller/BasicController.java
```diff
@@ -75,7 +75,7 @@ private void defaultRoute(WebContext ctx, Method method) {
      *
      * @return the currently active user
      */
-    protected UserInfo getUser() {
+    protected static UserInfo getUser() {
```
andyHa: whyy - UserContext.getCurrentUser() is already static

src/main/java/sirius/web/security/UserInfo.java
```diff
@@ -417,4 +417,8 @@ public UserSettings getSettings() {
             return settingsSupplier.apply(this);
         }
     }
+
+    protected void setPermissions(Set<String> permissions) {
```
andyHa: why? - withPermissions is already there

PR #434, opened 2018-03-19
Introduce messages of type success

fkrone:
Until now, only info, warn and error messages could be created via the Message class. However, sometimes it is needed to create a success message. To avoid creating these messages manually, this message type is now introduced for the Message class.


fkrone:
BUILD

PR #453, opened 2018-05-11
Content-Security-Policy

jakobvogel:
- Adds support for Content-Security-Policy header and `<meta http-equiv>`


src/main/resources/default/taglib/w/page.html.pasta
```diff
@@ -9,6 +9,7 @@
 <html lang="@lang">
 <head>
     <meta http-equiv="X-UA-Compatible" content="IE=Edge"/>
+    <meta http-equiv="Content-Security-Policy" content="@contentSecurityPolicy"/>
```
andyHa: omit if left empty? &lt;i:if test="isFilled(contentSecurityPolicy)"&gt;...&lt;/

src/main/resources/component-web.conf
```diff
@@ -71,6 +71,11 @@ http {
     # agents caring about that). A detailed description of P3P can be found here: http://en.wikipedia.org/wiki/P3P
     addP3PHeader = true
 
+    # Determines the content security policy, i.e. which internet resources to load from which locations. By default,
+    # an unsecure "allow everything" policy is set. A more secure value could be "script-src 'self'". A description of
```
andyHa: shouldn't we also recommed inline-insecure-whatever?

jakobvogel: done

PR #459, opened 2018-05-26
Provides the possibility to overwrite the default error handling respons

tbiScireum:
In case Services need to create error responses using a specific format (e.g.) soap fault, webu, etc.
we need to be able to overwrite the default error handling.

By default nothing changes, which is why i implemented an interface default method rather then changing the interface to an abstract class.

### DIFF
--- a/src/main/java/sirius/web/services/ServiceCall.java
+++ b/src/main/java/sirius/web/services/ServiceCall.java
@@ -37,28 +37,15 @@ public abstract class ServiceCall {
      * Handles the given exception by creating an error response.
      *
      * @param error the exception to report
-     * @return <tt>true</tt> if the response was already committed, <tt>false</tt> otherwise
+     * @return {@link HandledException} the handled Exception
      */
-    public boolean handle(Throwable error) {
-        HandledException he = Exceptions.handle()
+    public HandledException handle(Throwable error) {
+        return Exceptions.handle()
                 .to(ServiceCall.LOG)
                 .error(error)
                 .withSystemErrorMessage("Service call to '%s' failed: %s (%s)",
                         ctx.getRequestedURI())
                 .handle();
-
-        if (ctx.isResponseCommitted()) {
-            ServiceCall.LOG.WARN("Cannot send service error for: %s. "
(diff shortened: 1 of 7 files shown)


src/main/java/sirius/web/services/ServiceCall.java
```diff
@@ -39,49 +36,29 @@ protected ServiceCall(WebContext ctx) {
     /**
      * Handles the given exception by creating an error response.
      *
-     * @param errorCode the error code to send
-     * @param error     the exception to report
+     * @param error the exception to report
+     * @return <tt>true</tt> if the response was already committed, <tt>false</tt> otherwise
      */
-    public void handle(String errorCode, Throwable error) {
+    public boolean handle(Throwable error) {
```
andyHa: see below

src/main/java/sirius/web/services/ServiceCall.java
```diff
@@ -176,7 +153,9 @@ protected void invoke(StructuredService serv) {
             // If the user unexpectedly closes the connection, we do not need to log an error...
             Exceptions.ignore(ex);
         } catch (Exception t) {
-            handle(null, t);
+            if (!handle(t)) {
```
andyHa: invoke handle just to log the exception and add the if (ctx.isResponseCommitted()) { check here

src/main/java/sirius/web/services/StructuredService.java
```diff
@@ -33,4 +37,45 @@
...
     void call(ServiceCall call, StructuredOutput out) throws Exception;
+
+    /**
+     * Handles the exception caused by handling the incoming call.
+     *
+     * @param call the HTTP request to process
+     * @param out  the encoder used to generate the desired output
+     * @param e    {@link Exception} causing an error while processing the request
+     */
+    default void callOnError(ServiceCall call, StructuredOutput out, Exception e) {
```
andyHa: handleException

andyHa: pass in a HandledException so that responsibilities are clear

PR #469, opened 2018-06-25
Easier item method for Page

fkrone:
The code in the new Page item method was used many times as boilerplate code. To avoid this the code is now used in the class itself.

Tags: page

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -9,7 +9,7 @@
         <version>3.14</version>
     </parent>
     <artifactId>sirius-web</artifactId>
-    <version>18.0</version>
+    <version>18.1</version>
     <name>SIRIUS web</name>
     <description>Provides a modern and scalable web server as SIRIUS module</description>
 
--- a/pom.xml
+++ b/pom.xml
@@ -9,7 +9,7 @@
         <version>3.14</version>
     </parent>
     <artifactId>sirius-web</artifactId>
-    <version>18.1</version>
+    <version>18.2</version>
     <name>SIRIUS web</name>
     <description>Provides a modern and scalable web server as SIRIUS module</description>
 
--- a/pom.xml
+++ b/pom.xml
@@ -9,7 +9,7 @@
         <version>3.14</version>
     </parent>
     <artifactId>sirius-web</artifactId>
-    <version>18.2</version>
+    <version>18.3</version>
     <name>SIRIUS web</name>
     <description>Provides a modern and scalable web server as SIRIUS module</description>
 
--- a/src/main/java/sirius/web/controller/Page.java
+++ b/src/main/java/sirius/web/controller/Page.java
@@ -86,12 +86,16 @@ public class Page<E> {
      * the supplied {@link Limit} as this method does also is able to determine whether
      * this page {@link Page#withHasMore(boolean) has more items to show} which are currently
      * not being displayed.
+     * <p>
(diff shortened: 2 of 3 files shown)


src/main/java/sirius/web/controller/Page.java
```diff
@@ -74,6 +76,30 @@
...
+     * This method should be supplied with a {@link List} which size was determined by
+     * the supplied {@link Limit} as this method does also is able to determine whether
+     * this page {@link Page#withHasMore(boolean) has more items to show} which are currently
+     * not being displayed.
+     *
+     * @param itemsSupplier the supplier to supply items to the current page limited by the provided limit
+     * @return the page itself for fluent method calls
+     */
+    public Page<E> withLimitedItemsSupplier(Function<Limit, List<E>> itemsSupplier) {
+        Limit supplierLimit = getCurrentLimit();
```
meg1502: can you set totalitems = supplieditems.size

fkrone: Not needed. The list is now sliced to the set page size

meg1502: not needed but pretty much expected to fill such fields in boileplate methode? wouldn't it make sense if i call page.getTotalItems to get the total number of items instead of 0 ?

fkrone: The number of total items is not known in this case. If the Limit is used with Limit::asPredicate a List is generated containing only pageSize + 1 elements.

meg1502: Then you have pagesize + 1 total items and not 0 total items

src/main/java/sirius/web/controller/Page.java
```diff
@@ -490,4 +516,18 @@ public boolean hasFacets() {
...
+     * Returns the current set {@link Limit} for this page based on the start set via {@link Page#withStart(int)}
+     * and size of the page set via {@link Page#withPageSize(int)} or their default values if not being set.
+     * <p>
+     * Using this limit e.g. via {@link Limit#asPredicate()} for slicing a list would return a list of size
+     * {@link Page#pageSize} + 1 for easier handling whether there are more elements to determine if there is
+     * another page.
+     *
+     * @return the current set limit for the page
+     */
+    public Limit getCurrentLimit() {
```
andyHa: this shouldn't be public? it reveals the inner workings of this class... as it is already used in withLimitedItemsSupplier in can be made private or even be inlined

src/test/java/sirius/web/controller/PageSpec.groovy
```diff
@@ -43,4 +43,26 @@ class PageSpec extends Specification {
...
+        elementsList.add("4")
+        elementsList.add("5")
+        elementsList.add("6")
+        elementsList.add("7")
+        when:
+        limitPage.withLimitedItemsSupplier{limit -> elementsList}
+        then:
+        limitPage.getItems().size() == 5
+        and:
+        limitPage.hasMore() == true
```
andyHa: add a test with less items and assert that more is false (and the code doesn't crash) ;-)

pom.xml
```diff
@@ -9,7 +9,7 @@
         <version>3.14</version>
     </parent>
     <artifactId>sirius-web</artifactId>
-    <version>18.0</version>
+    <version>18.1</version>
```
D3nnisH: jetzt `18.2`

PR #476, opened 2018-06-28
Adds support so resolving generically overloaded methods.

andyHa:
Using a real "getMethod" call takes generic type parameters
into account and selects the proper one. If this approach fails,
we use our own resolver using coercions etc.


tbiScireum:
BUILD

PR #480, opened 2018-07-03
 Allows disabled suggestions inside the autocomplete services

jmuscireum:


src/main/java/sirius/web/controller/AutocompleteHelper.java
```diff
@@ -112,6 +113,17 @@ public Completion setDescription(String description) {
...
         }
 
+        /**
+         * Sets if the suggestion should be disabled
+         * <p>
+         * Uses standard select2 functionality to have entries in the select, that can't be selected.
+         *
+         * @param disabled <tt>true</tt> if the suggestion is disabled, <tt>false</tt> otherwise
+         */
+        public void setDisabled(boolean disabled) {
```
andyHa: return this for fluent method calls

PR #481, opened 2018-07-04
 Reads InputStreamHandler as Unsigned Byte as per InputStream.read() spec

meg1502:
As the diff is not clear here is a quick rundown:

InputStreamHandler.read() returns only values from -1 to 255 as noted here:
https://docs.oracle.com/javase/7/docs/api/java/io/InputStream.html#read()

deleted read(byte[] b, int off, int len) implementation since the default implementation just calls read() so these bytes will also be between -1 and 255


src/main/java/sirius/web/http/InputStreamHandler.java
```diff
@@ -219,22 +219,7 @@ public int read() throws IOException {
...
         }
         try {
-            return buffer.readByte();
-        } finally {
-            buffer.release();
-        }
-    }
-
-    @Override
-    public int read(byte[] b, int off, int len) throws IOException {
```
andyHa: noo - this will fallback to a HIGHLY inefficient default implementation

src/main/java/sirius/web/http/InputStreamHandler.java
```diff
@@ -219,22 +219,7 @@ public int read() throws IOException {
...
-        ByteBuf buffer = getBuffer();
-        if (buffer == null) {
-            return -1;
-        }
-        try {
-            len = Math.min(buffer.readableBytes(), len);
-            buffer.readBytes(b, off, len);
-            return len;
+            return buffer.readUnsignedByte();
```
andyHa: please provide a test where this makes a difference

PR #488, opened 2018-08-10
Adds SafePost Testrequests with an attached CSRF Token

PhilHup:
Adds a SAFEPOST Request which allows for testing of Requests which require a valid CSRF Token.
Adds a dummy session to the TestRequestClass since a TestRequest itself has no ChannelHandlerContext.
Changes Visibility of LAST_CSRF_RECOMPUTE static string in CSRFHelper.
Adds a CSRFTokenSpec test whether SAFEPOST attaches a valid CSRF Token to the request.

### DIFF
--- a/src/main/java/sirius/web/http/CSRFHelper.java
+++ b/src/main/java/sirius/web/http/CSRFHelper.java
@@ -27,6 +27,9 @@ public class CSRFHelper {
      */
     public static final String CSRF_TOKEN = "CSRFToken";
 
+    /**
+     * Contains the parameter name of the date at which point the csrf token was last recomputed.
+     */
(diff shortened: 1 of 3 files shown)


src/main/java/sirius/web/http/CSRFHelper.java
```diff
@@ -27,7 +27,7 @@
      */
     public static final String CSRF_TOKEN = "CSRFToken";
 
-    private static final String LAST_CSRF_RECOMPUTE = "lastCSRFRecompute";
+    public static final String LAST_CSRF_RECOMPUTE = "lastCSRFRecompute";
```
andyHa: javadoc

PhilHup: done

PR #489, opened 2018-08-16
Custom id for editForm taglib

fkrone:
A custom id can now be defined for the editForm in order to make the taglib usable more than once in a template without id clashes.

### DIFF
--- a/src/main/resources/default/taglib/w/editForm.html.pasta
+++ b/src/main/resources/default/taglib/w/editForm.html.pasta
@@ -1,10 +1,11 @@
 <i:arg type="String" name="url" />
 <i:arg type="String" name="class" default=""/>
+<i:arg type="String" name="id" default="editForm"/>
 
 <i:pragma name="inline" value="true" />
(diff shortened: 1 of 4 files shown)


src/test/resources/application.conf
```diff
@@ -1 +0,0 @@
-sirius.customizations = ["customized"]
```
andyHa: please kill this whole file if not done already

src/main/resources/default/taglib/w/editForm.html.pasta
```diff
@@ -1,10 +1,11 @@
 <i:arg type="String" name="url" />
 <i:arg type="String" name="class" default=""/>
+<i:arg type="String" name="id" default="editForm"/>
 
 <i:pragma name="inline" value="true" />
 <i:pragma name="description" value="Renders an editor form within a Wondergem template" />
 
-<form action="@url" method="post" id="editForm" class="edit-form @class">
+<form action="@url" method="post" id="@id" class="edit-form @class">
```
andyHa: why would we need this?

PR #491, opened 2018-08-30
Recompute CSRF tokens in advance

fkrone:
Currently the CSRF tokens can expire by staying on a page for a few minutes in the time window they expire. A request in this scenario would submit an expired CSRF token which can cause errors. Therefore CSRF tokens will now be recomputed in advance to avoid them expiring while staying a few minutes on a site.

Tags: csrf

### DIFF
--- a/src/main/java/sirius/web/http/CSRFHelper.java
+++ b/src/main/java/sirius/web/http/CSRFHelper.java
@@ -35,6 +35,9 @@ public class CSRFHelper {
     @ConfigValue("http.csrfTokenLifetime")
     private static Duration csrfTokenLifetime;
 
+    @ConfigValue("http.csrfTokenAutoRenew")
+    private static Duration csrfTokenAutoRenew;
+
     /**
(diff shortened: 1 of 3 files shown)


src/main/java/sirius/web/http/CSRFHelper.java
```diff
@@ -35,25 +35,51 @@
...
         }
 
         return ctx.getSessionValue(CSRF_TOKEN).asString();
     }
 
-    private boolean isCSRFTokenOutdated(long lastCSRFRecompute) {
-        return Duration.between(Instant.ofEpochMilli(lastCSRFRecompute), Instant.now()).toMinutes()
-               > csrfTokenLifetime.toMinutes();
+    private boolean isCSRFTokenOutdated(long lastCSRFRecompute, boolean autoRenew) {
+        return Duration.between(Instant.ofEpochMilli(lastCSRFRecompute), Instant.now()).toMinutes() > getTokenLifetime(
```
andyHa: durations kann man direkt verlgiechen

PR #495, opened 2018-09-07
Provides a fix for IE Placeholder Bug for select2

tbiScireum:
Also fixes an issue where IE can not unselect a selected value because it can not select
an empty value on its own (This is also a safari issue)
- Fixes: SE-4768

### DIFF
--- a/src/main/java/sirius/web/http/UserAgent.java
+++ b/src/main/java/sirius/web/http/UserAgent.java
@@ -33,6 +33,8 @@ public class UserAgent {
     private static final String BLACKBERRY_10 = "bb10";
     private static final String WINDOWS = "windows";
     private static final String ARM = "arm";
+    private static final String INTERNET_EXPLORER = "trident";
+    private static final String MS_EDGE = "edge";
 
     private String userAgentString;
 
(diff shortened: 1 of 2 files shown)


src/main/resources/default/templates/wondergem/page-script.html.pasta
```diff
@@ -115,17 +115,29 @@
                     'background-repeat': 'no-repeat',
                     'background-position': '100% -22px'
                 });
-                var placeholder = field.attr('data-searchtext');
-                if (typeof placeholder === 'undefined' || placeholder === '') {
-                    placeholder = '@i18n("template.html.placeholder")';
+
+                ___if(!call.getUserAgent().isInternetExplorer())
+                {
```
andyHa: same line

src/main/java/sirius/web/http/UserAgent.java
```diff
@@ -75,6 +79,9 @@ protected void parseUserAgent() {
     }
 
     private boolean isWindowsDevice(String lowerCaseUserAgent) {
+        internetExplorer = lowerCaseUserAgent.contains(INTERNET_EXPLORER);
```
meg1502: this only detects IE 11 - is this intended?

tbiScireum: we could discuss supporting older IE versions as well but that should be a separat discussion i think

PR #498, opened 2018-09-17
Adjustments for mobile devices and responsive layouts

D3nnisH:


src/main/resources/default/taglib/w/tableRow.html.pasta
```diff
@@ -0,0 +1,47 @@
+<i:arg name="titleKey" type="String" default=""/>
```
andyHa: this template seems oddly specific... and also there is infobox and infoboxElement

D3nnisH: you are right, infobox is good 😋

PR #504, opened 2018-10-10
Proper cleanup of locals after for-loop

Tobsucht:
This commit changes the cleanup of local variables after a for-loop using the for-handler ('@for') or for-tag('<i:for>').
The problem was, that only the last element was popped of the stack, which only works if the loop just introduced the
loop variable but no additional local variables using '<i:local>'.


D3nnisH:
könntest nochn Test hinzufügen, wo du die nicht mehr sichtbare `<i:local name="test" ...` Variable später referenzierst und der Compiler sagt nö

PR #507, opened 2018-10-10
Proper cleanup of locals after if-else blocks

Tobsucht:
This commit changes the cleanup of local variables after a if-else using the if-handler ('@if') or if-tag('<i:if>').
The problem was, no elements were popped of the stack, when leaving such a block.

### DIFF
--- a/src/main/java/sirius/tagliatelle/compiler/CompilationContext.java
+++ b/src/main/java/sirius/tagliatelle/compiler/CompilationContext.java
@@ -190,9 +190,7 @@ public class CompilationContext {
     }
 
     /**
-     * Pops locals off the stack as long as their <tt>stackIndex</tt> is greater or equal than the one provided. In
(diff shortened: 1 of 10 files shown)


src/main/java/sirius/tagliatelle/compiler/CompilationContext.java
```diff
@@ -190,7 +190,9 @@ public void pop(Position position) {
     }
 
     /**
-     * Pops locals off the stack as long as their <tt>stackIndex</tt> is greater or equal than the one provided.
+     * Pops locals off the stack as long as their <tt>stackIndex</tt> is greater or equal than the one provided. In
+     * contrast to {@link #popUntil(Position, int)} this method does generate a {@link ParseError} when the stack
```
andyHa: tryPopUntil

PR #510, opened 2018-10-23
Makes adding data to the CrunchlogKernel thread-safe

welzfabian:
We are missing Crunchlog-Entries in OX and hope this solves this (OX-4451)

### DIFF
--- a/src/main/java/sirius/web/crunchlog/CrunchlogKernel.java
+++ b/src/main/java/sirius/web/crunchlog/CrunchlogKernel.java
@@ -10,7 +10,6 @@ package sirius.web.crunchlog;
 
 import com.alibaba.fastjson.JSON;
 import com.google.common.base.Charsets;
-import com.google.common.collect.Queues;
 import com.google.common.hash.Hashing;
 import sirius.kernel.Stoppable;
 import sirius.kernel.async.BackgroundLoop;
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/web/crunchlog/CrunchlogKernel.java
```diff
@@ -73,7 +73,7 @@
     private static final int MAX_FILE_SIZE = 1024 * 1024 * 10;
     private static final int MAX_BUFFERED_LINES = 16 * 1024;
 
-    protected Queue<Context> buffer = Queues.newArrayDeque();
+    protected Queue<Context> buffer = new ArrayBlockingQueue<>(MAX_BUFFERED_LINES);
```
andyHa: is this guaranteed to be threadsafe? isn't "blocking" a bad idea in our context - maybe use ConcurrentLinkedQueue

welzfabian: > is this guaranteed to be threadsafe?
yes, it is using a `ReentrantLock` internally and the test succeeds, which it did not before.

> isn't "blocking" a bad idea in our context?
yes, but we don't call a blocking method. one _could_ add a time out to `offer` and `poll` to make them blocking or use `put` and `take`

I looked into `ConcurrentLinkedQueue` which seems like a solid replacement, but I noticed, the `size()`-method is O(n) for `ConcurrentLinkedQueue` instead of O(1) for `ArrayBlockingQueue`. And we call `size` every time we add to the buffer.

PR #518, opened 2018-11-09
Provides a method for getting a File from the POST-Parameters

welzfabian:


src/main/java/sirius/web/http/WebContext.java
```diff
@@ -504,6 +504,30 @@ private Value fetchPostAttribute(String key) {
...
 
+    /**
+     * Returns a {@link File file} supplied by the request as POST-parameter.
+     *
+     * @param key used to specify which part of the post request should be used.
+     * @return a {@link File file} sent for the given key or <tt>null</tt> if none is available
+     */
+    public File getFile(String key) {
+        try {
+            FileUpload fileUpload = getFileData(key);
```
andyHa: handle parameters which are already a file - see getContentAsFile. also mark as @Nullable

PR #519, opened 2018-11-13
Add #handleError to RowProcessor to handle errors in LineBasedProcessors

welzfabian:


src/main/java/sirius/web/data/RowProcessor.java
```diff
@@ -14,15 +14,26 @@
...
      */
     void handleRow(int lineNumber, Values row);
+
+    /**
+     * Called if an {@link Exception} is thrown while parsing a row.
+     *
+     * @param error The {@link Exception} which was thrown
+     * @return <tt>true</tt> to continue execution, <tt>false</tt> to abort execution
+     */
+    default boolean handleError(Exception error) {
```
andyHa: why not require this to be present?

welzfabian: the thinking was, like that it still is a functional interface and can be used as lambda. also so old code does not have to change and we don't have to implement a handleError-method everywhere

But I am not sure the method really belongs in this interface, maybe a "withErrorHandler" in "LineBasedProcessor" would be smarter? what do you think?

src/main/java/sirius/web/data/LineBasedProcessor.java
```diff
@@ -42,7 +43,9 @@ static LineBasedProcessor create(String name, InputStream input) {
      * Starts processing and sends each line to the given rowProcessor.
      *
      * @param rowProcessor the processor which handles each row of the file
+     * @param errorHandler errorHandler which gets called on exceptions. returns <tt>true</tt> if exception was handled
+     *                     and <tt>false</tt> if exception should be rethrown.
```
andyHa:
```suggestion
     *                     and <tt>false</tt> if the exception should be rethrown.
```

PR #523, opened 2018-11-14
Aha/sparklines

andyHa:


src/main/resources/default/assets/wondergem/sparklines/sparkline.js
```diff
@@ -0,0 +1,60 @@
...
+ * http://www.scireum.de - info@scireum.de
+ *
+ * Inspired by https://github.com/eanbowman/sparkline.js
+ */
+function sparkline(element) {
+    var ctx = element.getContext("2d");
+    var data = element.getAttribute("data-sparkline");
+    var spark = data.split(',');
+    for (var i in spark) {
+        spark[i] = Math.round(parseFloat(spark[i]) * 100);
```
welzfabian: isn't that unnecessary because of `ratioH`?

andyHa: This simply saves the first two fractional digits before rounding...

andyHa: all computations past that point operate on intergers

welzfabian: has that performance benefits or why do we want to operate on intergers?

we still multiply the the spark-elements with `ratioH` which will probably be a float, doesn't that negate that?

PR #524, opened 2018-11-14
Allows to pass a custom SMTPConfiguration when sending mails

D3nnisH:


src/main/java/sirius/web/mails/SMTPConfiguration.java
```diff
@@ -54,7 +54,15 @@
...
-        Settings settings = UserContext.getSettings();
+        this(UserContext.getSettings());
+    }
+
+    /**
+     * Creates a new configuration based on the given settings.
+     *
+     * @param settings the SMTP settings
+     */
+    public SMTPConfiguration(Settings settings) {
```
andyHa: maybe provide also an explicit constructor with host et al as parameters?

PR #527, opened 2018-11-20
Creates a tagliatelle template for time inputs

D3nnisH:


src/main/java/sirius/tagliatelle/macros/ToTimeInputValueMacro.java
```diff
@@ -0,0 +1,63 @@
+/*
```
andyHa: remove redundancy - all 3 macros share a good amount of common code.

src/main/resources/default/taglib/w/timefield.html.pasta
```diff
@@ -0,0 +1,39 @@
...
+            @label
+        </span>
+        </label>
+    </i:if>
+    <input name="@name"
+           @if(isFilled(id)) { id="@id" }
+           type="time"
+           value="@toTimeInputValue(value)"
+           class="form-control input-block-level"
+           placeholder="HH:MM"
```
andyHa: i18n?

PR #528, opened 2018-11-22
Fixes the SMTPConfiguration class.

D3nnisH:
The settings from the conf files were not taken into account anymore, because Value#asString(defaultValue) does not return the default value if the value is an empty string.

This misbehaviour was created in #524

### DIFF
--- a/src/main/java/sirius/web/mails/SMTPConfiguration.java
+++ b/src/main/java/sirius/web/mails/SMTPConfiguration.java
@@ -63,13 +63,25 @@ public class SMTPConfiguration {
      * @param settings the SMTP settings
      */
     public SMTPConfiguration(Settings settings) {
-        this(settings.get("mail.host").asString(smtpHost),
-             settings.get("mail.port").asString(smtpPort),
-             settings.get("mail.user").asString(smtpUser),
-             settings.get("mail.password").asString(smtpPassword),
-             settings.get("mail.sender").asString(smtpSender),
-             settings.get("mail.senderName").asString(smtpSenderName),
-             settings.get("mail.useEnvelopeFrom").asBoolean(smtpUseEnvelopeFrom));
+        if (settings.get("mail.host").isFilled()) {
+            host = settings.get("mail.host").getString();
+            port = settings.get("mail.port").getString();
+            user = settings.get("mail.user").getString();
+            password = settings.get("mail.password").getString();
+        } else {
+            host = smtpHost;
+            port = smtpPort;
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/web/mails/SMTPConfiguration.java
```diff
@@ -63,13 +63,25 @@ public SMTPConfiguration() {
...
     public SMTPConfiguration(Settings settings) {
-        this(settings.get("mail.host").asString(smtpHost),
-             settings.get("mail.port").asString(smtpPort),
-             settings.get("mail.user").asString(smtpUser),
-             settings.get("mail.password").asString(smtpPassword),
-             settings.get("mail.sender").asString(smtpSender),
-             settings.get("mail.senderName").asString(smtpSenderName),
-             settings.get("mail.useEnvelopeFrom").asBoolean(smtpUseEnvelopeFrom));
+        if (settings.get("mail.host").isFilled()) {
+            host = settings.get("mail.host").getString();
```
meg1502: if its not important that all these settings depend on filled mail.host, maybe `value.replaceEmptyWith(xxx) `can be used?

D3nnisH: I would keep it like this because it makes no sense to mix the host/port/user/password settings from different places

andyHa: maybe we should move this code outside of the constructor and have _one instance_ with the default config (or an empty one). Doing this in a constructor semms convoluted.

src/main/java/sirius/web/mails/SMTPConfiguration.java
```diff
@@ -63,13 +63,22 @@ public SMTPConfiguration() {
...
+            user = settings.get("mail.user").getString();
+            password = settings.get("mail.password").getString();
+        } else {
+            host = smtpHost;
+            port = smtpPort;
+            user = smtpUser;
+            password = smtpPassword;
+        }
+
+        mailSender = settings.get("mail.sender").replaceEmptyWith(smtpSender).asString();
```
andyHa: do we really want to mix and merge those settings?

D3nnisH: no

src/main/java/sirius/web/mails/SMTPConfiguration.java
D3nnisH: +1

PR #529, opened 2018-11-28
Allows to fetch the browser language from WebsocketSessions

D3nnisH:


src/main/java/sirius/web/http/LangHelper.java
```diff
@@ -0,0 +1,56 @@
...
+    private static final Pattern ACCEPT_LANGUAGE_PATTERN =
+            Pattern.compile(" *([a-z]{2})(-[a-z]{2})? *(;q=([0-9.]+) *)?");
+
+    /**
+     * Returns the accepted language of the client as two-letter language code.
+     *
+     * @param request the HttpRequest to obtain the <tt>Accept-Language</tt> header from
+     * @return the two-letter code of the accepted language of the user agent or <tt>null</tt> if no valid accept language was found
+     */
+    public static String fromHttpRequest(HttpRequest request) {
```
andyHa: add @Nullable or even better return Optional<String>
accept a string as parameter to make this helper more universal

PR #535, opened 2018-12-12
Moves the setCurrentUser Logik into the UserManager

tbiScireum:
this methods binds the given user info the the user context and enables us to
extend the logic within the user manager.

This is necessary if we want to execute code after the user was sucessfully logged in
- Fixes: SE-4967

### DIFF
--- a/src/main/java/sirius/web/security/UserContext.java
+++ b/src/main/java/sirius/web/security/UserContext.java
@@ -24,11 +24,7 @@ import sirius.web.http.WebContext;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
-import java.util.Collection;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Optional;
+import java.util.*;
 
 /**
  * Used to access the current user and scope.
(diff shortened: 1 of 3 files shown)


src/main/java/sirius/web/security/UserContext.java
```diff
@@ -207,13 +207,18 @@ private void bindScopeToRequest(WebContext ctx) {
...
-            UserManager manager = getUserManager();
-            UserInfo user = manager.bindToRequest(ctx);
-            setCurrentUser(user);
+            user = manager.bindToRequest(ctx);
         } else {
-            setCurrentUser(UserInfo.NOBODY);
+            user = UserInfo.NOBODY;
         }
+
+        setCurrentUser(user);
```
andyHa:
```suggestion
       // Install the user to perform the verification on a fully populated context...
        setCurrentUser(user);
```

src/main/java/sirius/web/security/UserContext.java
```diff
@@ -207,13 +207,18 @@ private void bindScopeToRequest(WebContext ctx) {
...
+        user = manager.verifyUser(user);
+        setCurrentUser(user);
```
andyHa:
```suggestion
        // Install the effective user - which might be NOBODY to signal that the current user was blocked...
        setCurrentUser(user);
```

PR #536, opened 2018-12-13
Improves SMTPConfiguration merging.

D3nnisH:
Uses mailSender, mailSenderName and useSenderAndEnvelopeFrom from the second SMTPConfiguration even if first one is already valid.


src/main/java/sirius/web/mails/SMTPConfiguration.java
```diff
@@ -225,7 +225,15 @@ public boolean isUseSenderAndEnvelopeFrom() {
      */
     public SMTPConfiguration orElse(SMTPConfiguration other) {
```
andyHa: kill this method an use a proper logic per product...

D3nnisH:
build

D3nnisH:
build

PR #539, opened 2018-12-18
Some modifications to the tokenfield and autocompletion

welzfabian:


src/main/resources/default/templates/wondergem/scripts/tokenfield.html.pasta
```diff
@@ -47,26 +48,28 @@
...
                     this.value = this.element.val();
-                    var tokenfieldConfig = $.extend({delimiter: " "}, config.tokenfield );
+                    var tokenfieldConfig = $.extend({delimiter: " "}, config.tokenfield);
                     this.element.tokenfield(tokenfieldConfig);
 
                     this.tokenElement = $("#" + this.id + "-tokenfield");
 
-                    this.tokenRemovalElement = $(input.tokenRemovalTemplate);
-                    this.tokenRemovalElement.insertAfter(this.tokenElement);
+                    if (config.showRemovalElement === undefined || config.showRemovalElement) {
```
D3nnisH: `config.showRemovalElement !== false`

src/main/resources/default/templates/wondergem/scripts/autocomplete.html.pasta
```diff
@@ -14,15 +14,15 @@
          * @@type {{name: string, element: undefined, init: init}}
          */
         var anchor = {
-            name: "autocomplete-anchor",
+            selector: "#autocomplete-anchor",
```
tbiScireum: this would be a breaking change for us

sabieber: Whats your proposal to make it non breaking? The previous solution just isn't flexible enough for usage in all products

PR #540, opened 2018-12-20
Some additions to sirius.tokenfield and sirius.autocomplete

welzfabian:


src/main/resources/default/templates/wondergem/scripts/tokenfield.html.pasta
```diff
@@ -278,6 +275,10 @@
 
                 clearTokens: function () {
                     input.clearTokens();
+                },
+
+                removeToken: function (token) {
```
welzfabian: should be in next commit

src/main/resources/default/templates/wondergem/scripts/tokenfield.html.pasta
```diff
@@ -61,9 +61,6 @@
                     input.updateTokenRemovalVisibility();
 
                     this.element.on('tokenfield:removedtoken', function (e) {
-                        var identifier = e.attrs.value;
-                        $('input[data-identifier="' + identifier + '"]').remove();
-
```
welzfabian: i can't see why we do this, but I may miss something

src/main/resources/default/templates/wondergem/scripts/tokenfield.html.pasta
```diff
@@ -53,9 +53,6 @@
                         this.tokenRemovalElement.insertAfter(this.tokenElement);
                     }
 
-                    // delete the element value because they are tokens now and the value should not be displayed twice
-                    this.tokenElement.val("");
```
welzfabian: I can't see why we do this, but I maybe missed something

src/main/resources/default/templates/wondergem/scripts/tokenfield.html.pasta
welzfabian: I believe this were hidden inputs which where added to be available in posts. Theses were probably removed again.

I reimplemented the hidden inputs in https://github.com/scireum/sirius-web/pull/540/commits/d3acd32c5ca7e757901b18c26a1907b571a3035d

sabieber:
Some JSDoc for public API methods and event listeners would be highly appreciated

PR #552, opened 2019-01-21
Better statistics output for prometheus.

andyHa:
Improves metrics output to match the expectations of prometheus.io


src/main/java/sirius/web/http/IPRange.java
```diff
@@ -139,8 +139,22 @@ private static RangeSet createLocalHostRangeSet() {
      *                                   can be an address or a sub net in CIDR notation.
      * @return a RangeSet representing the given input. If the input was empty, a "no filter" set is returned, which
      * accepts all IPs.
+     * @deprecated typo in method name
      */
+    @Deprecated
     public static RangeSet paraseRangeSet(String commaSeparatedListOfRanges) {
+        return parseRangeSet(commaSeparatedListOfRanges);
```
meg1502: Exceptions.logDeprecatedMethodUse() ?

PR #554, opened 2019-01-22
Remove query string for Fontawsome

fkrone:
Some browsers (namely IE 11) sometimes do not load fonts if they are served with a query string.

Tags: ie, font, fontawsome

### DIFF
--- a/src/main/resources/default/assets/wondergem/font-awesome/css/font-awesome.css
+++ b/src/main/resources/default/assets/wondergem/font-awesome/css/font-awesome.css
@@ -6,8 +6,8 @@
  * -------------------------- */
 @font-face {
   font-family: 'FontAwesome';
-  src: url('/assets/wondergem/font-awesome/fonts/fontawesome-webfont.eot?v=4.5.0');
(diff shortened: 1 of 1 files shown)


src/main/resources/default/assets/wondergem/font-awesome/css/font-awesome.css
```diff
@@ -6,8 +6,8 @@
  * -------------------------- */
 @font-face {
   font-family: 'FontAwesome';
-  src: url('/assets/wondergem/font-awesome/fonts/fontawesome-webfont.eot?v=4.5.0');
-  src: url('/assets/wondergem/font-awesome/fonts/fontawesome-webfont.eot?#iefix&v=4.5.0') format('embedded-opentype'), url('/assets/wondergem/font-awesome/fonts/fontawesome-webfont.woff2?v=4.5.0') format('woff2'), url('/assets/wondergem/font-awesome/fonts/fontawesome-webfont.woff?v=4.5.0') format('woff'), url('/assets/wondergem/font-awesome/fonts/fontawesome-webfont.ttf?v=4.5.0') format('truetype'), url('/assets/wondergem/font-awesome/fonts/fontawesome-webfont.svg?v=4.5.0#fontawesomeregular') format('svg');
```
andyHa: are we 100% sure that killing off the "iefix" doesn't harm anything?

fkrone: This is only relevant for some legacy browsers (IE6 - IE8):
https://stackoverflow.com/a/8636675/9758089

PR #555, opened 2019-01-22
Use valid default for SMTP protocol

fkrone:
The setting value has to be in upper case.

Tags: smtp, configuration

### DIFF
--- a/src/main/java/sirius/web/mails/SMTPConfiguration.java
+++ b/src/main/java/sirius/web/mails/SMTPConfiguration.java
@@ -89,16 +89,19 @@ public class SMTPConfiguration {
      * @return a new configuration based on the config files.
      */
     public static SMTPConfiguration fromConfig() {
+
+        SMTPProtocol smtpProtocol = Sirius.getSettings()
+                                          .get("mail.smtp.protocol")
+                                          .getEnum(SMTPProtocol.class)
+                                          .orElseGet(() -> Value.of(Sirius.getSettings()
+                                                                          .get("mail.smtp.protocol")
+                                                                          .toUpperCase())
+                                                                .getEnum(SMTPProtocol.class)
+                                                                .orElse(SMTPProtocol.SMTP));
+
         return new SMTPConfiguration(smtpHost,
                                      smtpPort,
-                                     Sirius.getSettings()
-                                           .get("mail.smtp.protocol")
-                                           .getEnum(SMTPProtocol.class)
-                                           .orElseGet(() -> Value.of(Sirius.getSettings()
-                                                                           .get("mail.smtp.protocol")
-                                                                           .toUpperCase())
-                                                                 .getEnum(SMTPProtocol.class)
-                                                                 .orElse(SMTPProtocol.SMTP)),
+                                     smtpProtocol,
                                      smtpUser,
                                      smtpPassword,
                                      smtpSender,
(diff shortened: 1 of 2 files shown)


src/main/resources/component-web.conf
```diff
@@ -351,7 +351,7 @@ mail {
         port = "25"
 
         # Protocol and encryption of the SMTP connection (SMTP, SMTPS or STARTTLS)
-        protocol = "smtp"
+        protocol = "SMTP"
```
andyHa: can't we rather/also make the reader smarter?

D3nnisH: enum values are case insensitive, so this wouldn't be unambiguously anymore

src/main/java/sirius/web/mails/SMTPConfiguration.java
```diff
@@ -89,7 +91,14 @@ public SMTPConfiguration(String host,
     public static SMTPConfiguration fromConfig() {
         return new SMTPConfiguration(smtpHost,
                                      smtpPort,
-                                     smtpProtocol,
+                                     Sirius.getSettings()
+                                           .get("mail.smtp.protocol")
+                                           .getEnum(SMTPProtocol.class)
+                                           .orElseGet(() -> Value.of(Sirius.getSettings()
```
sabieber: Ich finde das als Inline Parameter richtig schrecklich. Würde vorschlagen das in ne Variable auszulagern

src/main/java/sirius/web/mails/SMTPConfiguration.java
```diff
@@ -115,7 +124,11 @@ public static SMTPConfiguration fromUserSettings() {
     public static SMTPConfiguration fromSettings(Settings settings) {
         return new SMTPConfiguration(settings.get("mail.host").getString(),
                                      settings.get("mail.port").getString(),
-                                     settings.get("mail.protocol").asEnum(SMTPProtocol.class),
+                                     settings.get("mail.protocol")
+                                             .getEnum(SMTPProtocol.class)
+                                             .orElseGet(() -> Value.of(settings.get("mail.protocol").toUpperCase())
```
sabieber: Siehe oben

src/main/java/sirius/web/mails/SMTPConfiguration.java
```diff
@@ -63,6 +63,8 @@
      * @param mailSenderName           the sender name
      * @param useSenderAndEnvelopeFrom whether to fill the "Sender" and the "Envelope-From" header
      */
+    @SuppressWarnings("squid:S00107")
+    @Explain("All settings are needed for a proper SMTP setting")
     public SMTPConfiguration(String host,
```
andyHa: maybe refactor to a builder pattern using a chain of withXXX "setters"

andyHa: this is a parameter object already anyway

src/main/java/sirius/web/mails/SMTPConfiguration.java
```diff
@@ -87,6 +89,16 @@ public SMTPConfiguration(String host,
      * @return a new configuration based on the config files.
      */
     public static SMTPConfiguration fromConfig() {
+
+        SMTPProtocol smtpProtocol = Sirius.getSettings()
```
andyHa: extract into method Value -> SMTPProtocol "asProtocol" or the like..

PR #557, opened 2019-01-23
Refresh on hidden fields can be disabled

fkrone:
If no name for the hidden input fields is defined, they should not be refreshed as they would be generated with undefinded as name. In addition some products rely on their on logic to handle these input fields.

Tags: tokenfield

### DIFF
--- a/src/main/resources/default/templates/wondergem/scripts/tokenfield.html.pasta
+++ b/src/main/resources/default/templates/wondergem/scripts/tokenfield.html.pasta
@@ -64,7 +64,7 @@
                 this.hiddenInputsName = config.hiddenInputsName;
 
                 function refreshHiddenInputs() {
-                    if (input.hiddenInputsName === undefined) {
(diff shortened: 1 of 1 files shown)


src/main/resources/default/templates/wondergem/scripts/tokenfield.html.pasta
```diff
@@ -64,6 +64,9 @@
                 this.hiddenInputsName = config.hiddenInputsName;
 
                 function refreshHiddenInputs() {
+                    if (input.hiddenInputsName === undefined) {
```
sabieber:
```suggestion
                    if (typeof input.hiddenInputsName === "undefined") {
```

welzfabian: would something like `if (!input.hiddenInputsName)` not be the most simple solution?

or what is the advantage of this "typeof"-check?

PR #560, opened 2019-01-24
Use 'tokenfield' for 'autocompleteMultiSelect'

welzfabian:


src/main/resources/default/taglib/w/autocompleteMultiSelect.html.pasta
```diff
@@ -3,7 +3,6 @@
 <i:arg name="name" type="String"/>
 <i:arg name="type" type="String" default=""/>
 <i:arg name="serviceUri" type="String"/>
-<i:arg name="searchKey" type="String" default=""/>
```
welzfabian: "breaking change", because users have to remove parameter from their call. should I increment major-version?

sabieber: yes

andyHa: if we'd release a major - drop the two recently introduced @Deprecated (sonarlint shows them)

welzfabian: ok, should I add that to this PR or create a separate PR for that?

src/main/resources/default/taglib/w/multiSelect.html.pasta
```diff
@@ -35,7 +36,7 @@
         multiSelect({
             id: '@id',
             name: '@name',
-            noMatchesText : '@i18n("Select2.noMatches")',
+            noMatchesText : '@i18n("template.html.select2.noMatches")',
```
meg1502: would it make sense to make these keys more generic and drop references to select2?

sabieber: i think fabian is already working on this :)

welzfabian: +1

PR #561, opened 2019-01-31
Adds Tokens to tokenfield from autocomplete without rendering them in the tokenfield

meg1502:
Previously, tokens were added because the tokenField lost focus and the *selectedRow*s value was written to the tokenField, which creates the token for this value when it loses value.
Without a total rewrite, we now add the token by hand just before the autocomplete gets hidden and we have a *selectedRow*

### DIFF
--- a/src/main/resources/default/templates/wondergem/page-script.html.pasta
+++ b/src/main/resources/default/templates/wondergem/page-script.html.pasta
@@ -612,12 +612,8 @@
         var tokenfield = sirius.createTokenfield();
         var autocomplete = sirius.createAutocomplete();
 
-        autocomplete.on('onHoverRow', function ($row) {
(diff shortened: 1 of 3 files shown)


src/main/resources/default/templates/wondergem/page-script.html.pasta
```diff
@@ -612,12 +612,8 @@
         var tokenfield = sirius.createTokenfield();
         var autocomplete = sirius.createAutocomplete();
 
-        autocomplete.on('onHoverRow', function ($row) {
-            autocomplete.getInput().val($row.find('.autocomplete-data').data('autocomplete'));
-        });
-
-        autocomplete.on('onUnhoverRow', function ($row) {
-            autocomplete.getInput().val('');
+        autocomplete.on('onBeforeHide', function (selectedRow) {
```
welzfabian: every where else we call it `$row`

PR #564, opened 2019-02-01
Fix some Bugs in the MultiSelect

welzfabian:
The 4th and 5th commit are not relevant, because the later is and revert of the earlier. And the 6th is a merge from master.

### DIFF
--- a/src/main/resources/default/taglib/w/autocompleteMultiSelect.html.pasta
+++ b/src/main/resources/default/taglib/w/autocompleteMultiSelect.html.pasta
@@ -13,7 +13,6 @@
 <i:arg name="required" type="boolean" default="false"/>
 <i:arg name="readonly" type="boolean" default="false"/>
 <i:arg name="adminOnly" type="boolean" default="false"/>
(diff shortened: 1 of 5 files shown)


src/main/resources/default/taglib/w/autocompleteMultiSelect.html.pasta
```diff
@@ -13,6 +13,7 @@
 <i:arg name="required" type="boolean" default="false"/>
 <i:arg name="readonly" type="boolean" default="false"/>
 <i:arg name="adminOnly" type="boolean" default="false"/>
+<i:arg name="spaceAsDelimiter" type="boolean" default="false" description="If true the delimiter is ' '. If false the delimiter is '|'"/>
```
meg1502: are you sure " " does not work as default for a String arg ?

src/main/resources/default/templates/wondergem/page-script.html.pasta
```diff
@@ -653,6 +653,7 @@
             showRemovalElement: false,
             hiddenInputsName: args.name,
             tokenfield: {
+                delimiter: args.delimiter || '|',
```
meg1502: or just use ' ' as default here...

PR #571, opened 2019-02-11
Creates mail taglibs and a default template

D3nnisH:
Fixes: SIRI-18

moved from https://github.com/scireum/sirius-biz/pull/241

### DIFF
--- a/src/main/resources/default/taglib/mail/text.html.pasta
+++ b/src/main/resources/default/taglib/mail/text.html.pasta
@@ -1,0 +1,14 @@
+<i:arg type="String" name="labelKey" default=""/>
+<i:arg type="String" name="label" default="@i18n(labelKey)"/>
+<i:arg type="String" name="value"/>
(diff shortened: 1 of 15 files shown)


src/main/resources/default/taglib/mail/text.html.pasta
```diff
@@ -0,0 +1,14 @@
...
+<i:arg type="String" name="label" default="@i18n(labelKey)"/>
+<i:arg type="String" name="value"/>
+<i:arg type="String" name="class" default=""/>
+<i:arg type="boolean" name="showEmptyValue" default="false"/>
+
+<i:pragma name="inline" value="true"/>
+
+<i:if test="@isFilled(value) || showEmptyValue">
+    <tr class="align-top @class">
+        <td class="order-element-label">@label</td>
```
andyHa: order element?

D3nnisH: habs umbenannt

PR #572, opened 2019-02-11
Aha/siri 14

andyHa:


src/main/resources/default/templates/system/load.html.pasta
```diff
@@ -0,0 +1,53 @@
...
+        </div>
+        <div class="clearfix"></div>
+    </w:pageHeader>
+
+    <i:for type="sirius.web.health.LoadInfoProvider" var="provider" items="providers">
+        <w:heading>@provider.getLabel()</w:heading>
+        <table class="table table-striped">
+            <i:for type="sirius.web.health.LoadInfo" var="info" items="provider.collectLoadInfos()">
+                <i:if test="all || info.isNonZero()">
+                    <tr class="@if (info.isNonZero()) {G REEN } else { GRAY }">
```
meg1502:
```suggestion
                    <tr class="@if (info.isNonZero()) {GREEN } else { GRAY }">
```

src/main/java/sirius/web/health/CachingLoadInfoProvider.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.List;
+import java.util.function.Consumer;
+
+/**
+ * Provides a basic implementation which caches all computed infos for up to ten seconds.
+ * <p>
+ * Be aware, that subclasses have to specify <tt>LoadInfoProvider</tt> in their <tt>Register</tt> annotation:
+ * {@code @Register(classes = LoadInfoProvider.class)}.
+ */
+public abstract class CachingLoadInfoProvider implements LoadInfoProvider {
```
welzfabian: "CachedLoadInfoProvider"?
Or els it sounds like it would provide some kind of info about the "caching load", but it is supposed to be the cached load infos, if i am not mistaken

andyHa: Well it is a load info provider which does caching :) Otherwise is would be a CacheLoadInfoProvider (?)

welzfabian: ok :)

PR #579, opened 2019-02-20
 Use tokenfield for singleSelect and autocompleteSingleSelect

welzfabian:


src/main/resources/default/assets/wondergem/autocomplete/autocomplete.css
```diff
@@ -53,3 +53,52 @@
     border-bottom: 1px solid #d9d9d9;
     margin: 2px 0;
 }
+
+.multiselect.singleselect .tokenfield {
```
sabieber: I find these two classes in combination quite confusing :)

welzfabian:
Please wait for review of @tbiScireum before merging

PR #583, opened 2019-02-25
SE-8199: Fixes small tokenfield issues

tbiScireum:


src/main/resources/default/templates/wondergem/scripts/tokenfield.html.pasta
```diff
@@ -146,6 +146,13 @@
                 $(window).resize(function (e) {
                     input.resize(e);
                 });
+
+                // resizes the input field after the tokenfield init process is done
+                // for some reason the tokenfield init event is called to soon and the internal resizing is not yet
```
sabieber:
```suggestion
                // for some reason the tokenfield init event is called too soon and the internal resizing is not yet
```

src/main/resources/default/templates/wondergem/scripts/tokenfield.html.pasta
```diff
@@ -146,6 +146,13 @@
...
+                // done.
+                // this will fix make the tokenRemovalElement clickable right from the beginning rather from the first
```
sabieber:
```suggestion
                // this will make the tokenRemovalElement clickable right from the beginning rather than only after the first
```

src/main/resources/default/templates/wondergem/scripts/tokenfield.html.pasta
```diff
@@ -224,11 +231,11 @@
                     tokensWidth += $(this).outerWidth(true) + 1;
                 });
 
-                var removalElementWidth = 0;
-                if (input.tokenRemovalElement) {
-                    // + 1 because outerWidth gets the width rounded and else it might not fit because of 1px
-                    removalElementWidth = input.tokenRemovalElement.outerWidth(true) + 1;
-                }
+                // + 1 because outerWidth gets the width rounded and else it might not fit because of 1px
+                var removalElementWidth = 1;
```
welzfabian: maybe call it additionalElementsWidth or something like that?

PR #585, opened 2019-02-26
SIRI-43 Use the original selects to transfer the selection

jmuscireum:


jmuscireum:
I checked that already and it does

PR #598, opened 2019-03-14
Fix usage and reading of delayTime

welzfabian:
"this.delay = config.delayTime || this.delay;" sets delay to "this.delay" (400) if "config.delayTime" is 0, but we actually want to be able to set it to 0

### DIFF
--- a/src/main/resources/default/templates/wondergem/scripts/autocomplete.html.pasta
+++ b/src/main/resources/default/templates/wondergem/scripts/autocomplete.html.pasta
@@ -667,7 +667,7 @@
                     this.responseCache = config.cache;
                 }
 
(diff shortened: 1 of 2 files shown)


src/main/resources/default/templates/wondergem/scripts/autocomplete.html.pasta
```diff
@@ -667,7 +667,9 @@
                     this.responseCache = config.cache;
                 }
 
-                this.delay = config.delayTime || this.delay;
+                if (typeof config.delayTime !== "undefined") {
```
sabieber:
```suggestion
                if (typeof config.delayTime !== 'undefined') {
```

PR #605, opened 2019-03-21
Use plain JavaScript for check if input field is focused

welzfabian:
The '.is(":focus")' check fails in Safari, when the field was focused via the '.focus()' method.

### DIFF
--- a/src/main/resources/default/templates/wondergem/scripts/autocomplete.html.pasta
+++ b/src/main/resources/default/templates/wondergem/scripts/autocomplete.html.pasta
@@ -365,7 +365,7 @@
             },
 
             show: function () {
-                if (this.rows.length > 0 && input.element.is(":focus")) {
+                if (this.rows.length > 0 && document.activeElement === input.element.get(0)) {
                     this.update();
                     completions.element.show();
                     this.state = completions.VISIBLE;


src/main/resources/default/templates/wondergem/scripts/autocomplete.html.pasta
```diff
@@ -365,7 +365,7 @@
             },
 
             show: function () {
-                if (this.rows.length > 0 && input.element.is(":focus")) {
+                if (this.rows.length > 0 && document.activeElement === input.element.get(0)) {
```
meg1502:
```suggestion
                if (this.rows.length > 0 && document.activeElement && document.activeElement === input.element.get(0)) {
```

so it wont become true for for null === null

meg1502: maybe also check input.element for null/undefined to prevent exception when calling input.element.get

welzfabian: `input.element.get(0)` should never be null/undefined after the initialization of the autocomplete. and show is a 'private' method so it should never be called from outside this file.

so i am not really sure if we should add this check, because this check could then hide other problems

PR #606, opened 2019-03-21
Also blur input field of SingleSelect after adding a token

welzfabian:
Because in Safari this still can have focus after being hidden, unlike other browsers


welzfabian:
and if it keeps focus the completions are not closing

PR #611, opened 2019-03-26
Allow for installing a user into the context directly 

PhilHup:
Allows for installing and verifying a user into the current Context directly.
Splits up the old bindUserToRequest logik into two seperate functions.


src/main/java/sirius/web/security/UserContext.java
```diff
@@ -208,21 +208,29 @@ private void bindScopeToRequest(WebContext ctx) {
      * Loads the current user from the given web context.
      */
     private void bindUserToRequest(WebContext ctx) {
-        UserManager manager = getUserManager();
         UserInfo user;
```
andyHa: then kill this uninitialized variable and call installUserInfoContext in each branch

PR #613, opened 2019-03-27
Adds a tabindex parameter to the textfield

PhilHup:
Add tabindex to textfield taglib.


src/main/resources/default/taglib/w/textfield.html.pasta
```diff
@@ -14,6 +14,7 @@
 <i:arg name="type" type="String" default="text"/>
 <i:arg name="id" type="String" default=""/>
 <i:arg name="placeholder" type="String" default=""/>
+<i:arg name="tabindex" type="String" default=""/>
```
andyHa: tabIndex?

PR #619, opened 2019-05-14
Aha/better state

andyHa:


src/main/java/sirius/web/health/Cluster.java
```diff
@@ -167,13 +222,30 @@ private void updateClusterStateHistory() {
...
+        }
+
+        // Provide appropriate logs to support determining the downtime of a system or at least the
+        // duration of an incident.
+        if (alarmIsPresentNow) {
+            LOG.WARN("PROBLEM: The cluster state was RED in %s out of %s intervals.",
+                     countFailedIntervals(),
+                     getMonitoringIntervals());
+        } else {
+            LOG.INFO("OK: The cluster state returned to GREEN.");
```
tbiScireum: FINE ? otherwise this would create lots of logs ?

PR #622, opened 2019-05-17
Provides a dropdown with two options as an alternative to w:checkbox.

andyHa:
Layouting checkboxes is often difficult, as the only make up a single
line instead of the commonly used two.


andyHa:
grafik

PR #627, opened 2019-05-31
Move the "hasPermission" logic to the Permissions helper class

welzfabian:


src/main/java/sirius/web/security/Permissions.java
```diff
@@ -157,4 +172,59 @@ public static boolean wearsPermissionAnnotation(AnnotatedElement object) {
...
+     * <tt>and</tt>. An example would be "logged-in,important-customer+!locked". This would translate to "the user has
+     * to be logged in or it has to be an important customer and not be locked".
+     *
+     * @param permissionExpression the permission expression to check
+     * @param permissions          all permissions as set
+     * @return <tt>true</tt> if the given permissions contains the permission expression, <tt>false</tt> otherwise
+     */
+    @SuppressWarnings("squid:S2259")
+    @Explain("permissionExpression can not be null due to Strings.isEmpty")
+    public static boolean hasPermission(@Nullable String permissionExpression, @Nullable Set<String> permissions) {
```
welzfabian: I am planning to refactor the "hasRequiredPermissionForPermission" of (this)[https://github.com/scireum/sirius-biz/pull/350] PR to take a Set instead of a Predicate, so this method is usable there

welzfabian: no need to implement an `hasPermission(@Nullable String permissionExpression, @Nullable Predicate<String> hasPermission)` in my opinion

andyHa: use a Predicate<String> instead of a set here to make it easier to reuse

andyHa: no need for an implementation imposing a Set, Set::contains is an off the shelf impl for said prediacte.

src/main/java/sirius/web/security/UserInfo.java
```diff
@@ -286,42 +272,7 @@ public boolean hasPermissions(String... permissions) {
...
-        return true;
-    }
-
-    protected boolean permissionFullfilled(String permission) {
-        if (permission.startsWith("!")) {
-            return permissions == null || !permissions.contains(permission.substring(1));
-        } else {
-            return permissions != null && permissions.contains(permission);
-        }
+        return Permissions.hasPermission(permission, permissions == null ? s -> false : permission::contains);
```
andyHa: permissions:contains

andyHa: I wonder how this could pass the unit tests...

andyHa: Please change accordingly

welzfabian: 😱🤦‍♂️

PR #628, opened 2019-06-03
Make UserInfo::getPermissions null-safe

welzfabian:


meg1502:
isn't null something else than empty at this point? 
null = user has not been filled yet
empty = user has no permissions

Im not sure if its a good idea to treat them as the same, as this method might be used to indicate the 'null' state

PR #631, opened 2019-06-12
Fix linechart, add support for data unit in polar area charts

Tobsucht:


src/main/resources/default/templates/wondergem/page-script.html.pasta
```diff
@@ -372,7 +372,21 @@
             });
     }
 
-    function polarAreaChart(selector, labels, data) {
+    function polarAreaChart(selector, labels, data, dataUnit) {
+        var options = {
+            startAngle: -0.55 * Math.PI
+        };
+
+        if (!!dataUnit) {
```
andyHa: !! ?

src/main/resources/default/templates/wondergem/page-script.html.pasta
```diff
@@ -382,9 +396,7 @@
                 labels: labels
             },
             type: 'polarArea',
-            options: {
-                startAngle: -0.55 * Math.PI
```
andyHa: why is that dropped?

Tobsucht: just moved to the top of the function

src/main/resources/default/templates/wondergem/page-script.html.pasta
sabieber: das ist ein JS Cast zu boolean :)

PR #632, opened 2019-06-14
Adds an optional appendix for the username in UserInfo.

meg1502:
This is used in the new Method UserInfo.getProtocolName(), used to further identify a user.

### DIFF
--- a/src/main/java/sirius/web/security/UserInfo.java
+++ b/src/main/java/sirius/web/security/UserInfo.java
@@ -333,7 +333,14 @@ public class UserInfo extends Composable {
      * @return the name of the user
      */
     public String getProtocolUsername() {
-        if (nameAppendixSupplier == null) {
+        if (nameAppendixSupplier != null) {
+           String appendix = nameAppendixSupplier.get();
+           if (Strings.isFilled(appendix)) {
+              return Strings.apply("%s (%s)", username, nameAppendixSupplier.get());
+           }
+        }
+        
+        return username;
             return username;
         }
         return Strings.apply("%s (%s)", username, nameAppendixSupplier.get());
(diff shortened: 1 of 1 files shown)


src/main/java/sirius/web/security/UserInfo.java
```diff
@@ -101,6 +104,19 @@ public Builder withUsername(String name) {
...
         }
 
+        /**
+         * Contain additional info, added to the Username to be used in Protocols. See {@link #getProtocolUsername()}.
+         * This should be filled if multiple users can have the same {@link #username}s or if a user acts on the behalf of another user.
+         *
+         * @param appendix the additional info about the user.
+         * @return the builder itself for fluent method calls
+         */
+        public Builder withProtocolUsernameAppendix(String appendix) {
```
andyHa: Supplier&lt;String&gt;

src/main/java/sirius/web/security/UserInfo.java
```diff
@@ -300,6 +326,19 @@ public boolean isLoggedIn() {
...
     }
 
+    /**
+     * Returns the login or descriptive name of the user used in Protocols.
+     * Contains additional info given via {@link #nameAppendixSupplier} to further identify the user.
+     *
+     * @return the name of the user
+     */
+    public String getProtocolUsername() {
+        if (nameAppendixSupplier == null) {
```
andyHa:
```suggestion
        if (nameAppendixSupplier != null) {
           String appendix = nameAppendixSupplier.get();
           if (Strings.isFilled(appendix)) {
              return Strings.apply("%s (%s)", username, nameAppendixSupplier.get());
           }
        }
        
        return username;
```

PR #633, opened 2019-06-14
Hides a dropdown menu if it would be empty.

andyHa:


src/main/resources/default/taglib/w/menuDropdown.html.pasta
```diff
@@ -1,19 +1,23 @@
 <i:arg type="String" name="titleKey" default="" />
 <i:arg type="String" name="title" default="@i18n(titleKey)" />
 <i:arg type="String" name="permission" default="" />
+<i:arg type="String" name="framework" default="" />
 
 <i:pragma name="inline" value="true" />
 <i:pragma name="description" value="Renders menu dropdown within a Wondergem template" />
+<i:local name="contents" value="@renderToString('body')" />
```
welzfabian: why `local` if you only display it? 

did you mean to check if `isFilled`?

andyHa: indeed, fixed below

PR #634, opened 2019-06-19
Do not add Messages from MessageProviders to List in UserContext

welzfabian:
Or else the Messages of the MessageProvider and MaintenanceInfo will be added to the list everytime "public List<Message> getMessages()" is called, and therefore will be in the list multiple times if the method is called multiple times.

Messages where displayed twice, because sirius-biz calls "validate" which calls "getMessages()", when displaying a tenant, so "getMessages()" was called twice.
https://github.com/scireum/sirius-biz/blob/master/src/main/java/sirius/biz/tenants/TenantController.java#L145

### DIFF
--- a/src/main/java/sirius/web/security/UserContext.java
+++ b/src/main/java/sirius/web/security/UserContext.java
@@ -24,6 +24,7 @@ import sirius.web.http.WebContext;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


src/main/java/sirius/web/security/UserContext.java
```diff
@@ -327,14 +328,16 @@ public void addMessage(Message msg) {
...
-            getUser().tryAs(MessageProvider.class).ifPresent(provider -> provider.addMessages(this::addMessage));
-            messageProviders.forEach(provider -> provider.addMessages(this::addMessage));
+                      .ifPresent(messages::add);
+            getScope().tryAs(MessageProvider.class).ifPresent(provider -> provider.addMessages(messages::add));
+            getUser().tryAs(MessageProvider.class).ifPresent(provider -> provider.addMessages(messages::add));
+            messageProviders.forEach(provider -> provider.addMessages(messages::add));
+            return messages;
         }
 
         return msgList;
```
andyHa: protect against modifications

src/main/java/sirius/web/security/UserContext.java
```diff
@@ -327,14 +328,16 @@ public void addMessage(Message msg) {
...
                       .filter(Objects::nonNull)
-                      .ifPresent(this::addMessage);
-            getScope().tryAs(MessageProvider.class).ifPresent(provider -> provider.addMessages(this::addMessage));
-            getUser().tryAs(MessageProvider.class).ifPresent(provider -> provider.addMessages(this::addMessage));
-            messageProviders.forEach(provider -> provider.addMessages(this::addMessage));
+                      .ifPresent(messages::add);
+            getScope().tryAs(MessageProvider.class).ifPresent(provider -> provider.addMessages(messages::add));
+            getUser().tryAs(MessageProvider.class).ifPresent(provider -> provider.addMessages(messages::add));
+            messageProviders.forEach(provider -> provider.addMessages(messages::add));
+            return messages;
```
andyHa: protect against modifications

PR #635, opened 2019-07-02
Set and reset the language for the user in runAs

welzfabian:


src/main/java/sirius/web/security/UserContext.java
```diff
@@ -302,11 +302,17 @@ public void setCurrentUser(@Nullable UserInfo user) {
      */
     public void runAs(@Nullable UserInfo user, @Nonnull Runnable section) {
         UserInfo lastUser = getCurrentUser();
+        CallContext call = CallContext.getCurrent();
+        String lastLang = call.getLang();
         try {
             setCurrentUser(user);
```
andyHa: shouldn't we just set lang to null above and let the system figure out the rest (defaults etc.)

src/main/java/sirius/web/security/UserContext.java
```diff
@@ -302,11 +302,17 @@ public void setCurrentUser(@Nullable UserInfo user) {
...
+            if (user != null) {
+                call.setLang(user.getLang());
+            }
             section.run();
         } finally {
             setCurrentUser(lastUser);
```
andyHa: reset lang first?

welzfabian:
needs https://github.com/scireum/sirius-kernel/pull/202

PR #641, opened 2019-07-25
Aha/better tests

andyHa:


src/test/java/sirius/tagliatelle/ReportBrokenTemplates.java
```diff
@@ -0,0 +1,61 @@
...
+
+    @Part
+    private Tagliatelle engine;
+
+    @Part
+    private Resources resources;
+
+    @Override
+    public void beforeTests() {
+
```
meg1502: short comment indicating why this method is empty
```
//no setup required
```

PR #643, opened 2019-07-26
Applies permission profiles in order instead of randomly.

andyHa:
Also permission profiles can now have complex names like
permission1+permission2,permission3 to reflect complex requirements
on when to provide a permission.

### DIFF
--- a/src/main/java/sirius/web/security/Permissions.java
+++ b/src/main/java/sirius/web/security/Permissions.java
@@ -8,19 +8,19 @@
 
 package sirius.web.security;
 
-import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
 import sirius.kernel.Sirius;
 import sirius.kernel.commons.Explain;
 import sirius.kernel.commons.Strings;
-import sirius.kernel.commons.Value;
 import sirius.kernel.di.std.ConfigValue;
 import sirius.kernel.settings.Extension;
 
 import javax.annotation.Nullable;
 import java.lang.reflect.AnnotatedElement;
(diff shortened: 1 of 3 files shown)


src/main/java/sirius/web/security/Permissions.java
```diff
@@ -49,82 +49,100 @@
...
+                profile.removePermission(permission.getKey());
             }
         }
+
+        return profile;
     }
 
     /**
-     * Expands all permission profiles to obtain the effective set of permissions for a given list or permission and
+     * Expands all permission profiles to obtain the effective set of permissions for a set list or permission and
```
meg1502: ? javadoc does not make sense

src/main/java/sirius/web/security/Permissions.java
```diff
@@ -49,82 +49,100 @@
...
     /**
-     * Expands all permission profiles just like {@link #applyProfiles(java.util.Collection)}. Also all public roles
+     * Expands all permission profiles just like {@link #applyProfiles(java.util.Set)}. Also all public roles
      * defined in <tt>security.publicRoles</tt> are included to the roles set before profiles are expanded.
      *
      * @param roles the list of permissions and or profiles to expand
      * @return an effective list of permissions based on the profiles defined in <tt>security.profiles</tt> and the
      * public roles defined in <tt>security.publicRoles</tt>
      */
     public static Set<String> applyProfilesAndPublicRoles(Collection<String> roles) {
```
meg1502: this method signature should probably be changed too, to work like applyProfiles, given the similiar naming

PR #651, opened 2019-08-05
Aha/nullsafe context extenders

andyHa:


src/main/java/sirius/web/templates/Templates.java
```diff
@@ -78,7 +79,47 @@ public Generator generator() {
     public Map<String, Object> createGlobalContext() {
         Map<String, Object> result = new LinkedHashMap<>();
         for (GlobalContextExtender extender : extenders) {
-            extender.collectTemplate(result::put);
+            extender.collectTemplate(new GlobalContextExtender.Collector() {
```
meg1502: could this collector which just collects with

result.put(name, object); 

maybe be defined somewhere as defaultCollector? as it is used 3 times already.

PR #670, opened 2019-10-14
Fix autosizing for cells in streaming ExcelExports

welzfabian:
With the new Version of Apache-POI column-widths must be explicitly tracked if you want to be able to call "Sheet#autoSizeColumn(int column)" (like in ExcelExport#autosizeColumns)


welzfabian:
@sabieber yes actually a massive performance degregation:
for a million excel rows -> 105 seconds vs. 7 seconds

PR #675, opened 2019-10-22
Adds a limit to the ExcelExport according to the used Workbook

welzfabian:
So we do not overshoot excels limits


src/main/resources/web_en.properties
```diff
@@ -5,6 +5,7 @@ BasicController.objectDeleted = The object has been deleted
 BasicController.toDashboard = Back to the Dashboard
 BasicController.unknownObject = The data object was not found.
 ControllerDispatcher.invalidCSRFToken = Outdated or invalid security token. Please retry your last action.
+ExcelExport.maxRowsReached = The data was trimed, because it's to long for excel
```
fhaScireum:
```suggestion
ExcelExport.maxRowsReached = The data was trimed, because it's too long for excel
```

src/main/resources/web_en.properties
```diff
@@ -5,6 +5,7 @@ BasicController.objectDeleted = The object has been deleted
 BasicController.toDashboard = Back to the Dashboard
 BasicController.unknownObject = The data object was not found.
 ControllerDispatcher.invalidCSRFToken = Outdated or invalid security token. Please retry your last action.
+ExcelExport.maxRowsReached = The data was trimed, because it's too long for excel
```
idlira:
```suggestion
ExcelExport.maxRowsReached = Data output trimmed. Excel exports limited to 1.000.000 rows.
```

src/main/resources/web_de.properties
```diff
@@ -5,6 +5,7 @@ BasicController.objectDeleted = Das Objekt wurde gelöscht
 BasicController.toDashboard = Zurück zur Startseite
 BasicController.unknownObject = Das Datenobjekt wurde nicht gefunden.
 ControllerDispatcher.invalidCSRFToken = Abgelaufenes oder invalides Sicherheitstoken. Führen Sie Ihre letzte Aktion bitte erneut aus.
+ExcelExport.maxRowsReached = Die Daten wurden gekürzt, da zu lang für Excel
```
idlira:
```suggestion
ExcelExport.maxRowsReached = Datenausgabe getrimmt. Excel-Exporte sind auf eine Million Zeilen begrenzt.
```

src/main/java/sirius/web/data/ExcelExport.java
```diff
@@ -52,12 +52,15 @@
...
 import java.util.HashSet;
 import java.util.Set;
+import java.util.function.Consumer;
 
 /**
  * Generates an Excel file which can be sent as a response for a {@link sirius.web.http.WebContext}
  */
 public class ExcelExport {
 
+    public static final int MAX_NUM_ROWS = 1_000_000;
```
andyHa: javadoc

src/main/java/sirius/web/data/ExcelExport.java
```diff
@@ -399,6 +404,17 @@ public ExcelExport addRow(Object... row) {
      * @return the export itself for fluent method calls
      */
     public ExcelExport addRowAsList(Collection<?> row) {
+        if (rows > MAX_NUM_ROWS) {
+            return this;
+        }
+        if (rows == MAX_NUM_ROWS) {
+            Row r = currentSheet.createRow(rows++);
+            addCell(r, getMaxRowsReachedMessage(), 0, normalStyle);
```
andyHa: shouldn't we skip this if the message is null?

src/main/java/sirius/web/data/ExcelExport.java
andyHa: shouldn't we also have one for XLS?

src/main/java/sirius/web/data/ExcelExport.java
```diff
@@ -491,4 +507,16 @@ private boolean isNumeric(Object data) {
 
         return (data instanceof Amount) && ((Amount) data).isFilled();
     }
+
+    public String getMaxRowsReachedMessage() {
+        return Strings.firstFilled(NLS.smartGet(maxRowsReachedMessage), NLS.get("ExcelExport.maxRowsReached"));
```
andyHa: nope, permit to leave empty

src/main/java/sirius/web/data/ExcelExport.java
```diff
@@ -491,4 +507,16 @@ private boolean isNumeric(Object data) {
...
+    }
+
+    public void setMaxRowsReachedMessage(String maxRowsReachedMessage) {
```
andyHa: withMaxRows.... +javadoc +return this

src/main/java/sirius/web/data/ExcelExport.java
```diff
@@ -491,4 +507,16 @@ private boolean isNumeric(Object data) {
...
+        this.maxRowsReachedMessage = maxRowsReachedMessage;
+    }
+
+    public void setMaxRowsReachedHandler(Consumer<String> maxRowsReachedHandler) {
```
andyHa: withMaxRows.... +javadoc +return this

src/main/java/sirius/web/data/ExcelExport.java
```diff
@@ -399,6 +404,17 @@ public ExcelExport addRow(Object... row) {
      * @return the export itself for fluent method calls
      */
     public ExcelExport addRowAsList(Collection<?> row) {
+        if (rows > MAX_NUM_ROWS) {
```
andyHa: shouldn't we zero this once the sheet is changed or is this a global limit?

src/test/java/sirius/web/data/ExcelExportSpec.groovy
```diff
@@ -56,4 +58,38 @@ class ExcelExportSpec extends BaseSpecification {
         cleanup:
         Files.delete(testFile)
     }
+
+    def "only allow 1 million in a excel sheet rows"() {
```
andyHa: scope = Nightly - this sounds time consuming and memory hungry

src/test/java/sirius/web/data/ExcelExportSpec.groovy
```diff
@@ -56,4 +58,38 @@ class ExcelExportSpec extends BaseSpecification {
...
+        export.writeToStream(new FileOutputStream(testFile))
+        then:
+        sheetWithError.toString() == "Sheet0"
+        and:
+        Counter lineCounter = new Counter()
+        LineBasedProcessor.create(testFile.getName(), new FileInputStream(testFile))
+                          .run({
+                                   lineNum, row ->
+                                       lineCounter.inc()
+                                       assert lineNum <= ExcelExport.MAX_NUM_ROWS + 1
```
andyHa: assert? is that the spock way to do it? these might be disabled in the JVM, right?

src/main/java/sirius/web/data/ExcelExport.java
welzfabian: we do in the `createSheet`-method

src/test/java/sirius/web/data/ExcelExportSpec.groovy
welzfabian: ok, but are there nightly builds for sirius-web and does anybody ever check the results of those?

src/test/java/sirius/web/data/ExcelExportSpec.groovy
welzfabian: well I think so. it does not work without the assert.

here is an article describing what code is actually generated by spock, but the example is with an `if` an not with a closure
https://e.printstacktrace.blog/spock-assertion-inside-if-statement-doesnt-work-why/

src/main/java/sirius/web/data/ExcelExport.java
welzfabian: this limit applies for xls and xlsx

src/test/java/sirius/web/data/ExcelExportSpec.groovy
welzfabian: this test completes in about 25 seconds btw

src/test/java/sirius/web/data/ExcelExportSpec.groovy
welzfabian: so the `assert` is just a hint for Spock to generate the default assertion code also generated by regular spock assertions in the `then:` block

src/main/java/sirius/web/data/ExcelExport.java
welzfabian: oh okay XLS has a different limit, i'll add one 👌

src/test/java/sirius/web/data/ExcelExportSpec.groovy
andyHa: yes it does. yes they are reported. 25s = doubles the time of the current build.

src/test/java/sirius/web/data/ExcelExportSpec.groovy
andyHa: th,x good to know

src/main/java/sirius/web/data/ExcelExport.java
```diff
@@ -399,6 +402,20 @@ public ExcelExport addRow(Object... row) {
...
+            if (Strings.isFilled(getMaxRowsReachedMessage())) {
+                Row r = currentSheet.createRow(rows);
+                addCell(r, getMaxRowsReachedMessage(), 0, normalStyle);
+            }
+            rows++;
+            if (maxRowsReachedHandler != null) {
+                maxRowsReachedHandler.accept(currentSheet.getSheetName());
+            }
+            return this;
+        }
```
andyHa: shoulden the whole block be in the if below or shouldn't this return immediatelly?

src/main/java/sirius/web/data/ExcelExport.java
```diff
@@ -491,4 +508,34 @@ private boolean isNumeric(Object data) {
 
         return (data instanceof Amount) && ((Amount) data).isFilled();
     }
+
+    public String getMaxRowsReachedMessage() {
```
andyHa: does this need to be public? also, this isn't a real getter - determineMaxRowsMessage or the like?

src/main/java/sirius/web/data/ExcelExport.java
```diff
@@ -491,4 +508,34 @@ private boolean isNumeric(Object data) {
...
+        return NLS.smartGet(maxRowsReachedMessage);
+    }
+
+    /**
+     * The message to append at the end of an excel sheet which exceeds the maximum number of allowed rows.
+     * <p>
+     * When this is left empty, no message will be appended.
```
andyHa: add a hint that this will be {@link NLS#smartGet smart translated}

src/test/java/sirius/web/data/ExcelExportSpec.groovy
```diff
@@ -9,10 +9,14 @@
 package sirius.web.data
 
 import sirius.kernel.BaseSpecification
+import sirius.kernel.Scope
 import sirius.kernel.commons.Files
+import sirius.kernel.health.Counter
 
 class ExcelExportSpec extends BaseSpecification {
 
+    public static final int XLSX_MAX_ROWS = 0x100000
```
andyHa: private?

src/test/java/sirius/web/data/ExcelExportSpec.groovy
```diff
@@ -56,4 +60,40 @@ class ExcelExportSpec extends BaseSpecification {
...
+                                       } else {
+                                           assert row.at(0) == "Max rows reached."
+                                           assert lineNum == XLSX_MAX_ROWS
+                                       }
+                               },
+                               { e -> false })
+        lineCounter.getCount() == XLSX_MAX_ROWS
+        cleanup:
+        Files.delete(testFile)
+    }
```
andyHa: add another for xls?

src/main/java/sirius/web/data/ExcelExport.java
welzfabian: I dont think so. Why do you think so?

src/main/java/sirius/web/data/ExcelExport.java
```diff
@@ -399,14 +405,31 @@ public ExcelExport addRow(Object... row) {
      * @return the export itself for fluent method calls
      */
     public ExcelExport addRowAsList(Collection<?> row) {
-        if (row != null) {
-            maxCols = Math.max(maxCols, row.size());
-            int idx = 0;
-            Row r = currentSheet.createRow(rows++);
-            for (Object entry : row) {
-                addCell(r, entry, idx++, getCellStyleForObject(entry));
+        if (row == null) {
```
andyHa: if (row == null || rowLimitHit()) {
return this;
}

src/main/java/sirius/web/data/ExcelExport.java
```diff
@@ -399,14 +405,31 @@ public ExcelExport addRow(Object... row) {
...
+            return this;
+        }
+        if (rows > workbook.getSpreadsheetVersion().getMaxRows() - 1) {
+            return this;
```
andyHa: in new method -> return true;

src/main/java/sirius/web/data/ExcelExport.java
```diff
@@ -399,14 +405,31 @@ public ExcelExport addRow(Object... row) {
...
+        if (rows == workbook.getSpreadsheetVersion().getMaxRows() - 1) {
+            if (Strings.isFilled(determineMaxRowsReachedMessage())) {
+                Row r = currentSheet.createRow(rows);
+                addCell(r, determineMaxRowsReachedMessage(), 0, normalStyle);
             }
+            rows++;
+            if (maxRowsReachedHandler != null) {
+                maxRowsReachedHandler.accept(currentSheet.getSheetName());
+            }
+            return this;
```
andyHa: in new method -> return true;

src/main/java/sirius/web/data/ExcelExport.java
andyHa: has been moved

PR #679, opened 2019-10-24
Differentiate a login with a user password from one using an API token.

andyHa:
In the second case we must not report an error as the token might be
an SSO token which is computed entirely differently...


src/main/java/sirius/web/security/GenericUserManager.java
```diff
@@ -343,23 +343,43 @@ protected void log(String pattern, Object... params) {
                              Strings.apply(pattern, params));
     }
 
-    /*
+    /**
      * Tries to perform a login using "user" and "password".
+     *
+     * @param ctx the request the read the parameters from
```
sabieber: Refactor `ctx` to `webContext` please

PR #692, opened 2020-01-08
Creates a message level enum

tbiScireum:
and uses the enum instead of string constants in the message class
- Fixes: SE-9177

### DIFF
--- a/src/main/java/sirius/web/controller/MessageLevel.java
+++ b/src/main/java/sirius/web/controller/MessageLevel.java
@@ -26,10 +26,10 @@ public enum MessageLevel {
      */
     PROBLEM("alert-danger");
 
-    private String style;
+    private String cssClass;
 
-    MessageLevel(String style) {
-        this.style = style;
+    MessageLevel(String cssClass) {
+        this.cssClass = cssClass;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/main/resources/default/templates/wondergem/page-messages.html.pasta
```diff
@@ -1,7 +1,7 @@
 <div class="row">
     <div class="col-md-12" id="messageBox">
         <i:for type="sirius.web.controller.Message" var="msg" items="user.getMessages()">
-            <div class="alert @msg.getType()">
+            <div class="alert @msg.getType().getStyle()">
```
meg1502: Wouldn't it be better to also overwrite toString of the Type enum to return getStyle? 
This make this change to not break as much as '@msg.getType()' would also be valid in templates

src/main/java/sirius/web/controller/MessageLevel.java
```diff
@@ -0,0 +1,43 @@
...
+     * Declares a message as warning.
+     */
+    WARNING("alert-warning"),
+
+    /**
+     * Declares a message as error.
+     */
+    PROBLEM("alert-danger");
+
+    private String style;
```
meg1502: rename into styleClass or cssClass ?

PR #693, opened 2020-01-09
Adds Promise support for IE

PhilHup:
Adds support for promises in internet explorer. Uses https://github.com/stefanpenner/es6-promise

-Fixes: SIRI-147


src/main/resources/default/assets/wondergem/wondergem.js.pasta
```diff
@@ -21,3 +21,7 @@
 <i:include name="/assets/wondergem/bootstrap-tokenfield/bootstrap-tokenfield.js"/>
 
 <i:include name="/assets/wondergem/autocomplete/multiselect.js"/>
+
+<i:if test="@call.getUserAgent().contains('Trident')">
```
andyHa: nope, this will most definitely be cached by an upstream proxy - move this to the page template and load on demand there

PhilHup: moved to w:page

PR #697, opened 2020-01-16
Uses the correct mime type for JSONP responses 

meg1502:
SIRI-156

 See: https://blog.abelotech.com/posts/json-jsonp-correct-mime-type/


src/main/java/sirius/web/http/Response.java
```diff
@@ -1324,13 +1324,15 @@ public void tunnel(final String url,
      * <p>
      * By default, caching will be disabled. If the generated JSON is small enough, it will be transmitted in
      * one go. Otherwise a chunked response will be sent.
+     * If a callback parameter is given in the request, the output will automatically be boxed into that function as JSONP.
```
andyHa:
```suggestion
     * <p>
     * If a callback parameter is given in the request, the output will automatically be boxed into that function as JSONP.
```

PR #702, opened 2020-02-28
Removes multiselect optional parameter

meg1502:
As the only thing it did was not transparent

Fixes: SIRI-186


meg1502:
check if arg can be deprecated

PR #706, opened 2020-03-21
Creates a new method to create post json calls

tbiScireum:
Lots of api endpoints don't accept json data with a get request
- Fixes: SE-9304

### DIFF
--- a/src/main/java/sirius/web/services/JSONCall.java
+++ b/src/main/java/sirius/web/services/JSONCall.java
@@ -57,6 +57,22 @@ public class JSONCall {
     }
 
     /**
+     * Creates a new JSONCall as post request for the given url.
+     *
+     * @param url         the target URL to call
+     * @param contentType the Content-Type to use
+     * @return a new instance to perform the xml call
+     * @throws IOException in case of an IO error
+     */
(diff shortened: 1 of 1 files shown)


src/main/java/sirius/web/services/JSONCall.java
```diff
@@ -56,6 +56,22 @@ public static JSONCall to(URL url, String contentType) throws IOException {
         return result;
     }
 
+    /**
+     * Creates a new JSONCall as post request for the given url.
+     *
+     * @param url         the target URL to call
+     * @param contentType the Content-Type to use
+     * @return a new instance to perform the xml call
```
ymo-sci:
```suggestion
     * @return a new instance to perform the json call
```

src/main/java/sirius/web/services/JSONCall.java
```diff
@@ -56,6 +56,22 @@ public static JSONCall to(URL url, String contentType) throws IOException {
...
+     * @throws IOException in case of an IO error
+     */
+    public static JSONCall post(URL url, String contentType) throws IOException {
```
andyHa: why not just markAsPost() to re-use the c'tors above which also set a valid content type

PR #707, opened 2020-03-22
Adds support for propagating generic types down a call chain.

andyHa:
Especially for entities which have a reference like
`SQLEntityRef<OtherEntity> ref`;

Tagliatelle now knows that `entity.getRef().fetchValue()` is of type
`OtherEntity`.


src/test/java/sirius/tagliatelle/CompilerSpec.groovy
```diff
@@ -268,7 +281,10 @@ class CompilerSpec extends BaseSpecification {
         then:
         errors.size() == 1
         errors.get(0).getSeverity() == ParseError.Severity.WARNING
-        errors.get(0).getMessage().contains("The method sirius.tagliatelle.TestObject.deprecatedMethod is marked as deprecated")
+        errors.
```
sabieber: Auto formatting gone wrong :(

PR #708, opened 2020-03-24
Further improves the type interference capabilities of the Tagliatell…

andyHa:
…e compiler.

We now can deduce that <H> UserContext.getHelper(Class<H>) will return
an instance of the given class. More commonly we can resolve type variables
into classes if a given constant class is present which binds to the same type
variable.

### DIFF
--- a/src/main/java/sirius/tagliatelle/expression/MethodCall.java
+++ b/src/main/java/sirius/tagliatelle/expression/MethodCall.java
@@ -198,16 +198,101 @@ public class MethodCall extends Call {
     @Override
     public Type getGenericType() {
         if (method.getGenericReturnType() instanceof TypeVariable) {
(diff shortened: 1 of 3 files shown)


src/main/java/sirius/tagliatelle/expression/MethodCall.java
```diff
@@ -198,16 +198,101 @@ private boolean isMatchingVarargArray(int index, Object param) {
...
+            if (resolvedType != null) {
+                return resolvedType;
+            }
+        }
+
+        return null;
+    }
+
+    private Type tryResolveAsConstantClassParameter(String variableName, int parameterIndex) {
+        // Ensore that the parameter is of type Class<X> - abort otherwise
```
idlira:
```suggestion
        // Ensure that the parameter is of type Class<X> - abort otherwise
```

PR #714, opened 2020-04-07
Detects (and warns about) pointless transformer casts.

andyHa:


andyHa:
Even when using transformers a call like X.as(X.class) is pointless.
Also this could otherwise cover follow up errors as previously
(before Tagliatelle supported generic propagation), we operated
on Objects which were often casted to Entities of sirius-db. Now
we know it's an entity (which is also Transformable) so the unneccessarry
cast isn't discovered but executed a pointless Transformable.as().

src/main/java/sirius/tagliatelle/Tagliatelle.java
```diff
@@ -418,8 +418,8 @@ private void writeWarningsToLog(CompilationContext compilationContext) {
         message.append("Warnings when compiling ")
                .append(compilationContext.getTemplate().getShortName())
                .append(":\n");
-        compilationContext.getErrors().forEach(message::append);
-        message.append(" - Template: ");
+        compilationContext.getErrors().forEach(error -> message.append(error).append("\n\n"));
```
idlira, line 421: Why 2 line breaks?

andyHa: readability - could also be done with a single break - suggestions?

bwiedmann: Readability is better with two line breaks 👍

PR #716, opened 2020-04-13
Uses MethodHandles in favor of reflection.

andyHa:


src/main/java/sirius/web/controller/ControllerDispatcher.java
```diff
@@ -47,9 +47,15 @@
 public class ControllerDispatcher implements WebDispatcher {
 
     protected static final Log LOG = Log.get("controller");
-
     private static final String SYSTEM_MVC = "MVC";
 
+    /**
+     * Contains the pattern defined by the route bing matched.
```
idlira, line 53: bing? or should it be bind?

src/main/java/sirius/web/controller/Route.java
```diff
@@ -336,4 +346,43 @@ protected boolean isJSONCall() {
     protected String getPattern() {
         return pattern.toString();
     }
+
+    /**
+     * Invokes the route with the given parameters.
+     *
+     * @param params the parameters to supply
+     * @return the result of the route. This will most probably be null (as routes are mostly void), but might be
+     * a {@link Promise} which is used for JSON calls to indicate the the response will be completed in another thread.
```
idlira, line 355:
```suggestion
     * a {@link Promise} which is used for JSON calls to indicate that the response will be completed in another thread.
```

src/main/java/sirius/web/controller/ControllerDispatcher.java
sabieber, line 53:
```suggestion
     * Contains the pattern defined by the route being matched.
```

PR #717, opened 2020-04-14
Permits to attach completion callbacks in WebContexts.

andyHa:
Sometimes we want to perform some tasks (i.e. statistics)
if a request was completely handled - this can be used by
attaching handlers on the newly created promise.

### DIFF
--- a/src/main/java/sirius/web/http/Response.java
+++ b/src/main/java/sirius/web/http/Response.java
@@ -135,7 +135,7 @@ public class Response {
     private HttpHeaders headers;
 
     /*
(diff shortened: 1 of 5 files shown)


src/main/java/sirius/web/http/Response.java
```diff
@@ -134,6 +134,11 @@
      */
     private HttpHeaders headers;
 
+    /*
+     * Stores the the effective response code.
```
sabieber, line 138:
```suggestion
     * Stores the effective response code.
```

PR #718, opened 2020-04-15
SE-9538 Allows rendering interleaved 2 of 5 barcodes in PDFs

jmuscireum:
The code will be prepended with a 0 to ensure an even number of digits.

### DIFF
--- a/src/main/java/sirius/web/templates/ImageReplacedElementFactory.java
+++ b/src/main/java/sirius/web/templates/ImageReplacedElementFactory.java
@@ -223,7 +223,7 @@ class ImageReplacedElementFactory extends ITextReplacedElementFactory {
         if (code instanceof BarcodeInter25) {
             int length = BarcodeInter25.keepNumbers(src).length();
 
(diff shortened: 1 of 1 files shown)


src/main/java/sirius/web/templates/ImageReplacedElementFactory.java
```diff
@@ -145,29 +156,87 @@ private ReplacedElement createImageForBarcode(String type, String src, int cssWi
...
+     *
+     * @param code the instance of the barcode
+     * @param src  the code from the src attribute
+     * @return the padded code, or the original code if padding was not needed
+     */
+    private String padCodeIfNecessary(Barcode code, String src) {
+        if (code instanceof BarcodeInter25) {
+            int length = BarcodeInter25.keepNumbers(src).length();
+
+            // Length is unenven and no checksum will be added
```
jmuscireum, line 226:
```suggestion
            // Length is uneven and no checksum will be added
```

PR #719, opened 2020-04-20
BREAKING CHANGE: Uses a loop instead of a recursive call

andyHa:
This will simplify the internal handling of requests and also significantly reduce the length of the stack traces.

### DIFF
--- a/src/test/java/sirius/web/dispatch/TestDispatcher.java
+++ b/src/test/java/sirius/web/dispatch/TestDispatcher.java
@@ -26,8 +26,7 @@ public class TestDispatcher implements WebDispatcher {
     }
 
     @Override
-    public void dispatch(WebContext ctx, Consumer<WebContext> startOfPipeline, Consumer<WebContext> nextStage)
-            throws Exception {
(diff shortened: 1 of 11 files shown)


src/test/java/sirius/web/dispatch/TestDispatcher.java
```diff
@@ -26,28 +26,22 @@ public int getPriority() {
     }
 
     @Override
-    public void dispatch(WebContext ctx, Consumer<WebContext> startOfPipeline, Consumer<WebContext> nextStage)
-            throws Exception {
+    public DispatchDecision dispatch(WebContext ctx) throws Exception {
         if ("/large-blocking-calls".equalsIgnoreCase(ctx.getRequestedURI())) {
             // See WebServerSepc->"Invoke /large-blocking-calls with GET" to the appropriate test and explanation...
```
bwiedmann, line 31:
```suggestion
            // See WebServerSpec->"Invoke /large-blocking-calls with GET" to the appropriate test and explanation...
```

PR #747, opened 2020-09-03
Version bump

qw3ry:
Required changes for the next kernel version

### DIFF
--- a/src/test/java/sirius/web/service/JsonOutputSpec.groovy
+++ b/src/test/java/sirius/web/service/JsonOutputSpec.groovy
@@ -1,0 +1,33 @@
+/*
+ * Made with all the love in the world
+ * by scireum in Remshalden, Germany
+ *
+ * Copyright by scireum GmbH
+ * http://www.scireum.de - info@scireum.de
+ */
+
+package sirius.web.service
+
+import groovy.json.JsonSlurper
+import sirius.kernel.BaseSpecification
+import sirius.kernel.xml.Attribute
+import sirius.web.services.JSONStructuredOutput
+
+class JsonOutputSpec extends BaseSpecification {
+    def "json output uses attributes"() {
+        given:
+        OutputStream os = new ByteArrayOutputStream();
+        JSONStructuredOutput out = new JSONStructuredOutput(os, null, "UTF8");
+        when:
+        out.beginResult("test");
+        out.beginObject("1", Attribute.set("a", "b"))
+        out.endObject()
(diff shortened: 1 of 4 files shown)


src/test/java/sirius/web/service/JsonOutputSpec.groovy
```diff
@@ -0,0 +1,33 @@
...
+        given:
+        OutputStream os = new ByteArrayOutputStream();
+        JSONStructuredOutput out = new JSONStructuredOutput(os, null, "UTF8");
+        when:
+        out.beginResult("test");
+        out.beginObject("1", Attribute.set("a", "b"))
+        out.endObject()
+        out.endResult();
+        then:
+        compareJsonString(os.toString(), """{"1":{"a":"b"}}""");
```
meg1502, line 27:
```suggestion
os.toString() == """{"1":{"a":"b"}}"""
```

shouldn't this be sufficient also ?

meg1502 (approved):

src/test/java/sirius/web/service/JsonOutputSpec.groovy
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    private def compareJsonString(String str1, String str2) {
```
andyHa, line 30: not needed?

andyHa (changes requested):
...

sabieber (approved):

bwiedmann (approved):

mkeckmkeck (approved):

src/test/java/sirius/web/service/JsonOutputSpec.groovy
qw3ry: It is, indeed. But I there are other correct serializations of this JSON, so the test might break when the JSON is serialized differently. Would it be reasonable to do a simple string equality here?

src/test/java/sirius/web/service/JsonOutputSpec.groovy
qw3ry: Depends on https://github.com/scireum/sirius-web/pull/747#discussion_r482841806

src/test/java/sirius/web/service/JsonOutputSpec.groovy
meg1502: I don't really know, i thought we wanted the JSON to always look like this :)

PR #758, opened 2020-10-22
Provides the first steps for the new Tycho UI

andyHa:


src/main/java/sirius/web/templates/DefaultGlobalContextExtender.java
```diff
@@ -61,6 +61,7 @@ public void collectTemplate(Collector parameterCollector) {
         parameterCollector.collect("lang", NLS.getCurrentLang());
         parameterCollector.collect("contentHelper", ContentHelper.INSTANCE);
         parameterCollector.collect("wondergemRoot", wondergemRoot, String.class);
+        parameterCollector.collect("tychoRoot", wondergemRoot, String.class);
```
sabieber, line 64: Should this be the same value as wondergem root?

sabieber (changes requested):

src/main/resources/default/assets/tycho/lib.css.pasta
```diff
@@ -0,0 +1 @@
+___include("/assets/tycho/libs/bootstrap/css/bootstrap.css")
```
jmuscireum, line 1: Should be instead use the min versions for css and js?

bwiedmann (approved):
👍 🛫

src/main/java/sirius/web/templates/DefaultGlobalContextExtender.java
andyHa: yop for backwards compatibility

sabieber (approved):

idlira (approved):

bwiedmann (approved):

PR #767, opened 2020-11-25
Permits to split an authentication scope into "sub scopes".

andyHa:
This permits to grand users only access to partial areas of an application.
E.g. this could be used to create "API only" users. Also this will be used
by sirius-biz to create "FTP only" users.

### DIFF
--- a/src/main/java/sirius/web/controller/Route.java
+++ b/src/main/java/sirius/web/controller/Route.java
@@ -61,6 +61,7 @@ public class Route {
     private boolean preDispatchable;
     private boolean jsonCall;
     private Set<String> permissions = null;
+    private String subScope;
 
     /**
      * Compiles a method defined by a {@link Controller}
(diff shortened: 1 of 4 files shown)


src/main/java/sirius/web/controller/Route.java
```diff
@@ -285,6 +296,14 @@ protected String checkAuth(Supplier<UserInfo> user) {
         if (permissions == null) {
             return null;
         }
+
+        // Note that we intentionally only check and enforce the sub scope if the
+        // route is "not public accessible". This way, we to not have to fetch the
```
sabieber, line 301:
```suggestion
        // route is "not public accessible". This way, we do not have to fetch the
```

sabieber (changes requested):

idlira (approved):

mkeckmkeck (approved):

PR #768, opened 2020-12-07
SE-10166: Added system env macros - can determine current system environment

marvinwoop:
To determine the current system environment in templates, we need this SystemHelper.


src/main/java/sirius/web/util/SystemHelper.java
```diff
@@ -0,0 +1,74 @@
...
+        }
+    }
+
+    /**
+     * Returns the value of the given environment variable.
+     *
+     * @param varName the environment variable to look up
+     * @return the value of the environment variable
+     */
+    public String getEnvironmentVariable(String varName) {
```
andyHa, line 53: this is VERY DANGEROUS! why is that needed?

src/main/java/sirius/web/util/SystemHelper.java
```diff
@@ -0,0 +1,74 @@
...
+import sirius.web.security.HelperFactory;
+import sirius.web.security.ScopeInfo;
+
+import javax.annotation.Nonnull;
+
+/**
+ * Provides static system functionality and system environment detection based on {@link sirius.kernel.Setup.Mode} as
+ * helper for in-template usage.
+ */
+public class SystemHelper {
```
andyHa, line 23: we could also encapsulate all this in a isProd() / isStaging() / isDev() macro which is probably easier to use anyway... (?)

andyHa (changes requested):

src/main/java/sirius/web/util/SystemHelper.java
marvinwoop: This is not really needed and will therefore be removed.

src/main/java/sirius/web/util/SystemHelper.java
marvinwoop: This will be rebuilt into a macro as suggested.

src/main/java/sirius/tagliatelle/macros/IsProductiveMacro.java
```diff
@@ -0,0 +1,58 @@
...
+import sirius.tagliatelle.rendering.LocalRenderContext;
+
+import javax.annotation.Nonnull;
+import java.util.List;
+
+/**
+ * This macro determines if the current system environment is in production {@link sirius.kernel.Setup.Mode}.
+ */
+@Register
+public class IsProductiveMacro implements Macro {
```
andyHa, line 25: extract common superclass? so that there is only a single getMode() left to implement...

src/main/java/sirius/tagliatelle/macros/IsProductiveMacro.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    @Override
+    public Object eval(LocalRenderContext ctx, Expression[] args) {
+        return Sirius.isProd();
+    }
+
+    @Override
+    public boolean isConstant(Expression[] args) {
+        return false;
```
andyHa, line 45: this can't ever change -> true

marvinwoop: 👍  changed that as suggested in base class.

andyHa (approved):

mkeckmkeck (approved):

src/main/java/sirius/tagliatelle/macros/IsProductiveMacro.java
```diff
@@ -0,0 +1,39 @@
...
+
+    @Override
+    public String getDescription() {
+        return "Returns true if system environment in production mode, false otherwise.";
+    }
+
+    @Nonnull
+    @Override
+    public String getName() {
+        return "isProductive";
```
sabieber, line 37: Why use a different naming here that the sirius version and the enum value?

sabieber (changes requested):

src/main/java/sirius/tagliatelle/macros/IsProductiveMacro.java
marvinwoop: thanks, changed the naming. 👍

sabieber (approved):

PR #776, opened 2020-12-18
Simplifies creating helpers.

andyHa:
Removes the ability to reference a helper by its name.

This is essentially a deprecated approach which was left over
from the JSF days. We actually never did reference helpers by
their name and we probably will never, as we'd instantly loose
all benefits of Tagliatelle and its static compilation capabilities.

Note that now that HelperFactories are quite trivial as these are
only glorified constructors, we also permit to instantiate a helper
via its constructor (which might take ScopeInfo as parameter).

Note that a factory can still be used and will always be preferred
as we might use this to overwrite / change helpers in customizations.

Also note, that while refactoring the new instance creation, we also
added support for circular references across helpers and also made
sure that each helper only exists once per scope. (Previously the
@Helper annotation created a copy instead of referencing the real
helper).

### DIFF
--- a/src/main/java/sirius/web/security/ScopeInfo.java
+++ b/src/main/java/sirius/web/security/ScopeInfo.java
@@ -79,7 +79,7 @@ public class ScopeInfo extends Composable {
     private static List<HelperFactory<?>> factories;
 
     @Part
-    private static GlobalContext ctx;
+    private static GlobalContext globalContext;
 
     /**
      * Creates a new <tt>ScopeInfo</tt> with the given parameters.
@@ -279,7 +279,7 @@ public class ScopeInfo extends Composable {
     }
 
     private Object populateHelper(Class<?> type, Object helper, Map<Class<?>, Object> localContext) {
-        ctx.wire(helper);
+        globalContext.wire(helper);
         fillConfig(helper);
 
         // Note that we deliberately make the helper visible in the local context here so that "fillFriends" is
@@ -468,8 +468,8 @@ public class ScopeInfo extends Composable {
     public UserManager getUserManager() {
         if (userManager == null) {
             Extension ext = Sirius.getSettings().getExtension("security.scopes", getScopeType());
-            userManager = ctx.getPart(ext.get("manager").asString("public"), UserManagerFactory.class)
-                             .createManager(this, ext);
+            userManager = globalContext.getPart(ext.get("manager").asString("public"), UserManagerFactory.class)
+                                       .createManager(this, ext);
         }
 
         return userManager;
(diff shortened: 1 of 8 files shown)


src/main/java/sirius/web/security/ScopeInfo.java
```diff
@@ -208,124 +207,131 @@ public boolean is(@Nonnull Class<?> type) {
...
     }
 
-    private Object makeHelper(HelperFactory<?> factory) {
-        return makeHelper(factory, 0);
-    }
-
-    private Object makeHelper(HelperFactory<?> factory, int circuitBreaker) {
-        circuitBreaker++;
+    private Object populateHelper(Class<?> type, Object helper, Map<Class<?>, Object> localContext) {
+        ctx.wire(helper);
```
sabieber, line 282: could we also call this `context` or `globalContext` during this refactoring?

src/main/java/sirius/web/security/ScopeInfo.java
```diff
@@ -208,124 +207,131 @@ public boolean is(@Nonnull Class<?> type) {
...
+        fillConfig(helper);
 
-        Object result = factory.make(this);
+        // Note that we deliberately make the helper visible in the local context here so that "fillFriends" is
+        // catable of handling circular references...
```
idlira, line 286:
```suggestion
        // capable of handling circular references...
```

src/main/java/sirius/web/security/ScopeInfo.java
```diff
@@ -208,124 +207,131 @@ public boolean is(@Nonnull Class<?> type) {
...
-
-        return result;
+        return helper;
     }
 
     private void fillConfig(Object result) {
         Settings scopeSettings = getSettings();
         Reflection.getAllFields(result.getClass())
                   .stream()
                   .filter(f -> f.isAnnotationPresent(HelperConfig.class))
```
sabieber, line 297:
```suggestion
                  .filter(field -> field.isAnnotationPresent(HelperConfig.class))
```

src/main/java/sirius/web/security/ScopeInfo.java
```diff
@@ -208,124 +207,131 @@ public boolean is(@Nonnull Class<?> type) {
...
-                  .forEach(f -> scopeSettings.injectValueFromConfig(result,
-                                                                    f,
-                                                                    f.getAnnotation(HelperConfig.class).value()));
+                  .forEach(f -> {
+                      try {
+                          scopeSettings.injectValueFromConfig(result, f, f.getAnnotation(HelperConfig.class).value());
```
sabieber, line 300:
```suggestion
                  .forEach(field -> {
                      try {
                          scopeSettings.injectValueFromConfig(result, field, field.getAnnotation(HelperConfig.class).value());
```

sabieber (changes requested):

idlira (approved):
Typo fix

src/main/java/sirius/web/security/ScopeInfo.java
andyHa: done

sabieber (approved):

jmuscireum (approved):

PR #779, opened 2021-01-05
Logs heaps of additional timings to trace slow tunnelings.

andyHa:


idlira (approved):

bwiedmann (approved):

src/main/java/sirius/web/http/TunnelHandler.java
```diff
@@ -62,6 +64,11 @@
     private final IntConsumer failureHandler;
     private final CallContext cc;
     private final Watch watch;
+    private volatile long timeToDns;
+    private volatile long timeToConnectAttept;
```
mkeckmkeck, line 68:
```suggestion
    private volatile long timeToConnectAttempt;
```

mkeckmkeck (approved):

src/main/java/sirius/web/http/TunnelHandler.java
andyHa: Will be fixed with the update to the next async http client release.

PR #788, opened 2021-01-14
Fix: Mail Server Trust

jakobvogel:
- Trusts CA-signed mail servers by default


andyHa:
Maybe also add a comment in the component-web.conf that once the option is used, ca certs are no longer trusted

andyHa (approved):

marvinwoop (approved):
evtl.

mkeckmkeck (approved):

PR #792, opened 2021-01-19
Fixes an internal overflow of the SharedConstantPool.

andyHa:
We kept all referenced MethodHandles in the shared constant pool. Now as MethodHandle doesn't support "equals", each compilation increased the pool size more and more.

Now that we store bytecodes as 32 bit ints, where the upper 16 bits index the OpCode and the lower 16 bit can represent an index into the constant table or the shared constant pool.

As we didn't validate these indices and also didn't impose any limits on the shared constant pool, the index bits which are "OR"'ed into the opcode bits, effectively changed the op code which was eventually executed. This lead invalid sequences of op codes and thus to crashes of Noodle (with more or less funny causes).

The fixes in this PR all try to fix and prevent errors of this class in the future. As the impact of this error is pretty high,
several distinct fixes and improvements have been applied.

### DIFF
--- a/src/main/java/sirius/pasta/noodle/Invocation.java
+++ b/src/main/java/sirius/pasta/noodle/Invocation.java
@@ -13,8 +13,10 @@ import sirius.kernel.commons.Explain;
 import sirius.kernel.commons.Strings;
 import sirius.kernel.commons.Value;
 import sirius.kernel.di.transformers.Transformable;
+import sirius.kernel.health.Exceptions;
 import sirius.kernel.health.HandledException;
 import sirius.kernel.nls.NLS;
+import sirius.pasta.Pasta;
 import sirius.pasta.noodle.macros.Macro;
 import sirius.web.security.UserContext;
 
@@ -273,11 +275,35 @@ public class Invocation {
         StringBuilder result = new StringBuilder(" Line:\n");
         result.append(offendingLine);
         result.append("\n");
-        for (int i = 1; i < position.getPos(); i++) {
+        for (int i = 1; i < position.getPos() - 1; i++) {
             result.append(" ");
         }
         result.append("^\n");
         return result.toString();
+    }
+
+    /**
+     * Creates an exception which is used when Noodle detects an inconsistent or invalid internal state.
+     * <p>
+     * These states should normally not be reached as the compiler should only emit valid bytecodes. Therefore
+     * we provide quite some excessivle loggig here.
+     *
+     * @param message the message to log
+     * @return an exception with a short and concise message of what happened. Also a long and elaborate message will
+     * be logged which might provide further insight.
+     */
+    private IllegalArgumentException vmError(String message) {
+        // Create a full blown exception for the logs...
+        IllegalStateException internalError = new IllegalStateException(Strings.apply(
+                "A Noodle VM error occurred: %s%n%nIP: %s%nStack: %s%nCall:%n%n%s",
+                message,
+                instructionPointer,
+                stack,
+                this.compiledMethod.disassemble()));
+        Exceptions.handle(Pasta.LOG, internalError);
+
+        // Return a shortened exception which may even be delivered to the frontend...
+        return new IllegalArgumentException(message);
     }
 
     private Object add(Object a, Object b) {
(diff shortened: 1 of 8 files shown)


qw3ry (approved):

src/main/java/sirius/pasta/noodle/compiler/Assembler.java
```diff
@@ -25,6 +26,14 @@
  */
 public class Assembler {
 
+    /**
+     * Represents the maximal internal index or offset which can be encoded in an bytecode.
```
sabieber, line 30:
```suggestion
     * Represents the maximum internal index or offset which can be encoded in an bytecode.
```

src/main/java/sirius/pasta/noodle/Invocation.java
```diff
@@ -273,13 +275,37 @@ private String computeOffendingPosition(Position position) {
...
         }
         result.append("^\n");
         return result.toString();
     }
 
+    /**
+     * Creates an exception which is used when Noodle detects an inconsistent or invalid internal state.
+     * <p>
+     * These states should normally not be reached as the compiler should only emit valid bytecodes. Therefore
+     * we provide quite some excessivle loggig here.
```
sabieber, line 289:
```suggestion
     * we provide quite some excessive logging here.
```

src/main/java/sirius/pasta/noodle/Invocation.java
```diff
@@ -273,13 +275,37 @@ private String computeOffendingPosition(Position position) {
...
+     *
+     * @param message the message to log
+     * @return an exception with a short and concise message of what happened. Also a long and elaborate message will
+     * be logged which might provide further insight.
+     */
+    private IllegalArgumentException vmError(String message) {
```
sabieber, line 295: maybe call it `throwVmError`, `createVmError` or `handleVmError`?

src/main/java/sirius/pasta/noodle/SharedConstantPool.java
```diff
@@ -78,4 +77,13 @@ public synchronized int getIndex(Object constant) {
...
         }
     }
+
+    /**
+     * Returns a copy of the list of shared constants.
+     *
+     * @return a copy of all shared constants. Note that this list cannot be mutated.
+     */
+    public synchronized List<Object> getSharedConstants() {
+        return Collections.unmodifiableList(new ArrayList<>(sharedConstants));
```
sabieber, line 87: Is this double wrapping necessary? Wouldnt `Collections.unmodifiableList(sharedConstants)` work, too?

sabieber (changes requested):

jakobvogel (approved):
(Subject to the remarks by @sabieber 😉)

src/main/java/sirius/pasta/noodle/SharedConstantPool.java
qw3ry: Dachte erst ja, weil wir oben was adden. Aber sind ja beide Methoden synchronized

idlira (approved):

src/main/java/sirius/pasta/noodle/compiler/Assembler.java
mkeckmkeck:
```suggestion
     * Represents the maximum internal index or offset which can be encoded in a bytecode.
```

src/main/java/sirius/pasta/noodle/SharedConstantPool.java
andyHa: The underlying list is a plain array list which can be modified concurrently. The unmodifyableList delegates all call to its given list, so we need to create a copy while being within the sync block...

src/main/java/sirius/pasta/noodle/compiler/Assembler.java
```diff
@@ -25,6 +26,14 @@
  */
 public class Assembler {
 
+    /**
+     * Represents the maximum internal index or offset which can be encoded in an bytecode.
```
mkeckmkeck, line 30:
```suggestion
     * Represents the maximum internal index or offset which can be encoded in a bytecode.
```

mkeckmkeck (approved):

sabieber (approved):

mkeckmkeck (approved):

jmuscireum (approved):

PR #795, opened 2021-01-21
Set SameSite policy and Secure-Flag for login cookie

qw3ry:
Set the policy to "lax" for now, which is the default (for modern browsers).
The config can be overridden to use another policy.

If the user accesses the site via SSL, the cookie is marked "secure".
In general this should not be a problem, but subsequent http-calls will be unauthenticated.

### DIFF
--- a/src/main/java/sirius/web/http/WebContext.java
+++ b/src/main/java/sirius/web/http/WebContext.java
@@ -19,6 +19,7 @@ import io.netty.handler.codec.http.HttpResponseStatus;
 import io.netty.handler.codec.http.QueryStringDecoder;
 import io.netty.handler.codec.http.QueryStringEncoder;
 import io.netty.handler.codec.http.cookie.Cookie;
+import io.netty.handler.codec.http.cookie.CookieHeaderNames;
 import io.netty.handler.codec.http.cookie.DefaultCookie;
 import io.netty.handler.codec.http.cookie.ServerCookieDecoder;
 import io.netty.handler.codec.http.multipart.Attribute;
@@ -294,6 +295,12 @@ public class WebContext implements SubContext {
      */
     @ConfigValue("http.sessionCookieTTL")
     private static Duration defaultSessionCookieTTL;
+
+    /**
+     * The same site attribute of the Session Cookie
+     */
+    @ConfigValue("http.sessionCookieSameSite")
+    private static CookieHeaderNames.SameSite sessionCookieSameSite;
 
     /**
      * Determines the domain set for all cookies. If empty no domain will be set.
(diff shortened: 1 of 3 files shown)


src/main/java/sirius/web/http/WebContext.java
```diff
@@ -1214,6 +1221,9 @@ public String getCookieValue(String name) {
      * @param cookie the cookie to send to the client
      */
     public void setCookie(Cookie cookie) {
+        if (isSSL()) {
+            cookie.setSecure(true);
```
andyHa, line 1225: do we always want to enforce this - this could bring a whole lot of side effects with it...

qw3ry: I'd suggest yes - nobody should use http when https is available.
There might be some edge cases, even if I cannot think of any right now. If the cookie is set `secure`, both the client and the server (obviously) at least support https. Incoming http links will break, but it seems fair to force https for them, if we don't do this at some point, they will continue with insecure connections forever.

We can introduce a config flag, whether we should enforce this; in case something breaks we can easily roll back

src/main/java/sirius/web/http/WebContext.java
```diff
@@ -1225,11 +1235,12 @@ public void setCookie(Cookie cookie) {
      * <p>
      * The generated cookie will be a session cookie and varnish once the user agent is closed
```
andyHa, line 1236:
```suggestion
     * The generated cookie will be a session cookie and vanish once the user agent is closed
```

src/main/java/sirius/web/http/WebContext.java
mkeckmkeck: +1 for config flag

mkeckmkeck (approved):

src/main/java/sirius/web/http/WebContext.java
```diff
@@ -104,6 +105,21 @@
     private static final Log SESSION_CHECK = Log.get("session-check");
     private static final long SESSION_PIN_COOKIE_TTL = TimeUnit.DAYS.toSeconds(10L * 365);
 
+    public enum CookieSecurity {
```
jmuscireum, line 108: JavaDoc

src/main/java/sirius/web/http/WebContext.java
```diff
@@ -104,6 +105,21 @@
...
+        /**
+         * the cookie shall only be send via https
```
jmuscireum, line 110:
```suggestion
         * the cookie shall only be sent via https
```

jmuscireum (changes requested):

idlira (approved):

jmuscireum (approved):

ymo-sci (approved):

bwiedmann (approved):

PR #808, opened 2021-02-05
Adds polyfill for forEach function of NodeList

sabieber:
IE 11 actually doesnt support this. We chose the concise variant as according to MDN this is actually the way some browsers implement NodeList.forEach.

Fixes: SIRI-266

### DIFF
--- a/src/main/resources/default/templates/wondergem/scripts/polyfill.html.pasta
+++ b/src/main/resources/default/templates/wondergem/scripts/polyfill.html.pasta
@@ -73,4 +73,12 @@
             return lastIndex !== -1 && lastIndex === position;
         };
     }
+
+    /**
+     * Polyfill for the NodeList.forEach function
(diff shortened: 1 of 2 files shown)


mkeckmkeck (approved):

src/main/resources/default/templates/wondergem/scripts/polyfill.html.pasta
```diff
@@ -73,4 +73,12 @@
             return lastIndex !== -1 && lastIndex === position;
         };
     }
+
```
andyHa, line 76: also port to tycho version of this script.

andyHa (changes requested):

andyHa (approved):

idlira (approved):

bwiedmann (approved):

PR #809, opened 2021-02-08
Improves error handling.

andyHa:
Throwing a HandledException out of nowhere comes pretty
unexpected. We therefore collect up to 250 errors/warnings
and ignore/reject any further messages (without crashing).


sabieber (approved):

idlira (approved):

mkeckmkeck (approved):

qw3ry:
I believe it would make sense to separate Errors and Warnings somehow. It should not be problematic to have 250 warnings (at least the compilation will work fine), but its problematic to have 50 errors, and we don't have to continue then

qw3ry (approved):

bwiedmann (approved):

PR #811, opened 2021-02-10
SIRI-317 Fix the JMP_FALSE handling

jmuscireum:
The JMP_FALSE opcode pops a value from the stack and will jump to a given label if the value is false. In the AND operation this will prevent the evaluation of the right operand and the whole operation will evaluate as false (See Conjunction.java). Likewise in a OR operation the JMP_FALSE will prevent the evaluation of the second operand and the whole operation will evaluate as true in case the first operand is true (See Disjunction.java).

In the case of 'null AND true' the null is not handled as being false, leading to a true evaluation of the operation. To fix this an additional check was added to handle null as false when the JMP_FALSE opcode is encountered.

### DIFF
--- a/src/main/java/sirius/pasta/noodle/Invocation.java
+++ b/src/main/java/sirius/pasta/noodle/Invocation.java
@@ -124,9 +124,7 @@ public class Invocation {
                     environment.writeVariable(index, pop());
                     break;
                 case JMP_FALSE:
-                    if (Boolean.FALSE.equals(pop())) {
-                        instructionPointer += index;
-                    }
+                    handleJumpFalse(index);
                     break;
                 case JMP:
                     instructionPointer += index;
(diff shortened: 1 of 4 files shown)


src/main/java/sirius/pasta/noodle/Invocation.java
```diff
@@ -266,6 +264,13 @@ private void handlePopField() {
         }
     }
 
+    private void handleJumpFalse(int index) {
+        Object value = pop();
+        if (value == null || Boolean.FALSE.equals(value)) {
```
jmuscireum, line 269: Oder `if (!Boolean.TRUE.equals(value))`?

andyHa: I'd keep the original if as suggested

andyHa (approved):

sabieber (approved):

mkeckmkeck (approved):

idlira (approved):

PR #813, opened 2021-02-15
Adds support for exporting BigIntegers in Excel as this is used commonly by Clickhouse.

andyHa:


src/main/java/sirius/web/util/IdGeneratorContext.java
```diff
@@ -37,6 +37,6 @@ public SubContext fork() {
 
     @Override
     public void detach() {
-
+        // Nothing to detach as we want to keep generating unqiue IDs across all forked threads
```
mkeckmkeck, line 40:
```suggestion
        // Nothing to detach as we want to keep generating unique IDs across all forked threads
```

mkeckmkeck (approved):
mini typo

bwiedmann (approved):

bwiedmann (approved):

src/main/java/sirius/web/http/TunnelHandler.java
```diff
@@ -397,6 +394,14 @@ private void logTiming(org.asynchttpclient.HttpResponseStatus status) {
         }
     }
 
+    private String safeConvertDuration(long durationMillis) {
+        if (durationMillis <= 0) {
```
sabieber, line 398: this is a logic change. before 0 would be actually displayed as 0 and not -. Is this intended?

sabieber (changes requested):

jmuscireum (approved):

src/main/java/sirius/web/http/TunnelHandler.java
andyHa: Yep, as otherwise we skipped the output entirely which is sort of a bug unexpected input for NLS.formatDuration

sabieber (approved):

PR #815, opened 2021-02-18
SIRI-321 Fix Optional<T>.orElse(null) resulting in return type void

jmuscireum:
The problem surfaces when we try to resolve the type of a method call. In the case of `Optional<String>.empty().orElse(null)` we already know the return type (T) will be String. But our logic throws this info away and replaces it with void when propagating the type of the null parameter value.

### DIFF
--- a/src/main/java/sirius/pasta/noodle/compiler/TypeTools.java
+++ b/src/main/java/sirius/pasta/noodle/compiler/TypeTools.java
@@ -44,7 +44,7 @@ public class TypeTools {
         TypeVariable<?>[] variables = ((Class<?>) parameterizedType.getRawType()).getTypeParameters();
         Type[] actualParameters = parameterizedType.getActualTypeArguments();
         for (int i = 0; i < variables.length; i++) {
-            typeTable.put(variables[i].getName(), actualParameters[i]);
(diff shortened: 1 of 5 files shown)


src/main/java/sirius/pasta/noodle/compiler/TypeTools.java
```diff
@@ -114,7 +114,13 @@ private void propagateConstantValueTypeInfos(Type parameterType, Node parameter)
 
         String typeVariableName = ((TypeVariable<?>) parameterType).getName();
         Type genericType = parameter.getGenericType();
-        typeTable.put(typeVariableName, genericType == null ? parameter.getType() : genericType);
+        Type type = genericType == null ? parameter.getType() : genericType;
+
+        if ("void".equals(type.getTypeName())) {
```
andyHa, line 119: wouldn't work `type != void.class` here? And in this case we can skip the putIfAbsent entirely as this doesn't narrow down anything, right?

sabieber (approved):

idlira (approved):

mkeckmkeck (approved):

cschierle (approved):

qw3ry (approved):

PR #818, opened 2021-02-25
Adds javadoc to reduce caveats by the misleading method name.

bwiedmann:
Until a clean solution has been made with SIRI-335.
Fixes: SIRI-333

### DIFF
--- a/src/main/java/sirius/web/security/UserContext.java
+++ b/src/main/java/sirius/web/security/UserContext.java
@@ -354,7 +354,7 @@ public class UserContext implements SubContext {
     /**
      * Returns "has-error" if an error was added for the given field.
      * <p>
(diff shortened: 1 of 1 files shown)


src/main/java/sirius/web/security/UserContext.java
```diff
@@ -353,6 +353,9 @@ public boolean hasError(String field) {
 
     /**
      * Returns "has-error" if an error was added for the given field.
+     * <p>
+     * Warning: Doesn't add an error to the given as the name might tell. This method will be removed with SIRI-335,
```
idlira, line 357:
```suggestion
     * Warning: Doesn't render the field as erroneous as the method name suggests. This method will be removed with SIRI-335,
```

idlira (approved):

cschierle (approved):

PR #819, opened 2021-03-01
Allows to add rows into sheets already created

idlira:
A new method allow us to select the Sheet to add rows, needed when we need to jump back and forth across sheets.

Note that this won't change the actual visible sheet once Excel opens. The first sheet remains as the visible one. A method could be added to change the current visible sheet if the need arrises.

Fixes: OX-6630

### DIFF
--- a/src/main/java/sirius/web/data/ExcelExport.java
+++ b/src/main/java/sirius/web/data/ExcelExport.java
@@ -68,7 +68,7 @@ public class ExcelExport {
     private final Workbook workbook;
     private Sheet currentSheet;
     private final Map<String, Integer> rows = new HashMap<>();
-    private int maxCols = 0;
+    private final Map<String, Integer> maxCols = new HashMap<>();
     private final CellStyle dateStyle;
     private final CellStyle numeric;
     private final CellStyle borderStyle;
@@ -209,7 +209,6 @@ public class ExcelExport {
         if (currentSheet != null) {
             autosizeColumns();
             addAutoFilter();
-            maxCols = 0;
         }
         if (Strings.isFilled(name)) {
             currentSheet = workbook.createSheet(WorkbookUtil.createSafeSheetName(name));
@@ -217,6 +216,7 @@ public class ExcelExport {
             currentSheet = workbook.createSheet();
         }
         rows.put(currentSheet.getSheetName(), 0);
+        maxCols.put(currentSheet.getSheetName(), 0);
 
         currentSheet.createFreezePane(0, 1, 0, 1);
         PrintSetup ps = currentSheet.getPrintSetup();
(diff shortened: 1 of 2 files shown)


sabieber (approved):

src/main/java/sirius/web/data/ExcelExport.java
```diff
@@ -225,6 +228,21 @@ public void createSheet(@Nullable String name) {
         currentSheet.setRepeatingRows(new CellRangeAddress(0, 0, -1, -1));
     }
 
+    /**
+     * Sets the sheet provided as the current sheet to receive data.
+     * <p>
+     * Note that this won't change the Sheet visible when the file is open.
```
bwiedmann, line 234:
```suggestion
     * Note that this won't set the Sheet active when the file is open.
```

bwiedmann (approved):

src/main/java/sirius/web/data/ExcelExport.java
```diff
@@ -65,8 +67,8 @@
 
     private final Workbook workbook;
     private Sheet currentSheet;
-    private int rows = 0;
-    private int maxCols = 0;
+    private final Map<String, Integer> rows = new HashMap<>();
```
andyHa, line 70: AtomicInteger? this has incrementAndGet bultin....

src/main/java/sirius/web/data/ExcelExport.java
```diff
@@ -509,8 +527,10 @@ public void writeToStream(OutputStream stream) {
     }
 
     private void addAutoFilter() {
```
andyHa, line 529: this should be applied to all sheets?!

src/main/java/sirius/web/data/ExcelExport.java
```diff
@@ -519,7 +539,8 @@ private void autosizeColumns() {
             // we do not want to autosize columns of streamed excel sheets, because of performance reasons
```
andyHa, line 539: this should be applied to all sheets?!

src/main/java/sirius/web/data/ExcelExport.java
```diff
@@ -583,4 +604,11 @@ public ExcelExport withMaxRowsReachedHandler(Consumer<String> maxRowsReachedHand
         this.maxRowsReachedHandler = maxRowsReachedHandler;
         return this;
     }
+
+    private int getAndIncrementRowForSheet(String name) {
```
andyHa, line 608: AtomicInteger#incrementAndGet

andyHa (changes requested):

bwiedmann (approved):

src/main/java/sirius/web/data/ExcelExport.java
```diff
@@ -65,8 +68,8 @@
 
     private final Workbook workbook;
     private Sheet currentSheet;
-    private int rows = 0;
-    private int maxCols = 0;
+    private final Map<String, AtomicInteger> rows = new HashMap<>();
```
andyHa, line 71: shoudln't we use the sheet index here, as the name can be non-unique(?)

idlira: Names are unique, but pushing a version based on the index.

bwiedmann (approved):

sabieber (approved):

mkeckmkeck (approved):

PR #820, opened 2021-03-04
Updates sirius kernel and marks JSONCall as POST if sending output

meg1502:


sabieber (approved):

src/main/java/sirius/web/services/JSONCall.java
```diff
@@ -68,11 +68,14 @@ public void addHeader(String name, String value) {
 
     /**
      * Can be used to generate the JSON request.
+     * <p>
+     * This will mark the underlying {@link Outcall} as a POST request.
      *
-     * @return the an input which can be used to generate an XML document which is sent to the URL
+     * @return the an input which can be used to generate an JSON document which is sent to the URL
```
cschierle, line 74:
```suggestion
     * @return the an input which can be used to generate a JSON document which is sent to the URL
```
😉

cschierle (approved):

andyHa (approved):

PR #830, opened 2021-04-08
Simplifies the coercing of numbers when evaluating binary operations

jmuscireum:
The current construct is hard to understand. Also the bug fixed with #827 showed that this method could be called with type = Long and otherType = Object, which would fulfill the first if and then the Long would be falsely cast to Double.

### DIFF
--- a/src/main/java/sirius/pasta/noodle/compiler/Assembler.java
+++ b/src/main/java/sirius/pasta/noodle/compiler/Assembler.java
@@ -132,15 +132,18 @@ public class Assembler {
 
         if (Long.class.equals(boxedOtherType) && Integer.class.equals(boxedType)) {
             emitByteCode(OpCode.COERCE_INT_TO_LONG, 0, position);
+            return;
         }
 
         if (Double.class.equals(boxedOtherType)) {
             if (Long.class.equals(boxedType)) {
                 emitByteCode(OpCode.COERCE_LONG_TO_DOUBLE, 0, position);
+                return;
             }
 
             if (Integer.class.equals(boxedType)) {
                 emitByteCode(OpCode.COERCE_INT_TO_DOUBLE, 0, position);
+                return;
             }
         }
     }
--- a/src/main/java/sirius/pasta/noodle/compiler/Assembler.java
+++ b/src/main/java/sirius/pasta/noodle/compiler/Assembler.java
@@ -123,16 +123,24 @@ public class Assembler {
      * @param otherType the target type expected by the subsequent bytecodes
      */
     public void coerce(Position position, Class<?> type, Class<?> otherType) {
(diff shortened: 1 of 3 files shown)


src/main/java/sirius/pasta/noodle/compiler/Assembler.java
```diff
@@ -123,16 +123,24 @@ public void emitPushConstant(Object constant, Position position) {
...
-            if (CompilationContext.isAssignableTo(Double.class, CompilationContext.autoboxClass(otherType))) {
+        Class<?> boxedType = CompilationContext.autoboxClass(type);
+        Class<?> boxedOtherType = CompilationContext.autoboxClass(otherType);
+
+        if (boxedType.equals(boxedOtherType)) {
+            return;
+        }
+
+        if (Long.class.equals(boxedOtherType) && Integer.class.equals(boxedType)) {
+            emitByteCode(OpCode.COERCE_INT_TO_LONG, 0, position);
```
andyHa, line 134:
```suggestion
            emitByteCode(OpCode.COERCE_INT_TO_LONG, 0, position);
            return;
```

src/main/java/sirius/pasta/noodle/compiler/Assembler.java
```diff
@@ -123,16 +123,24 @@ public void emitPushConstant(Object constant, Position position) {
...
+        }
+
+        if (Double.class.equals(boxedOtherType)) {
+            if (Long.class.equals(boxedType)) {
+                emitByteCode(OpCode.COERCE_LONG_TO_DOUBLE, 0, position);
```
andyHa, line 139:
```suggestion
                emitByteCode(OpCode.COERCE_LONG_TO_DOUBLE, 0, position);
                return;
```

src/main/java/sirius/pasta/noodle/compiler/Assembler.java
```diff
@@ -123,16 +123,24 @@ public void emitPushConstant(Object constant, Position position) {
...
+            }
+
+            if (Integer.class.equals(boxedType)) {
                 emitByteCode(OpCode.COERCE_INT_TO_DOUBLE, 0, position);
```
andyHa, line 143:
```suggestion
                emitByteCode(OpCode.COERCE_INT_TO_DOUBLE, 0, position);
                return;
```

andyHa (commented):
Would return early to avoid accidential double coercions...

mkeckmkeck (approved):

andyHa (approved):

sabieber (approved):

qw3ry (approved):

fhaScireum (approved):

PR #831, opened 2021-04-16
Avoids a ConcurrentModificationException in Java9+.

andyHa:
The code itself was and if safe in the way we call it. However, Java9+
throws an exception for concurrent modifications within computeIfAbsent
as it fears infinite recursions for the same key (which isn't the case
in our scenario).

### DIFF
--- a/src/main/java/sirius/pasta/tagliatelle/Tagliatelle.java
+++ b/src/main/java/sirius/pasta/tagliatelle/Tagliatelle.java
@@ -102,7 +102,7 @@ public class Tagliatelle {
                 extensions.put(target, new ArrayList<>());
 
                 // Load and fill the result list. Note that we cannot use HashMap.computeIfAbsent here, as
-                // loadExtensions it self compiles templates which might end up calling this method leading
+                // loadExtensions itself compiles templates which might end up calling this method leading
                 // to a ConcurrentModificationException...
                 result = loadExtensions(target);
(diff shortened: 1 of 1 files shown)


fhaScireum (approved):

mkeckmkeck (approved):

src/main/java/sirius/pasta/tagliatelle/Tagliatelle.java
```diff
@@ -93,7 +94,20 @@
...
-            return Collections.unmodifiableList(extensions.computeIfAbsent(target, this::loadExtensions));
+            List<TemplateExtension> result = extensions.get(target);
+            if (result == null) {
+                // As we might re-enter this method while compiling templates in "loadExtensions" we provide a
+                // placeholder value here to avoid infinite recursions. (Under normal circumstances, this shouldn't
+                // happen anyway, as cyclic extensions make no sense at all).
+                extensions.put(target, new ArrayList<>());
+
+                // Load and fill the result list. Note that we cannot use HashMap.computeIfAbsent here, as
+                // loadExtensions it self compiles templates which might end up calling this method leading
```
idlira, line 105:
```suggestion
                // loadExtensions itself compiles templates which might end up calling this method leading
```

idlira (approved):

sabieber (approved):

PR #842, opened 2021-05-28
An UserInfo can now have "every permission"

qw3ry:


andyHa (approved):

sabieber (approved):

src/main/java/sirius/web/security/UserInfo.java
```diff
@@ -167,7 +169,23 @@ public Builder withLang(String lang) {
...
+            return this;
+        }
+
+        /**
+         * Indicates whether the user has all available permissions by default
+         *
+         * @param hasEveryPermission true when the user should have all permissions
+         * @return the builder itself for fluent method calls
+         */
+        public Builder withEveryPermission(boolean hasEveryPermission) {
```
andyHa, line 186: one could also make this "grantEveryPermission()" - as this is probably never invoked with false?

idlira (approved):

meg1502 (approved):

cschierle (approved):

src/main/java/sirius/web/security/UserInfo.java
qw3ry: I thought about this - I saw no benefit in terms of readability (`.withEveryPermission(true)` vs `.grantEveryPermission()`), it breaks the naming schema and it cannot be overridden later, so I kept this version. What benefits do you see from the signature change?

ymo-sci (approved):

mkeckmkeck (approved):

PR #856, opened 2021-06-18
Misc Fixes and Improvements for Tycho. Most notably a new macro to generate cachable asset URIs

andyHa:


src/main/resources/default/taglib/t/helpbox.html.pasta
```diff
@@ -1,8 +1,9 @@
+<i:arg type="String" name="class" default="" description="Permits to supply additional CSS classes" />
```
sabieber, line 1: Optional: Wouldnt it generally be better to first name the name of the parameter and then the type as the param name is the first thing you need to know when calling a template and in my opinion is also more important than the param type (especially for optional params)?

src/main/java/sirius/web/controller/Message.java
```diff
@@ -76,14 +76,14 @@ public Message withHTMLMessage(String htmlMessage) {
         public Message withTextAndLink(String textMessage, String label, String link, @Nullable String icon) {
             if (Strings.isFilled(icon)) {
                 return new Message(type,
-                                   Strings.apply("<span>%s</span><a href=\"%s\"><i class=\"fa %s\"></i> %s</a>",
+                                   Strings.apply("<span class\"pr-2\">%s</span><a href=\"%s\"><i class=\"fa %s\"></i> %s</a>",
```
sabieber: Optional: Escaping of the quotes could potentially be omitted when using `'` instead.

src/main/resources/default/assets/tycho/scripts/tooltips.js
```diff
@@ -0,0 +1,5 @@
+// Initializes all Bootstrap tooltips...
+sirius.ready(function() {
+    // This needs to use jQuery for Bootstrap compatibility reasons...
```
sabieber, line 3: Optional: Thats why it would be nice to upgrade to BS 5.0.1 right away as it doesnt seem to require jQuery for things like this: https://getbootstrap.com/docs/5.0/components/tooltips/

src/main/resources/default/assets/tycho/scripts/focus.js
```diff
@@ -0,0 +1,44 @@
...
+ * by scireum in Remshalden, Germany
+ *
+ * Copyright by scireum GmbH
+ * http://www.scireum.de - info@scireum.de
+ */
+
+sirius.ready(function() {
+
+    function focus(field) {
+        field.focus();
```
sabieber, line 12: Optional: use the _ prefix for all DOM elements in this file, such as _field, _primaryField, _form, ...

src/main/java/sirius/web/controller/Message.java
```diff
@@ -64,6 +64,18 @@ public Message withHTMLMessage(String htmlMessage) {
...
 
+        /**
+         * Specifies a text message followed by a link.
+         *
+         * @param textMessage the plain text message to show
+         * @param label       the label of the link to show
+         * @param link        the target of the link. Use <tt>javascript:</tt> as prefix to invoke a JavaScript function
+         * @return the generated message
+         */
+        public Message withTextAndLink(String textMessage, String label, String link) {
```
sabieber, line 75: Feedback: Nice, thx 👌

sabieber (approved):

idlira (approved):

src/main/resources/default/taglib/t/helpbox.html.pasta
andyHa: true - coming from java it was kind of the default to declare a variable "String class" and thus the pattern was chose. We can switch this around, but to make it consistent I'd do it for every template in a separate PR

src/main/resources/default/assets/tycho/scripts/tooltips.js
andyHa: Yep, but I think moving to BS5 requires its own PR and approach without any modifications next to it.

src/main/resources/default/taglib/t/helpbox.html.pasta
sabieber: Sounds great

sabieber (approved):

ymo-sci (approved):

PR #873, opened 2021-06-29
Uses the newest version of token-autocomplete lib

meg1502:
Fixes: OX-6973


sabieber (changes requested):
Why is the .ts not updated?

sabieber (approved):

cschierle (approved):

PR #875, opened 2021-06-30
Adds a JS function to listen for elements becoming visible

jmuscireum:


src/main/resources/default/assets/common/core.js
```diff
@@ -71,3 +71,92 @@ sirius.postJSON = function(url, params) {
...
+            last = now;
+            fn.apply(context, args);
+        }
+    };
+}
+
+/**
+ * Adds an event listener that triggers exactly once per element when it becomes visible or is about to become visible.
+ * <p>
+ * The event will trigger once a element comes within a certain distance to the viewport. By default the vertical and
```
sabieber, line 110:
```suggestion
 * The event will trigger once an element comes within a certain distance to the viewport. By default the vertical and
```

src/main/resources/default/assets/common/core.js
```diff
@@ -71,3 +71,92 @@ sirius.postJSON = function(url, params) {
...
+            if (!element.getAttribute('data-scroll-event-handled') && isElementNowOrSoonVisible(element)) {
+                element.setAttribute('data-scroll-event-handled', true);
+
+                listener(element);
+            }
+        }
+    };
+
+    // Check after the page is loaded and when resizing or scrolling the page
+    sirius.ready(handleVisibleElements)
```
sabieber, line 149:
```suggestion
    sirius.ready(handleVisibleElements);
```

sabieber (approved):

idlira (approved):

src/main/resources/default/assets/common/core.js
```diff
@@ -71,3 +71,92 @@ sirius.postJSON = function(url, params) {
...
+
+/**
+ * Throttles a javascript function by calling it only every x milliseconds.
+ * <p>
+ * The throttled function is called once immediately, and then every x milliseconds.
+ *
+ * @param fn        the function to be throttled
+ * @param threshold the time delaying each function call (in milliseconds).
+ */
+sirius.throttle = function (fn, threshold) {
```
meg1502, line 83: can we add a boolean to control if a leading call happens or not?

src/main/resources/default/assets/common/core.js
```diff
@@ -71,3 +71,92 @@ sirius.postJSON = function(url, params) {
...
+        const args = arguments;
+
+        if (last && now < last + threshold) {
+            // hold on to it
+            clearTimeout(deferTimer);
+
+            deferTimer = setTimeout(function () {
+                last = now;
+                fn.apply(context, args);
+            }, threshold);
```
meg1502, line 99: shouldnt this timeout be the remaining time instead of the treshhold? else it behaves more like debounce -

e.g. if i call it a 0, with a treshhold of 1000, and again at 800, the timeout is set to execute at 1800, and the correct behavior would be 1000, i think?

meg1502 (changes requested):

src/main/resources/default/assets/common/core.js
jmuscireum: You're right, i corrected that

src/main/resources/default/assets/common/core.js
jmuscireum: In what use case do you not want to have an immediate reaction to an event?

mkeckmkeck (approved):

src/main/resources/default/assets/common/core.js
meg1502: since these methods use cases are so universal, there should be a lot of use cases for any configuration, e.g. if you want to be "snappy" for the first interaction but still want to throttle in the long run, but we can add these options later when we neeed them

meg1502 (approved):

PR #881, opened 2021-07-05
Improves the sirius.isEmpty and sirius.isFilled methods

jmuscireum:


sabieber (approved):

src/main/resources/default/assets/common/core.js
```diff
@@ -1,11 +1,11 @@
...
+sirius.isEmpty = function (value) {
+    return value === null || value === undefined || typeof value === 'undefined' || value === '';
 }
 
-// Checks if the given string is non-empty.
-sirius.isFilled = function (aString) {
-    return aString !== null && aString !== undefined && aString !== '';
+// Checks if the given value is non-empty.
+sirius.isFilled = function (value) {
+    return value !== null && value !== undefined && typeof value !== 'undefined' && value !== '';
```
idlira, line 8: Why not `return !sirius.isEmpty()`?

src/main/resources/default/assets/common/core.js
```diff
@@ -1,11 +1,11 @@
-// Checks if the given string is empty or null.
-sirius.isEmpty = function (aString) {
-    return aString === null || aString === undefined || aString === '';
+// Checks if the given value is empty or null.
+sirius.isEmpty = function (value) {
+    return value === null || value === undefined || typeof value === 'undefined' || value === '';
```
idlira, line 3: Should that trim value before the check?

jmuscireum: I don't know. Opinions?

qw3ry: I prefer to be consistent with [`Strings.isEmpty`](https://github.com/scireum/sirius-kernel/blob/48bc76c7459b9bbeaeb9cc15eedf986f524f5fd0/src/main/java/sirius/kernel/commons/Strings.java#L115) which doesn't do a trim

qw3ry (approved):

idlira (approved):

meg1502:
We could also extend this to arrays with no items and objects with no own properties

src/main/resources/default/assets/common/core.js
meg1502, line 3:
```suggestion
    return value === null || typeof value === 'undefined' || value === undefined || value === '';
```
i think they need to be flipped to prevent some errors

jmuscireum:
> We could also extend this to arrays with no items and objects with no own properties

We can do that when it's needed. It seems to not be trivial especially with our Object polyfills.

mkeckmkeck (approved):

meg1502 (approved):

PR #885, opened 2021-07-12
Only binds logged in user to the context

cschierle:
Returns the current user, only if they are logged in. 
Otherwise, calls without active login were resolved with an incorrect and inconsistent language (e.g. login page displayed in German -> submit invalid user data -> login page switched to English)

Fixes: SIRI-376

### DIFF
--- a/src/main/java/sirius/web/security/UserContext.java
+++ b/src/main/java/sirius/web/security/UserContext.java
@@ -219,7 +219,6 @@ public class UserContext implements SubContext {
 
         if (currentUser != null) {
             if (currentUser.isLoggedIn()) {
-                setCurrentUser(currentUser);
                 return Optional.of(currentUser);
             } else {
                 return Optional.empty();
--- a/src/main/java/sirius/web/security/UserContext.java
+++ b/src/main/java/sirius/web/security/UserContext.java
@@ -218,7 +218,12 @@ public class UserContext implements SubContext {
         }
 
         if (currentUser != null) {
-            return Optional.of(currentUser);
+            if (currentUser.isLoggedIn()) {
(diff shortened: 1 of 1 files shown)


ymo-sci (approved):

src/main/java/sirius/web/security/UserContext.java
```diff
@@ -218,7 +218,12 @@ private void bindUserToRequest(WebContext ctx) {
         }
 
         if (currentUser != null) {
-            return Optional.of(currentUser);
+            if (currentUser.isLoggedIn()) {
+                setCurrentUser(currentUser);
```
andyHa, line 222: I'd say this is redundant and can be skipped... (we're already reading currentUser)...

andyHa (changes requested):

idlira (approved):

sabieber (approved):

mkeckmkeck (approved):

src/main/java/sirius/web/security/UserContext.java
```diff
@@ -218,7 +218,12 @@ private void bindUserToRequest(WebContext ctx) {
...
+                return Optional.of(currentUser);
+            } else {
+                return Optional.empty();
+            }
```
qw3ry: this _could_ be simplified to
```suggestion
            return Optional.of(currentUser).filter(UserInfo::isLoggedIn);
```

jmuscireum (approved):

qw3ry (approved):

meg1502 (approved):

fhaScireum (approved):

PR #890, opened 2021-07-16
Removes the Macro.isConstant default impl and fixes a bunch of sonarlint and IntelliJ inspections.

andyHa:
Although nearly all macros are constant, we want the developer to make
a conscious decision here, because for non-constant macros it is quite
a dangerous default and easily missed - which leads to all kinds of
dirty side-effects.


sabieber (approved):

src/main/java/sirius/pasta/noodle/macros/ToListMacro.java
```diff
@@ -57,7 +57,7 @@ public String getName() {
 
     @Override
     public String getDescription() {
-        return "Returns a list containing of the given items or the given array";
+        return "Returns a list containing of the given items ,or the given array";
```
qw3ry, line 60: .

jmuscireum (approved):

src/main/java/sirius/pasta/noodle/compiler/TypeTools.java
```diff
@@ -226,26 +223,21 @@ private void reduceTable() {
...
             return typeTable.getOrDefault(getTypeVariableKey(getScopeOfType(typeVariable), typeVariable.getName()),
                                           type);
         }
-        if (type instanceof WildcardType) {
-            if (((WildcardType) type).getLowerBounds().length > 0) {
-                return simplify(((WildcardType) type).getLowerBounds()[0]);
-            }
+        if ((type instanceof WildcardType) && ((WildcardType) type).getLowerBounds().length > 0) {
+            return simplify(((WildcardType) type).getLowerBounds()[0]);
         }
```
qw3ry, line 232: Not sure if
```
if ((type instanceof WildcardType wildcardType) && wildcardType.getLowerBounds().length > 0 )
```
works, but if it does it would be nice

src/main/java/sirius/web/controller/Message.java
```diff
@@ -229,7 +229,7 @@ public static Builder error() {
      * @return a new message with the given content and ERROR as type
      * @deprecated Use {@code success().withText(textMessage)}
```
qw3ry, line 230: looks like a copy paste err

src/main/java/sirius/web/data/ExcelExport.java
```diff
@@ -171,16 +171,12 @@ protected float getScaleFactor() {
...
-                    return Workbook.PICTURE_TYPE_JPEG;
-                case "image/bmp":
-                    return Workbook.PICTURE_TYPE_DIB;
-                default:
-                    return -1;
-            }
+            return switch (mimeType) {
+                case MimeHelper.IMAGE_PNG -> Workbook.PICTURE_TYPE_PNG;
+                case MimeHelper.IMAGE_JPEG -> Workbook.PICTURE_TYPE_JPEG;
+                case "image/bmp" -> Workbook.PICTURE_TYPE_DIB;
```
qw3ry, line 177: Is it worth to add a constant to MimeHelper?

src/main/java/sirius/pasta/noodle/compiler/TypeTools.java
andyHa: works!

PR #908, opened 2021-08-16
Passes the keydown event to the event handler

jmuscireum:


andyHa (approved):

src/main/resources/default/assets/common/core.js
```diff
@@ -43,7 +43,7 @@ sirius.addEnterListener = function(_node, listener) {
     _node.addEventListener('keydown', function(event) {
         if (event.code === 'Enter') {
```
meg1502, line 44: event.code is not supported in IE 11, we have to use keyCode as long as we support IE

jmuscireum: event.key can be used in the case of Enter

meg1502 (approved):

PR #920, opened 2021-08-30
Adds a parameter to customize the btn class of t:dropdown

meg1502:


mkeckmkeck (approved):

src/main/resources/default/taglib/t/dropdown.html.pasta
```diff
@@ -5,13 +5,15 @@
...
+<i:arg type="String" name="btnClass" default="btn btn-link"
+       description="Defines the classes that define the label button."/>
 
 <i:pragma name="description" value="Renders an inline dropdown within a Tycho template"/>
 <i:local name="contents" value="@renderToString('body')"/>
 
 <i:if test="isFilled(contents)">
     <div class="dropdown @class">
-        <a class="btn btn-link"
+        <a class="@btnClass"
```
idlira, line 16: Shouldn't it be `<a class="btn btn-link @btnClass"` instead?
I think that's how we do in some other templates, where one can **append** classes...

andyHa (approved):

sabieber (approved):

src/main/resources/default/taglib/t/dropdown.html.pasta
sabieber: Actually the default for the template param is the old "btn btn-link", so maybe the indended usecase here is to be able to overwrite the class

meg1502: yeah there are quit a lot of buttons like btn-primary btn-secondary btn-danger etc.. and while most dropdowns will probably use btn-link, this parameter makes it able to customize the type of btn - in my case btn-primary for the createButton dropdown

PR #928, opened 2021-09-14
Makes tycho history work with links to the same site

meg1502:
Previously, links to the same side where behaving incorrectly as in triggering an unwanted back operation. This was the case because unlike posts to the same site, links to the same site will in most browser not append a new history entry, but instead replace the current state (with null). Therefore we store the length of the browser history together with our own array, so we know with which history entry the last url corresponds.

Fixes: SIRI-421


sabieber (approved):
Finde ja man könnte den Code noch n bissl kommentieren, so richtig intuitiv ist die API ja net

cschierle (approved):

fhaScireum (approved):

idlira (approved):

mkeckmkeck (approved):

PR #972, opened 2021-11-24
Improves styles of imageUpload tag and gives it a fixed height

meg1502:
Fixes: OX-7837
Screenshot 2021-11-24 at 12 40 59

### DIFF
--- a/src/main/resources/default/assets/tycho/styles/forms.scss
+++ b/src/main/resources/default/assets/tycho/styles/forms.scss
@@ -136,17 +136,6 @@
   min-height: 0;
 }
 
-.dropzone.sirius-imageupload {
-  height: 350px;
-
-  .img-preview {
-    display: flex;
-    justify-content: center;
-    align-items: center;
-    flex-grow: 1;
-  }
-}
-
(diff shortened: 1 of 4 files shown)


sabieber (approved):

idlira (approved):

cschierle (approved):

fhaScireum (approved):

src/main/resources/default/assets/tycho/styles/forms.scss
```diff
@@ -137,23 +137,20 @@
 }
 
 .dropzone.sirius-imageupload {
+  height: 350px;
+
   .img-preview {
```
andyHa, line 142: This isn't idiomatic to bootstrap4. We should remove this class and add the appropriate marker css classes to the img / div itself like d-flex justify-content-center flex-grow-1 ....

src/main/resources/default/assets/tycho/styles/forms.scss
```diff
@@ -137,23 +137,20 @@
 }
 
 .dropzone.sirius-imageupload {
+  height: 350px;
```
andyHa, line 140: this could be an inline style as it is only in one place(?)

PR #973, opened 2021-11-25
Properly supports barcodes for GTIN-14 using the ITF format

cschierle:
This is commonly used to create barcodes for GTINs with 14 digits used for certain types of packaging (see https://en.wikipedia.org/wiki/ITF-14).

`BarcodePdfReplaceHandler` and `BarcodeController` now both use `zxing` to generate barcodes (except for `interleaved2of5checksummed` in PDF templates, which remains unchanged).

**NOTE:** `BarcodePdfReplaceHandler` did generate barcodes for such GTINs, but they were unreliable when scanning.

Example:
The previous PDF replace produced a barcode for ``, that cannot be reliably resolved back to "12345678987654":
12345678987654_old

With these changes, the produced barcode looks like this and can be scanned reliably:
12345678987654_new

Fixes: SIRI-487

### DIFF
--- a/src/main/java/sirius/web/templates/pdf/handlers/BarcodePdfReplaceHandler.java
+++ b/src/main/java/sirius/web/templates/pdf/handlers/BarcodePdfReplaceHandler.java
@@ -44,7 +44,7 @@ public class BarcodePdfReplaceHandler extends PdfReplaceHandler {
             throw new IllegalArgumentException("The URI is required to match the format 'barcode://type/content'");
         }
 
-        Image awtImage = BarcodeController.generateBarcodeImage(barcodeInfo[0], barcodeInfo[1]);
+        Image awtImage = BarcodeController.generateBarcodeImage(barcodeInfo[0], barcodeInfo[1], cssWidth, cssHeight);
 
         FSImage fsImage = new ITextFSImage(com.lowagie.text.Image.getInstance(awtImage, Color.WHITE, true));
 
--- a/src/main/java/sirius/web/templates/pdf/handlers/BarcodePdfReplaceHandler.java
+++ b/src/main/java/sirius/web/templates/pdf/handlers/BarcodePdfReplaceHandler.java
@@ -8,17 +8,16 @@
 
 package sirius.web.templates.pdf.handlers;
 
+import com.google.zxing.WriterException;
 import com.lowagie.text.pdf.Barcode;
-import com.lowagie.text.pdf.Barcode128;
-import com.lowagie.text.pdf.BarcodeEAN;
 import com.lowagie.text.pdf.BarcodeInter25;
 import org.xhtmlrenderer.extend.FSImage;
 import org.xhtmlrenderer.extend.UserAgentCallback;
 import org.xhtmlrenderer.pdf.ITextFSImage;
 import sirius.kernel.commons.Strings;
 import sirius.kernel.di.std.Register;
+import sirius.web.util.BarcodeController;
 
-import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import java.awt.Color;
 import java.awt.Image;
@@ -53,10 +52,7 @@ public class BarcodePdfReplaceHandler extends PdfReplaceHandler {
             throw new IllegalArgumentException("The URI is required to match the format 'barcode://type/content'");
         }
 
-        Barcode code = createBarcode(barcodeInfo[0]);
-        code.setCode(padCodeIfNecessary(code, barcodeInfo[1]));
-
-        Image awtImage = code.createAwtImage(Color.BLACK, Color.WHITE);
+        Image awtImage = getBarcodeImage(barcodeInfo[0], barcodeInfo[1], cssWidth, cssHeight);
 
         int scaleFactor = calculateBarcodeScaleFactor(cssWidth, cssHeight, awtImage);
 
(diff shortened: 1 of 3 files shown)


src/main/java/sirius/web/util/BarcodeController.java
```diff
@@ -35,10 +36,10 @@
...
     /**
-     * Creates an QR code for the given content.
+     * Creates a QR code for the given content.
      * <p>
-     * The parameter <tt>content</tt> determines the contents of the qr code. The parameters <tt>with</tt> and
-     * <tt>height</tt> its dimensions.
+     * The parameter <tt>content</tt> determines the contents of the qr code. The parameters <tt>width</tt> and
+     * <tt>height</tt> determine its dimensions.
      * </p>
      *
```
sabieber, line 44:
```suggestion
     * <tt>height</tt> determine its dimensions.
     *
```

src/main/java/sirius/web/util/BarcodeController.java
```diff
@@ -49,6 +50,16 @@ public void qr(WebContext ctx) throws Exception {
         barcode(ctx, BarcodeFormat.QR_CODE);
     }
 
+    /**
+     * Creates a barcode for the given content.
+     * <p>
+     * The parameter <tt>content</tt> determines the contents of the barcode. The parameters <tt>width</tt> and
+     * <tt>height</tt> determine its dimensions.
+     * </p>
+     *
```
sabieber, line 59:
```suggestion
     * <tt>height</tt> determine its dimensions.
     *
```

sabieber: Its crazy but JavaDoc is no real HTML syntax. So they use HTML tags to do some stuff, but for example Paragraph Tags should only open and not be closed ...

sabieber (approved):

src/main/java/sirius/web/util/BarcodeController.java
cschierle: Thank you, that closing tag slipped through my cleanup :)

sabieber (approved):

src/main/java/sirius/web/templates/pdf/handlers/BarcodePdfReplaceHandler.java
```diff
@@ -73,65 +69,39 @@ public FSImage resolveUri(String uri, UserAgentCallback userAgentCallback, int c
         return fsImage;
     }
 
-    private int calculateBarcodeScaleFactor(int cssWidth, int cssHeight, Image awtImage) {
-        return (int) Math.max(Math.ceil(cssWidth / (float) awtImage.getWidth(null)),
-                              Math.ceil(cssHeight / (float) awtImage.getHeight(null)));
-    }
+    private Image getBarcodeImage(String barcodeType, String content, int width, int height) throws WriterException {
```
sabieber, line 72: Shouldnt we name those methods something other than `get...` as they are not getters but do more work? Something like `generateBarcodeImage` for example

fhaScireum (approved):

src/main/java/sirius/web/templates/pdf/handlers/BarcodePdfReplaceHandler.java
```diff
@@ -73,65 +69,40 @@ public FSImage resolveUri(String uri, UserAgentCallback userAgentCallback, int c
...
-        }
-
-        if (BARCODE_TYPE_EAN.equalsIgnoreCase(type)) {
-            return new BarcodeEAN();
+        if (BARCODE_TYPE_INTERLEAVED_2_OF_5_CHECKSUMMED.equalsIgnoreCase(barcodeType)) {
+            Barcode code = new BarcodeInter25();
+            code.setGenerateChecksum(true);
+            // Pads the code if necessary: Length is even but a checksum will be added
+            code.setCode(BarcodeInter25.keepNumbers(content).length() % 2 == 0 ? "0" + content : content);
+            return code.createAwtImage(Color.BLACK, Color.WHITE);
```
jmuscireum, line 81: Can this somehow be replaced by zxing?

cschierle: Didn't find a suitable way to do this with zxing.

There were some discussions about this (e.g. https://github.com/zxing/zxing/issues/858), but also no obvious suggested solution.

cschierle: Solved by applying the calculation of the previously used library, which is still used for PDF Images anyway. Barcode rendering itself is now fully performed using zxing.

fhaScireum (approved):

idlira (approved):

sabieber (approved):

andyHa (approved):

src/main/java/sirius/web/util/BarcodeController.java
```diff
@@ -59,39 +83,94 @@ private void barcode(WebContext webContext, BarcodeFormat format) throws WriterE
...
+    /**
+     * Generates an image of a barcode
+     *
+     * @param type    the desired barcode type
+     * @param content the content of the barcode
+     * @return a barcode of the given data as a 200x200 px image
+     * @throws WriterException if generating the image fails
+     */
+    public static Image generateBarcodeImage(String type, String content) throws WriterException {
+        if (!NUMERIC.matcher(content).matches()) {
```
jmuscireum, line 115: Does this apply to every type?

src/main/java/sirius/web/util/BarcodeController.java
```diff
@@ -59,39 +83,94 @@ private void barcode(WebContext webContext, BarcodeFormat format) throws WriterE
...
+        }
+
+        BarcodeFormat format = determineFormat(type);
+
+        content = alignContentForItfFormat(content, type);
+
+        format = useItfFormatForGtin14(content, format);
+
+        Writer writer = determineWriter(format);
+        BitMatrix matrix = writer.encode(content, format, 200, 200);
```
jmuscireum, line 127: The cssWidth and cssHeight needs to be used here. At least that was the case in my tests

src/main/java/sirius/web/util/BarcodeController.java
cschierle: To all 1-dimensional codes, yes. Forgot to exclude QR from this check though

jmuscireum (approved):

PR #975, opened 2021-11-30
Improvs reporting for leftover raw class literals.

andyHa:
The most common case is a missing ".class" like in a @part(MyClass)
(instead of @part(MyClass.class).

We now report a proper error for this case.


qw3ry (approved):
👍

cschierle (approved):

sabieber (approved):

PR #994, opened 2022-01-31
Properly implements the "default-template" mechanism for extensions.

andyHa:
If a template in "default/extensions/name" and one in "extensions/name"
exists, we now only load the second (in accordance with all other places
where default templates are overwritten by ordinary ones with the same
name).

This way a default extension in a library can be provided which can be
overwritten in a product).

### DIFF
--- a/src/main/java/sirius/pasta/tagliatelle/Tagliatelle.java
+++ b/src/main/java/sirius/pasta/tagliatelle/Tagliatelle.java
@@ -8,6 +8,7 @@
 
 package sirius.pasta.tagliatelle;
 
+import com.google.common.collect.Streams;
 import parsii.tokenizer.ParseError;
 import parsii.tokenizer.Position;
 import sirius.kernel.Sirius;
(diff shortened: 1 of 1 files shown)


jakobvogel (approved):

src/main/java/sirius/pasta/tagliatelle/Tagliatelle.java
```diff
@@ -124,15 +125,29 @@
...
-                                           + "/.*.pasta"))
-                     .map(m -> m.group(0))
-                     .map(this::resolveToTemplateExtension)
-                     .filter(Objects::nonNull)
-                     .sorted()
-                     .collect(Collectors.toList());
+        List<TemplateExtension> extensionsForTarget = Sirius.getClasspath()
+                                                            .find(Pattern.compile("(customizations/[^/]+/)?extensions/"
+                                                                                  + Pattern.quote(target)
+                                                                                  + "/.*.pasta"))
```
sabieber, line 131: Extract pattern into constant so it is not always recompiled?

src/main/java/sirius/pasta/tagliatelle/Tagliatelle.java
```diff
@@ -124,15 +125,29 @@
...
+                                                            .map(m -> m.group(0))
+                                                            .map(this::resolveToTemplateExtension)
+                                                            .filter(Objects::nonNull)
+                                                            .toList();
+
+        // Also check for extensions provided as default - which are only loaded, if no overwrites (same path without
+        // /default/ in front) exists...
+        return Streams.concat(extensionsForTarget.stream(),
+                              Sirius.getClasspath()
+                                    .find(Pattern.compile("default/extensions/" + Pattern.quote(target) + "/.*.pasta"))
```
sabieber, line 141: s.a.

sabieber (changes requested):

src/main/java/sirius/pasta/tagliatelle/Tagliatelle.java
andyHa: this isn't constant as "target" is passed in - however the result of the method call is cached...

src/main/java/sirius/pasta/tagliatelle/Tagliatelle.java
andyHa: this isn't constant as "target" is passed in - however the result of the method call is cached...

sabieber (approved):

idlira (approved):

qw3ry (approved):

PR #1015, opened 2022-03-29
Removes a broadly applied style and replaces it with a more local limit.

andyHa:
The previous change was also intended to limit the card title, however,
it screwed up icons etc. by applying a min-width on almost any element
within a card.

We now provide a proper overflow to limit the title width.
(Previous ticket: SIRI-470)

Fixes: OX-8388

### DIFF
--- a/src/main/resources/default/assets/tycho/styles/page.scss
+++ b/src/main/resources/default/assets/tycho/styles/page.scss
@@ -10,17 +10,6 @@ body {
   background-color: $backgroundGray;
 }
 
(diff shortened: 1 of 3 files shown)


sabieber (commented):
This still works for all previous places where we had to limit the length of card elements?

andyHa:
> This still works for all previous places where we had to limit the length of card elements?

at least with those linked in the original ticket and the one at hands...

sabieber (approved):

mkeckmkeck (approved):

jakobvogel (approved):

PR #1020, opened 2022-04-06
Records and outputs contention blocks.

andyHa:
If a channel is not writeable, we actually block and await for it to
become writable. As we figured, this might slow down tunneling in
high load conditions, we added some loggings to check if there is a
correlations.


sabieber (approved):

src/main/java/sirius/web/http/WebServer.java
```diff
@@ -758,6 +760,11 @@ public void gather(MetricsCollector collector) {
                                      "HTTP Bytes-Out",
                                      bytesOut.get() / 1024d / 60,
                                      "KB/s");
+        collector.differentialMetric("http_contention_blocks",
+                                     "http-contention-blocks",
+                                     "HTTP ConentionBlocks",
```
idlira, line 765:
```suggestion
                                     "HTTP ConnectionBlocks",
```

idlira (approved):

idlira (approved):

PR #1028, opened 2022-04-13
Console: No Message Box (SIRI-559)

jakobvogel:
> https://scireum.myjetbrains.com/youtrack/issue/SIRI-559

### DIFF
--- a/src/main/resources/default/templates/system/console.html.pasta
+++ b/src/main/resources/default/templates/system/console.html.pasta
@@ -63,6 +63,10 @@
                 max-height: 100%;
             }
 
+            #wrapper-body > #page-header, #wrapper-body > #main-container > .row {
+                display: none;
+            }
+
             #term {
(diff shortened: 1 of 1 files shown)


meg1502 (approved):

src/main/resources/default/templates/system/console.html.pasta
```diff
@@ -59,10 +59,14 @@
                 overflow: hidden;
             }
 
-            #main-container {
+            #wrapper-body > #main-container {
                 max-height: 100%;
             }
 
+            #wrapper-body > #page-header, #wrapper-body > #main-container > .row, #wrapper-body > #main-container > .row > #message-box {
```
andyHa, line 66: We _could_also overwrite the messages block like `<i:block name="messages"></i:block>` This would disable
the generation of the messages all together instead of creating and the hiding them...

PR #1041, opened 2022-05-20
BREAKING: Aligns behaviour of postJSON with the one of replaced jQuery AJAX

sabieber:
**BREAKING CHANGE:** This is potentially breaking when existing calls to `postJSON` are made with `undefined` parameters that are expected to be passed on the server-side.
But current calls to `postJSON` in our projects don't seem to pass empty parameters.

This is quite useful when building complex request payloads as no extensive filtering has to occur outside of the postJSON call.

Fixes: OX-8650

### DIFF
--- a/src/main/resources/default/assets/common/core.js.pasta
+++ b/src/main/resources/default/assets/common/core.js.pasta
@@ -135,7 +135,7 @@ sirius.getJSON = function (url, params) {
  * Calls the given URL, posts the given params and retrieves the resulting JSON.
  *
  * @param url the URL to invoke
- * @param params the parameters to send POST data
+ * @param params the parameters to send POST data (parameters with 'undefined' values are not send to the server)
  * @returns {Promise<any>} the received JSON data
  */
 sirius.postJSON = function (url, params) {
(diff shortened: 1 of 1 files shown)


idlira (approved):

andyHa (approved):

src/main/resources/default/assets/common/core.js.pasta
```diff
@@ -135,13 +135,16 @@ sirius.getJSON = function (url, params) {
  * Calls the given URL, posts the given params and retrieves the resulting JSON.
  *
  * @param url the URL to invoke
- * @param params the parameters to send POST data
+ * @param params the parameters to send POST data (parameters with 'undefined' values are not send to the server)
```
cschierle, line 138: 😛 
```suggestion
 * @param params the parameters to send POST data (parameters with 'undefined' values are not sent to the server)
```

cschierle (approved):

meg1502 (approved):

src/main/resources/default/assets/common/core.js.pasta
meg1502: being sent? 🤷‍♂️

fhaScireum (approved):

idlira (approved):

PR #1046, opened 2022-06-02
Object.entries polyfill

marvinwoop:
Polyfill for Object.entires

### DIFF
--- a/src/main/resources/default/assets/common/common-polyfills.js.pasta
+++ b/src/main/resources/default/assets/common/common-polyfills.js.pasta
@@ -185,15 +185,15 @@ if (typeof Object.assign != 'function') {
 * see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries#polyfill
 */
 if (!Object.entries) {
-    Object.entries = function(obj) {
-        var ownProps = Object.keys(obj),
-            i = ownProps.length,
-            resArray = new Array(i); // preallocate the Array
+    Object.entries = function (obj) {
+        const keys = Object.keys(obj);
+        const result = new Array(keys.length);
+        let index = keys.length;
 
-        while (i--) {
-            resArray[i] = [ownProps[i], obj[ownProps[i]]];
+        while (index-- > 0) {
+            result[index] = [keys[index], obj[keys[index]]];
         }
 
-        return resArray;
+        return result;
     };
 }
--- a/src/main/resources/default/assets/common/common-polyfills.js.pasta
+++ b/src/main/resources/default/assets/common/common-polyfills.js.pasta
@@ -179,3 +179,21 @@ if (typeof Object.assign != 'function') {
         configurable: true
     });
 }
+
+/**@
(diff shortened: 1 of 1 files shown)


src/main/resources/default/assets/common/common-polyfills.js.pasta
```diff
@@ -179,3 +179,21 @@ if (typeof Object.assign != 'function') {
...
+
+/**@
+* Polyfill for the Object.entries function
+* see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries#polyfill
+*/
+if (!Object.entries) {
+    Object.entries = function(obj) {
+        var ownProps = Object.keys(obj),
+            i = ownProps.length,
+            resArray = new Array(i); // preallocate the Array
```
andyHa, line 191:
```suggestion
        const keys = Object.keys(obj);
        const result = new Array(keys.length);
        
        let index = keys.length;
```

src/main/resources/default/assets/common/common-polyfills.js.pasta
```diff
@@ -179,3 +179,21 @@ if (typeof Object.assign != 'function') {
...
+
+        while (i--) {
+            resArray[i] = [ownProps[i], obj[ownProps[i]]];
+        }
+
+        return resArray;
```
andyHa, line 197:
```suggestion
        while (index-- > 0) {
            result[index] = [keys[index], obj[keys[index]]];
        }

        return result;
```

src/main/resources/default/assets/common/common-polyfills.js.pasta
andyHa: I'd prefer let/const over var....

andyHa (changes requested):

src/main/resources/default/assets/common/common-polyfills.js.pasta
marvinwoop: The current usage of var enables us to use Object.keys in IE versions > 8. If we use const / let, we'd only support IE versions > 10 (effectively only 11). So the question is: Would it be enough to only support IE11?

src/main/resources/default/assets/common/common-polyfills.js.pasta
marvinwoop: `index--` and `index-- > 0' are effectively the same as the `0` is interpreted as false. Or did you suggest to change it because of readability?

src/main/resources/default/assets/common/common-polyfills.js.pasta
andyHa: yep

src/main/resources/default/assets/common/common-polyfills.js.pasta
andyHa: for readability

fhaScireum (approved):

sabieber (approved):

mkeckmkeck (approved):

PR #1055, opened 2022-06-24
Only removes a chunked write handler if it has been properly installed before.

andyHa:
This prevents an NPE in edge cases.

### DIFF
--- a/src/main/java/sirius/web/http/Response.java
+++ b/src/main/java/sirius/web/http/Response.java
@@ -835,6 +835,15 @@ public class Response {
         }
     }
 
+    protected void removedChunkedWriteHandler(ChannelFuture writeFuture) {
+        writeFuture.addListener(ignored -> {
+            if (ctx.channel().pipeline().get(ChunkedWriteHandler.class) != null && ctx.channel().isOpen()) {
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/web/http/Response.java
```diff
@@ -835,6 +835,15 @@ protected void installChunkedWriteHandler() {
         }
     }
 
+    protected void removedChunkedWriteHandler(ChannelFuture writeFuture) {
+        writeFuture.addListener(ignored -> {
+            if (ctx.channel().pipeline().get(ChunkedWriteHandler.class) != null && ctx.channel().isOpen()) {
```
idlira, line 840: should we really check if the channel is open? If the client closes the channel will close but the chunkwriter wont be removed?

idlira (changes requested):

src/main/java/sirius/web/http/Response.java
andyHa: die pipeline + channel + everything will be discarded anyway...

idlira (approved):

sabieber (approved):

mkeckmkeck (approved):

jmuscireum (approved):

PR #1079, opened 2022-08-15
Fix for IE11, which does not support for-of

qw3ry:
Breaking: `sirius.getFormData` -> `getFormData`
Careful: This is used in sirius biz!

### DIFF
--- a/src/main/resources/default/assets/tycho/scripts/form.js.pasta
+++ b/src/main/resources/default/assets/tycho/scripts/form.js.pasta
@@ -44,3 +44,22 @@ sirius.ready(function () {
     });
 });
 
+/**@
+ * Extracts the values of all form elements inside a form.
+ * <p>
+ * This attempts to replace jQuery's serialize() / serializeArray() method.
+ *
+ * @param _node the DOM-node of the <form> to get the values from
+ * @returns a map of name-value pairs, mapping the name of the input to its value
+ */
+function getFormData(_node) {
+    const object = {};
+    var entries = new FormData(_node).entries();
+    var entry = entries.next();
+    while (!entry.done) {
+        object[entry.value[0]] = entry.value[1];
+        entry = entries.next();
+    }
+    return object;
+}
+
--- a/src/main/resources/default/assets/common/core.js.pasta
+++ b/src/main/resources/default/assets/common/core.js.pasta
@@ -168,26 +168,6 @@ sirius.postJSON = function (url, params) {
 }
 
 /**@
- * Extracts the values of all form elements inside a form.
- * <p>
- * This attempts to replace jQuery's serialize() / serializeArray() method.
- *
- * @param _node the DOM-node of the <form> to get the values from
- * @returns a map of name-value pairs, mapping the name of the input to its value
- */
-sirius.getFormData = function (_node) {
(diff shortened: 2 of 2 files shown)


jakobvogel (commented):
Laut https://developer.mozilla.org/en-US/docs/Web/API/FormData/keys wird der IE11 das auch nicht können? Die Frage ist, ob wir die Methode auf IE11 überhaupt brauchen, oder ob es uns genügt, dass er es nur parsen kann?

qw3ry:
> Laut https://developer.mozilla.org/en-US/docs/Web/API/FormData/keys wird der IE11 das auch nicht können? Die Frage ist, ob wir die Methode auf IE11 überhaupt brauchen, oder ob es uns genügt, dass er es nur parsen kann?

oh thats bad :( So we either decide we don't care (and close this PR), look for an appropriate FormData Polyfill, or find _some_ other way to compute this. We could also use jQuery here, but iirc we are trying to get rid of it (as Bootstrap 5 does not include it anymore...)

jakobvogel:
> > Laut https://developer.mozilla.org/en-US/docs/Web/API/FormData/keys wird der IE11 das auch nicht können? Die Frage ist, ob wir die Methode auf IE11 überhaupt brauchen, oder ob es uns genügt, dass er es nur parsen kann?
> 
> 
> 
> oh thats bad :( So we either decide we don't care (and close this PR), look for an appropriate FormData Polyfill, or find _some_ other way to compute this. We could also use jQuery here, but iirc we are trying to get rid of it (as Bootstrap 5 does not include it anymore...)

Right now, without this PR, the code does not even parse in IE11. We need some fix as OX integrations currently do not even load on IE11. (But need to!) If you don't need the new method on IE11 I would rather go with this PR as it should at least parse, other than the for-of-construct. Else, we need a proper solution 😊

qw3ry:
> Right now, without this PR, the code does not even parse in IE11. We need some fix as OX integrations currently do not even load on IE11. (But need to!) If you don't need the new method on IE11 I would rather go with this PR as it should at least parse, other than the for-of-construct. Else, we need a proper solution 😊

I think this PR will suffice. The only usages (I'm aware of) affect only the Backend (Jobs), so it shouldn't be much of a problem.

jakobvogel (approved):
Provided this method is not required in a setting that needs to be supported by IE11.

jakobvogel:
In Absprache mit @andyHa: Bitte versuche, IE11-kompatibel umzubauen, siehe etwa https://stackoverflow.com/questions/37938955/iterating-through-formdata-in-ie

src/main/resources/default/assets/tycho/scripts/form.js.pasta
```diff
@@ -44,3 +44,22 @@ sirius.ready(function () {
...
+function getFormData(_node) {
+    const object = {};
+    var entries = new FormData(_node).entries();
+    var entry = entries.next();
+    while (!entry.done) {
+        object[entry.value[0]] = entry.value[1];
+        entry = entries.next();
+    }
+    return object;
+}
```
jakobvogel, line 64: We can go back to the shorter `for…of` if we are excluding IE11 anyways

jakobvogel (approved):

PR #1081, opened 2022-08-18
Trimming of Parsed HTML Templates (SIRI-618)

jakobvogel:
> https://scireum.myjetbrains.com/youtrack/issue/SIRI-618

- Trims HTML and XML templates after parsing, in order to enable punctuation without whitespace after custom tags
- **JavaDoc will be done** once we agree on the entire thing 😅

### DIFF
--- a/src/main/java/sirius/pasta/tagliatelle/compiler/TemplateCompiler.java
+++ b/src/main/java/sirius/pasta/tagliatelle/compiler/TemplateCompiler.java
@@ -253,14 +253,11 @@ public class TemplateCompiler extends InputProcessor {
      * <p>
      * If the tag is built-in (<tt>i:</tt>) or one in a taglib, an appropriate {@link TagHandler} is created and
      * invoked. Otherwise, the tag is parsed as static text.
-     * <p>
-     * Note that the return value indicates whether an emitter was created, thus indicating the need to create a new
-     * {@link ConstantEmitter}.
      *
      * @param parentHandler the outer tag handler
      * @param block         the block to which the tag should be added
      * @param staticText    the emitter which is responsible for consuming static text
-     * @return <b>true</b> if an {@link Emitter} was generated, <b>false</b> otherwise
+     * @return <b>true</b> if the tag was handled, <b>false</b> else
      */
     private boolean processTag(TagHandler parentHandler, CompositeEmitter block, ConstantEmitter staticText) {
         if (!reader.current().is('<')) {
--- a/src/main/java/sirius/pasta/tagliatelle/compiler/TemplateCompiler.java
+++ b/src/main/java/sirius/pasta/tagliatelle/compiler/TemplateCompiler.java
@@ -28,6 +28,7 @@ import sirius.pasta.noodle.compiler.NoodleCompiler;
 import sirius.pasta.tagliatelle.emitter.CompositeEmitter;
 import sirius.pasta.tagliatelle.emitter.ConstantEmitter;
 import sirius.pasta.tagliatelle.emitter.Emitter;
+import sirius.pasta.tagliatelle.tags.ArgTag;
 import sirius.pasta.tagliatelle.tags.TagHandler;
 import sirius.web.services.JSONStructuredOutput;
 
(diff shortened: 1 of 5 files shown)


src/main/java/sirius/pasta/tagliatelle/emitter/ConstantEmitter.java
```diff
@@ -62,6 +62,33 @@ public ConstantEmitter append(String stringToAppend) {
...
+    }
+
+    public ConstantEmitter strip() {
+        if (Strings.isEmpty(value)) {
+            return this;
+        }
+
+        value = value.strip();
+        return this;
+    }
```
qw3ry, line 90: unused, I think?

src/main/java/sirius/pasta/tagliatelle/emitter/ConstantEmitter.java
```diff
@@ -62,6 +62,33 @@ public ConstantEmitter append(String stringToAppend) {
...
+
+    public ConstantEmitter stripTrailing() {
+        if (Strings.isEmpty(value)) {
+            return this;
+        }
+
+        value = value.stripTrailing();
+        return this;
+    }
```
qw3ry, line 81: We could instead add this methods to `Emitter` directly as NOOP, which would avoid the `instanceof` checking.
Would probably be a bit cleaner, but on the expense of more boilerplate code; you decide.

qw3ry (approved):

meg1502 (approved):

mko-sci (approved):

src/main/java/sirius/pasta/tagliatelle/emitter/ConstantEmitter.java
jakobvogel: Yeah, I am just not sure whether the respective methods would have semantic significance within `Emitter`. Let's see, I am still thinking whether we really should strip all the whitespace, or whether it is enough (and safer) to only omit the last line break.

src/main/java/sirius/pasta/tagliatelle/emitter/ConstantEmitter.java
jakobvogel: Yes, it is just for the sake of completeness 😅

src/main/java/sirius/pasta/tagliatelle/compiler/TemplateCompiler.java
```diff
@@ -79,7 +79,16 @@ public List<CompileError> compile() throws CompileException {
         }
 
         try {
-            Emitter emitter = parseBlock(null, null).reduce();
+            // identify content type via file extension, as is done in Template.setupEscaper(GlobalRenderContext), in
+            // order to decide whether to trim whitespace after parsing
+            String effectiveFileName = getContext().getTemplate().getEffectiveFileName();
+            boolean trimWhitespace = effectiveFileName.endsWith(".html") || effectiveFileName.endsWith(".xml");
```
andyHa, line 85: Template#setupEscaper has already a similar check, maybe we could extract this as Template#isXmlContentExpected

src/main/java/sirius/pasta/tagliatelle/compiler/TemplateCompiler.java
```diff
@@ -79,7 +79,16 @@ public List<CompileError> compile() throws CompileException {
...
+
+            CompositeEmitter compositeEmitter = parseBlock(null, null);
+            if (trimWhitespace) {
+                compositeEmitter.stripWhitespace();
+            }
```
andyHa, line 90: maybe instead of first adding and then trimming - we could add a parameter to parseBlock and "simply" skip leading and trailing whitespace

src/main/java/sirius/pasta/tagliatelle/compiler/TemplateCompiler.java
jakobvogel: done

src/main/java/sirius/pasta/tagliatelle/emitter/ConstantEmitter.java
jakobvogel: removed

src/main/java/sirius/pasta/tagliatelle/emitter/ConstantEmitter.java
jakobvogel: reworked

andyHa (approved):

qw3ry (approved):

idlira (approved):

fhaScireum (approved):

PR #1090, opened 2022-09-01
Exposes pooled byte buffer mem usage

idlira:
Exposes the current heap and direct memory used by the pooled byte buffer allocation, converted in MB.

Note that we are not defining threshold limits in the config (they are initialized with zero), and system can set those if needed as they strongly depend on the actual amount of memory allocated to both heap and direct memory.

Fixes: [OX-9027](https://scireum.myjetbrains.com/youtrack/issue/OX-9027)

### DIFF
--- a/src/main/java/sirius/web/http/WebServer.java
+++ b/src/main/java/sirius/web/http/WebServer.java
@@ -26,6 +26,7 @@ import io.netty.handler.codec.http.multipart.DiskAttribute;
 import io.netty.handler.codec.http.multipart.DiskFileUpload;
 import io.netty.handler.codec.http.multipart.HttpDataFactory;
 import io.netty.util.ResourceLeakDetector;
+import io.netty.util.internal.PlatformDependent;
 import sirius.kernel.Killable;
 import sirius.kernel.Sirius;
 import sirius.kernel.Startable;
@@ -838,6 +839,11 @@ public class WebServer implements Startable, Stoppable, Killable, MetricProvider
                          "pooled-byte-buffer-used-direct-mem",
                          "Pooled byte buffer allocation used direct memory",
                          PooledByteBufAllocator.DEFAULT.metric().usedDirectMemory() / 1024d / 1024d,
+                         "MB");
+        collector.metric("max_direct_mem",
+                         "max-direct-mem",
+                         "Maximum direct memory",
+                         PlatformDependent.maxDirectMemory() / 1024d / 1024d,
                          "MB");
     }
 
--- a/src/main/java/sirius/web/http/WebServer.java
+++ b/src/main/java/sirius/web/http/WebServer.java
@@ -829,6 +829,16 @@ public class WebServer implements Startable, Stoppable, Killable, MetricProvider
                          queueTime.getAndClear(),
                          "ms");
         collector.metric("http_websockets", "http-websockets", "Open Websockets", websockets.get(), null);
+        collector.metric("pooled_byte_buffer_used_heap_mem",
+                         "pooled-byte-buffer-used-heap-mem",
+                         "Pooled byte buffer allocation used heap memory",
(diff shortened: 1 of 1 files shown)


src/main/java/sirius/web/http/WebServer.java
```diff
@@ -829,6 +829,16 @@ public void gather(MetricsCollector collector) {
                          queueTime.getAndClear(),
                          "ms");
         collector.metric("http_websockets", "http-websockets", "Open Websockets", websockets.get(), null);
+        collector.metric("pooled_byte_buffer_used_heap_mem",
+                         "pooled-byte-buffer-used-heap-mem",
+                         "Pooled byte buffer allocation used heap memory",
+                         PooledByteBufAllocator.DEFAULT.metric().usedHeapMemory() / 1024d / 1024d,
```
jakobvogel, line 835: 1024 -> MiB, 1000 -> MB … please divide by 1000 as we usually do, following SI units

```suggestion
                         PooledByteBufAllocator.DEFAULT.metric().usedHeapMemory() / 1000d / 1000d,
```

src/main/java/sirius/web/http/WebServer.java
```diff
@@ -829,6 +829,16 @@ public void gather(MetricsCollector collector) {
...
+                         "MB");
+        collector.metric("pooled_byte_buffer_used_direct_mem",
+                         "pooled-byte-buffer-used-direct-mem",
+                         "Pooled byte buffer allocation used direct memory",
+                         PooledByteBufAllocator.DEFAULT.metric().usedDirectMemory() / 1024d / 1024d,
```
jakobvogel, line 840:
```suggestion
                         PooledByteBufAllocator.DEFAULT.metric().usedDirectMemory() / 1000d / 1000d,
```

src/main/java/sirius/web/http/WebServer.java
idlira: That will give unclear values, compared to how one passes settings to the JVM, such as -XX:MaxDirectMemory=16g
If so, happy to change the unit from MB to MiB. What others say?

idlira: Other metrics are also calculated this way. See https://github.com/scireum/sirius-db/blob/03a1f8ab9cf2ab91e839753bae3b66f0e925cad0/src/main/java/sirius/db/redis/RedisMetricProvider.java

jakobvogel: Well, that is not good 😅 ... I don't mind to write MiB, but the prefixes need to match the factors, and we should fix it everywhere. 🙃

idlira: Go tell Microsoft that Windows calculates files wrong, since, EVER! 🤣

src/main/java/sirius/web/http/WebServer.java
andyHa, line 840: Should we add PlatformDependent.getMaxDirectMemory (or the like) as additional value while we're at it?

sabieber (approved):

src/main/java/sirius/web/http/WebServer.java
idlira: Will track in a follow up ticket.

src/main/java/sirius/web/http/WebServer.java
idlira: Will track in a follow up ticket.

src/main/java/sirius/web/http/WebServer.java
idlira: [SIRI-625](https://scireum.myjetbrains.com/youtrack/issue/SIRI-625)

andyHa (approved):

fhaScireum (approved):

PR #1112, opened 2022-10-07
Adds a new class to properly style columns which only contain an icon.

andyHa:
Most notably this is the delete button in tables.

Fixes: SIRI-640


fhaScireum (approved):

andyHa:
Nevenmind - "delete-btn-colunm"

PR #1125, opened 2022-11-16
Macro for Inlining SVG Files

jakobvogel:
> [OX-9299](https://scireum.myjetbrains.com/youtrack/issue/OX-9299)

Or combine with `inlineResource`?

### DIFF
--- a/src/main/java/sirius/pasta/noodle/macros/InlineSvgMacro.java
+++ b/src/main/java/sirius/pasta/noodle/macros/InlineSvgMacro.java
@@ -1,0 +1,129 @@
+/*
+ * Made with all the love in the world
+ * by scireum in Remshalden, Germany
+ *
+ * Copyright by scireum GmbH
+ * http://www.scireum.de - info@scireum.de
+ */
+
+package sirius.pasta.noodle.macros;
+
+import org.w3c.dom.Document;
+import org.w3c.dom.Element;
+import sirius.kernel.commons.Strings;
+import sirius.kernel.di.std.Part;
+import sirius.kernel.di.std.Register;
+import sirius.kernel.health.Exceptions;
+import sirius.kernel.tokenizer.Position;
+import sirius.pasta.noodle.Environment;
+import sirius.pasta.noodle.compiler.CompilationContext;
+import sirius.pasta.noodle.compiler.ir.Node;
+import sirius.web.resources.Resource;
+import sirius.web.resources.Resources;
+
+import javax.annotation.Nonnull;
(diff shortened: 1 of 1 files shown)


sabieber (approved):

src/main/java/sirius/pasta/noodle/macros/InlineSvgMacro.java
```diff
@@ -0,0 +1,129 @@
...
+            DocumentBuilderFactory factory = DocumentBuilderFactory.newInstance();
+            return factory.newDocumentBuilder().parse(stream);
+        } catch (Exception exception) {
+            throw Exceptions.handle(exception);
+        }
+    }
+
+    private String stringifyElement(Element element) {
+        try (StringWriter writer = new StringWriter()) {
+            TransformerFactory factory = TransformerFactory.newInstance();
```
idlira, line 108: Can we strip this variable and just use `TransformerFactory.newInstance().newTransformer()` below? Or maybe even better, lazy-load it one inside the part?

idlira (changes requested):

src/main/java/sirius/pasta/noodle/macros/InlineSvgMacro.java
jakobvogel: This is pretty much the standard approach we use elsewhere, too, for instance when parsing a SAML response. I am not sure whether it is worth the effort to optimise away this single variable, considering that we are loading files from disk and sending them away over the network. (And I would expect the JIT to do that automatically anyway?)

idlira: But each call will create a new factory, that could be done once as long as the part lives.
Something like that would improve it a bit, especially if this macro is called lots of times?
```
TransformerFactory factory = null;
function fetchTransformerFactory() {
  if (factory == null) {
    this.factory = TransformerFactory.newInstance();
  }
  return factory;
}
```

sabieber: My 2 ct: Most of the time the argument passed to this macro will be constant (as in a string with the path to the SVG) in which case I would guess Tagliatelle can optimize caching and wont invoke the macro all of the time in those places.

andyHa: I understand the concerns, but given the caching/constant nature of most calls, I'd go with the current implementation in the name of simplicity ;-)

andyHa (approved):

PR #1136, opened 2022-12-02
Error Check in Base64 Macro (SIRI-597)

jakobvogel:
> [SIRI-597](https://scireum.myjetbrains.com/youtrack/issue/SIRI-597)

### DIFF
--- a/src/main/java/sirius/pasta/noodle/macros/Base64ResourceMacro.java
+++ b/src/main/java/sirius/pasta/noodle/macros/Base64ResourceMacro.java
@@ -83,10 +83,8 @@ public class Base64ResourceMacro extends BasicMacro implements SassFunction {
             throw new IllegalArgumentException("Only assets can be inlined for security reasons.");
         }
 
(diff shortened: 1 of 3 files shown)


src/main/java/sirius/pasta/noodle/macros/Base64ResourceMacro.java
```diff
@@ -99,7 +100,12 @@ public String eval(FunctionCall call) {
             path = path.substring(1, path.length() - 1);
         }
 
-        return encodeResource(path);
+        String code = encodeResource(path);
+        if (Strings.isEmpty(code)) {
+            throw new IllegalArgumentException("Unknown resource: " + path);
```
jakobvogel, line 105: Alternatively, we could also throw an exception in `encodeResource(String)` above, instead of returning an empty string. That would also affect the Tagliatelle part, though, and not just the SASS part. Opinions?

sabieber (approved):

jakobvogel:
Merging based on previous reviews

PR #1154, opened 2023-01-23
API Docs: various enhancements

cschierle:
For more details, see the individual commits.

The most visible new feature is a table of contents for all documented services within an api:
Bildschirm­foto 2023-01-23 um 10 50 32

The sidebar sticks to the top, when scrolling through the page content:
Bildschirm­foto 2023-01-23 um 10 53 43


src/main/resources/default/templates/system/api.html.pasta
```diff
@@ -92,7 +92,7 @@
                                     <td>
                                         <i:raw>@expandMessage(smartTranslate(parameter.description()))</i:raw>
                                     </td>
-                                    <td>@toUserString(parameter.required())</td>
+                                    <td>@parameter.required()</td>
```
sabieber: To improve the rendering a little bit we could maybe choose two unicode chars for true/false
Something like ✓ and ✗ for example
Or we could use a read-only checkbox element

sabieber (approved):

jakobvogel (approved):
Assuming the Pasta file works 😅

src/main/resources/default/templates/system/api.html.pasta
cschierle: Feels to me like not much benefit, but if there are more votes for this, I can add it

PR #1160, opened 2023-01-29
Adds the optional argument to the booleanSelect tag in order to be able to enforce a user to actively select a state

mko-sci:
Fixes: [9321](https://scireum.myjetbrains.com/youtrack/issue/OX-9321/Video-Modernisierung-Backend-User-Interface-Teil-3-Weitere-Seiten)

### DIFF
--- a/src/main/resources/default/taglib/t/booleanSelect.html.pasta
+++ b/src/main/resources/default/taglib/t/booleanSelect.html.pasta
@@ -11,6 +11,7 @@
 <i:arg name="helpKey" type="String" default=""/>
 <i:arg name="help" type="String" default="@i18n(helpKey)"/>
 <i:arg name="readonly" type="boolean" default="false"/>
+<i:arg name="optional" type="boolean" default="false"/>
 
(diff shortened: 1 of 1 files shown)


jakobvogel (approved):
Assuming that the deleted `@` chars do not matter

idlira (approved):

mko-sci:
> Assuming that the deleted `@` chars do not matter

yep, within the @selected it does not matter

andyHa (approved):

andyHa:
> Assuming that the deleted `@` chars do not matter

yep, Tagliatelle knows that an expression is expected - also it knows that the result has to evaluate to boolean so it derives that the input cannot be a string literal. The @ is only required if a tag attribute accepts a string but an expression or variable is passed in. In this case, Tagliatelle cannot determine if the string or the evaluated expression should be passed in - therefore an @ is required to mark expressions....

fhaScireum (approved):

mkeckmkeck (approved):

PR #1161, opened 2023-01-30
Overhauls the security sandbox of Noodle

andyHa:
We now have a system wide setting if the sandbox is enabled at all.
Additionally one or more SandboxDetectors can determine if a template
is to be sandboxed or not.

Additionally, instead of just enabling or disabling the sandbox, we now
also support a "WARN_ONLY" mode, which enables the sandbox but
only reports warnings.

Finally the rules and the way we evaluate them (along with annotations)
have been made a bit more flexible (and some standard have been
added).

### DIFF
--- a/src/main/java/sirius/pasta/noodle/compiler/ir/ConstructorCall.java
+++ b/src/main/java/sirius/pasta/noodle/compiler/ir/ConstructorCall.java
@@ -8,19 +8,19 @@
 
 package sirius.pasta.noodle.compiler.ir;
 
-import sirius.kernel.tokenizer.Position;
 import sirius.kernel.di.std.Part;
 import sirius.kernel.health.Exceptions;
 import sirius.kernel.health.Log;
+import sirius.kernel.tokenizer.Position;
 import sirius.pasta.noodle.OpCode;
 import sirius.pasta.noodle.compiler.Assembler;
 import sirius.pasta.noodle.compiler.CompilationContext;
 import sirius.pasta.noodle.compiler.TypeTools;
 import sirius.pasta.noodle.sandbox.Sandbox;
+import sirius.pasta.noodle.sandbox.SandboxMode;
 
 import javax.annotation.Nullable;
 import java.lang.reflect.Constructor;
-import java.lang.reflect.Method;
 import java.lang.reflect.Type;
 import java.lang.reflect.TypeVariable;
 import java.util.Arrays;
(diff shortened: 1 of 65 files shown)


src/main/java/sirius/pasta/noodle/sandbox/NoodleSandbox.java
```diff
@@ -21,5 +21,23 @@
  */
 @Retention(RetentionPolicy.RUNTIME)
 @Target({ElementType.METHOD, ElementType.TYPE, ElementType.FIELD})
-public @interface PublicApi {
+public @interface NoodleSandbox {
+
+    /**
+     * Controls the accessibility of the annotated element.
+     * <p>
+     * <tt>GRANTED</tt> is probably the default case, however, if a whole class is granted, <tt>REJECTED</tt>
```
mkeckmkeck, line 29:
```suggestion
     * <tt>GRANTED</tt> might be the correct value for most cases, however, if a whole class is granted, <tt>REJECTED</tt>
```

src/main/java/sirius/pasta/noodle/macros/PartMacro.java
```diff
@@ -27,7 +26,6 @@
  * this is a constant call and will be optimized away entirely.
  */
 @Register
```
mkeckmkeck, line 28: We currently use this at artefacts. Current usages could probably be changed/gitified. But for future quickfixes keeping this public would probably be helpful.

src/main/java/sirius/pasta/noodle/compiler/ir/ConstructorCall.java
```diff
@@ -100,12 +100,18 @@ private void checkDeprecation(CompilationContext context) {
...
 
     private void checkSandbox(CompilationContext context) {
-        if (context.isSandboxEnabled() && !sandbox.canInvoke(constructor)) {
-            context.error(position,
-                          "Cannot invoke the constructor on %s as this is restricted by the security sandbox!",
-                          constructor.getDeclaringClass().getName());
-            this.constructor = null;
-        }
+        context.ifEnforceSandbox(() -> !sandbox.canInvoke(constructor), mode -> {
+            if (mode == SandboxMode.ENABLED) {
```
idlira, line 104: If not overkill, can we generalize this? Feels like the same check is used in several places to either call context.error x context.warning.

src/main/java/sirius/pasta/noodle/sandbox/Sandbox.java
```diff
@@ -41,62 +101,67 @@ public class Sandbox {
...
     }
 
-    private boolean isAllowedViaAnnotation(Executable method) {
-        if (method.isAnnotationPresent(PublicApi.class)) {
-            return true;
-        }
-        if (method.getDeclaringClass().isAnnotationPresent(PublicApi.class)) {
-            return true;
+    private String cleanupKey(String key) {
+        if (key.startsWith("")) {
```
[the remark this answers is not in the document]
andyHa: BS is happening - the check has been fixed. Sadly all strings start with "" so the code seemed to work as intended :(

src/main/java/sirius/pasta/noodle/macros/PartMacro.java
andyHa: done

src/main/java/sirius/pasta/noodle/compiler/ir/ConstructorCall.java
andyHa: Sadly the three usages all differ each a little bit (having side-effects on error etc.)

PR #1163, opened 2023-01-31
Adds a dialog component

sabieber:
This behaves similar to the overlay component but in a more blocking way and also provides some helpers for displaying a simple message, a preloader, ...

image

Sample usage:
```javascript
    sirius.dialog.create({
        translationProvider: (key) => {
            switch (key) {
                case 'close': return 'Schließen';
                default: return key;
            }
        }
    });

    sirius.dialog.fillWithPreloader('Demo Anfrage', 'Daten werden geladen ...');

    setTimeout(() => {
        sirius.dialog.fillWithMessage('Demo Anfrage', 'Daten nicht vorhanden.');
        sirius.dialog.clearFooter();
        sirius.dialog.addFooterButton('Test 1', () => window.alert('Test 1'));
        sirius.dialog.addFooterButton('Test 2', () => window.alert('Test 2'));
        sirius.dialog.addFooterButton('Test 3', () => window.alert('Test 3'));
    }, 2000);
```

Fixes: [OX-9375](https://scireum.myjetbrains.com/youtrack/issue/OX-9375)


andyHa (approved):

idlira (approved):

fhaScireum (approved):

jakobvogel (commented):
Do we really want this kind-of-singleton-pattern? I understand we do this quite often, see for instance the overlay, but I do often have the feeling that an object would be more versatile?

sabieber:
> Do we really want this kind-of-singleton-pattern? I understand we do this quite often, see for instance the overlay, but I do often have the feeling that an object would be more versatile?

Could be discussed, for sure.
I find it a bit more forgiving in the usage of the API as we help to clean up old dialogs when just opening a new dialog without manually destroying the previous one.
Also I don't really see a use case for having multiple dialogs open at the same time.

jakobvogel (approved):

mkeckmkeck (approved):

PR #1167, opened 2023-02-06
Whitelists the inlineResource macro

jmuscireum:
This allows the inlineResource macro in sandboxes, but only if static resources are referenced. It should never be required for user-created templates to inline tagliatelle resources.

### DIFF
--- a/src/main/java/sirius/pasta/noodle/macros/InlineResourceMacro.java
+++ b/src/main/java/sirius/pasta/noodle/macros/InlineResourceMacro.java
@@ -8,13 +8,15 @@
 
 package sirius.pasta.noodle.macros;
 
-import sirius.kernel.tokenizer.Position;
 import sirius.kernel.commons.Strings;
 import sirius.kernel.di.std.Part;
 import sirius.kernel.di.std.Register;
+import sirius.kernel.tokenizer.Position;
 import sirius.pasta.noodle.Environment;
 import sirius.pasta.noodle.compiler.CompilationContext;
 import sirius.pasta.noodle.compiler.ir.Node;
+import sirius.pasta.noodle.sandbox.NoodleSandbox;
+import sirius.pasta.noodle.sandbox.SandboxMode;
 import sirius.pasta.tagliatelle.Tagliatelle;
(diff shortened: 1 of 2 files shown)


sabieber (approved):

mkeckmkeck (approved):

src/main/java/sirius/pasta/noodle/macros/InlineResourceMacro.java
```diff
@@ -68,6 +73,26 @@ public void verify(CompilationContext context, Position position, List<Node> arg
                 context.warning(position, "Unknown resource: %s", resourceName);
             }
         }
+
+        if (context.getSandboxMode() != SandboxMode.DISABLED) {
+            if (!args.get(0).isConstant()) {
+                raiseSandboxRestriction(context,
+                                        position,
+                                        "Only constant resource paths are supported due to sandbox restrictions.");
+            } else if (String.valueOf(args.get(0).getConstantValue()).endsWith(PASTA_SUFFIX)) {
```
andyHa, line 82: shouldn't be a ! in there?

andyHa: and shouldn't we check like "either .pasta" or "/assets" ?

andyHa: (also, don't we have such a check elsewhere in Tagliatelle?

src/main/java/sirius/pasta/noodle/macros/InlineResourceMacro.java
```diff
@@ -68,6 +73,26 @@ public void verify(CompilationContext context, Position position, List<Node> arg
...
+                raiseSandboxRestriction(context,
+                                        position,
+                                        "Only constant resources are supported due to sandbox restrictions.");
```
andyHa, line 85: shouldn't this be more specific like "only pasta templates...."

andyHa (changes requested):

andyHa (approved):

PR #1168, opened 2023-02-08
Response: pass provided cache headers to redirect response

cschierle:
Calling `createFullResponse()` passes along all present headers. Unfortunately, the cache headers were not properly initialized before, so values provided via `cachedForSeconds()` were not copied to the new response.

Fixes: [OX-9491](https://scireum.myjetbrains.com/youtrack/issue/OX-9491)

### DIFF
--- a/src/main/java/sirius/web/http/Response.java
+++ b/src/main/java/sirius/web/http/Response.java
@@ -807,6 +807,8 @@ public class Response {
             userMessagesCache.cacheUserMessages(webContext);
         }
 
+        setDateAndCacheHeaders(System.currentTimeMillis(), cacheSeconds == null ? 0 : cacheSeconds, isPrivate);
+
         HttpResponse response = createFullResponse(status, true, Unpooled.EMPTY_BUFFER);
         response.headers().set(HttpHeaderNames.LOCATION, url);
         complete(commit(response));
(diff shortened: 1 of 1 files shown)


sabieber (approved):

idlira (approved):

src/main/java/sirius/web/http/Response.java
```diff
@@ -812,11 +799,17 @@ public void redirectToGet(String url) {
...
+        performRedirect(url, HttpResponseStatus.MOVED_PERMANENTLY);
+    }
+
+    private void performRedirect(String url, HttpResponseStatus status) {
         if (cacheSeconds == null || cacheSeconds == 0) {
             userMessagesCache.cacheUserMessages(webContext);
         }
 
-        HttpResponse response = createFullResponse(HttpResponseStatus.MOVED_PERMANENTLY, true, Unpooled.EMPTY_BUFFER);
+        setDateAndCacheHeaders(System.currentTimeMillis(), cacheSeconds == null ? 0 : cacheSeconds, isPrivate);
```
andyHa, line 810: shouldn't we just leave this out for 0/null - otherwise this would be a semantic change to the current code and would require intense testing...

src/main/java/sirius/web/http/Response.java
```diff
@@ -812,11 +799,17 @@ public void redirectToGet(String url) {
      * @param url the URL to redirect to
      */
     public void redirectPermanently(String url) {
+        performRedirect(url, HttpResponseStatus.MOVED_PERMANENTLY);
+    }
+
+    private void performRedirect(String url, HttpResponseStatus status) {
         if (cacheSeconds == null || cacheSeconds == 0) {
             userMessagesCache.cacheUserMessages(webContext);
         }
```
andyHa, line 808: did we challenge this logic with our new setup / approach?

cschierle: Good point, will have an eye on this, but leave it as is for now

PR #1176, opened 2023-02-21
Charts: Library Upgrade & Improved Stacked Line Chart Behaviour

sabieber:
# BREAKING

The [migration guide](https://www.chartjs.org/docs/latest/migration/v4-migration.html) to ChartJS version 4.X mentions some changes in the names of configuration settings so Tycho pages rendering their own charts without using our charts helper may have to receive small adaptations (as seen in the upgrade commit).

# Screenshots

## Inline Charts still work after upgrade

image

## Normal charts still work after upgrade

image

## Tooltips of stacked line charts are improved

image

Fixes: [SIRI-657](https://scireum.myjetbrains.com/youtrack/issue/SIRI-657)


idlira (approved):

idlira (approved):

ymo-sci (approved):

src/main/resources/default/assets/tycho/scripts/charts.js.pasta
```diff
@@ -235,6 +235,23 @@ function _lineChart(selector, labels, datasets, stacked) {
...
+        customConfig.interaction = {
+            mode: 'nearest',
+            axis: 'x',
+            intersect: false
+        }
+        customConfig.plugins.tooltip = {
+            callbacks: {
+                footer: function (tooltipItems) {
+                    return '___i18n("Charts.total"): ' + tooltipItems.reduce(function (acc, item) {
+                        return acc + item.parsed.y;
```
jakobvogel, line 248: Are there cases where displaying such a total may be misleading/superfluous/undesired?

jakobvogel (approved):

src/main/resources/default/assets/tycho/scripts/charts.js.pasta
sabieber: Can't think of one. As it is already calculatable manually by adding all values in the popup or roughly by reading off the highest (stacked) line on the chart

jakobvogel: Hmm, to give an artificial (and stupid) example that is entirely synthetic:

- Percentage of users who prefer cats over dogs
- Percentage of users who prefer green over blue
- Percentage of users who prefer pasta over pizza

It may make sense to display this together, but the sum of the percentages does not make sense, as the signals are independent but happen to use the same scale. In the example, the total may be a pointless value such as "123.45 %"

Of course, this depends massively on the realist use of charts, and whether we really care if a pointless total is displayed

cschierle: Would it be possible to add an optional parameter like `hideTooltipTotal` with default true? Then it could be deactivated if necessary

cschierle (approved):

andyHa:
Has the LICENSE.txt file in the chartjs directory been checked to still match the license of the library?

src/main/resources/default/assets/tycho/scripts/charts.js.pasta
sabieber: It wouldn't even make sense to use the stacking mode in this case imho.
The total in the tooltip is only used when using a stacking line chart

mko-sci (approved):

PR #1207, opened 2023-04-20
Continuity: Allows to jump back multiple history states when entry contains depth counter

sabieber:
This allows for some more complex scenarios where a single history entry also has multiple history sub-entries and popping the entry should undo all of the sub states at once.

Depth counting is currently done manually where necessary.

The depth is always omitted from the deeplink URL as opening a deeplink won't include all previous history states (that were open when creating the deeplink), so jumping back multiple states at once would lead somewhere completely wrong.

Example:
Search that writes a new state each time a filter is changed, rendered within an overlay. Popping (closing) the overlay should not simply undo a single search state change but undo them all at once so the overlay is actually closed.

Fixes: [OX-9814](https://scireum.myjetbrains.com/youtrack/issue/OX-9814)

### DIFF
--- a/src/main/resources/default/assets/common/scripts/continuity.js.pasta
+++ b/src/main/resources/default/assets/common/scripts/continuity.js.pasta
@@ -317,6 +317,8 @@ window.sirius.Continuity = (function () {
             const entryName = entryKey.split(this.entryScopeSeparator)[0];
             if (this.options.entryTypes[entryName] && !this.options.entryTypes[entryName].excludeFromParameter) {
                 const paramStateEntry = Object.assign({}, state[entryKey]);
+                // Omit the entry depth from the parameter state as it only makes sense on the current page with the current history stack.
+                // When opening the URL in a new tab the entry depth would be wrong.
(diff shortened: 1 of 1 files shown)


jakobvogel (approved):

idlira (approved):

fhaScireum (approved):
Alternatively, we could give each entry a "type" like "search-overlay-1" and if you pop one entry with a type we pop every entry before it until an entry comes up that has no or a different type.

sabieber:
> Alternatively, we could give each entry a "type" like "search-overlay-1" and if you pop one entry with a type we pop every entry before it until an entry comes up that has no or a different type.

This sounds like the solution in Tycho history handling which makes the Tycho history handling awkward to debug and hard to understand the behaviour in code.
It would require to remember some kind of marker (Tycho history uses lokal storage for this) that tells the popstate-listener to keep going back.

Its highly likely that we switch the overlay logic after the patch anyways which may result in a more elegant solution for this usage scenario.

PR #1235, opened 2023-06-21
Tycho: Allows ids for tags

cschierle:
Fixes: [SIRI-606](https://scireum.myjetbrains.com/youtrack/issue/SIRI-606)

### DIFF
--- a/src/main/resources/default/taglib/t/heading.html.pasta
+++ b/src/main/resources/default/taglib/t/heading.html.pasta
@@ -1,7 +1,9 @@
 <i:arg type="String" name="labelKey" default=""/>
 <i:arg type="String" name="label" default="@i18n(labelKey)"/>
 <i:arg type="String" name="class" default=""/>
+
 <i:local name="markupLabel" value="@renderToString('body')"/>
+
 <i:if test="isFilled(label)">
     <div class="bb-gray pb-2 mt-4 mb-3 h5 text-black-75 align-items-end d-flex flex-row @class">
         <span class="overflow-hidden">@label<i:render name="label"/></span>
--- a/src/main/resources/default/taglib/t/heading.html.pasta
+++ b/src/main/resources/default/taglib/t/heading.html.pasta
@@ -1,17 +1,17 @@
 <i:arg type="String" name="labelKey" default=""/>
 <i:arg type="String" name="label" default="@i18n(labelKey)"/>
-<i:arg type="String" name="class" default="" />
-<i:local name="markupLabel" value="@renderToString('body')" />
+<i:arg type="String" name="class" default=""/>
+<i:local name="markupLabel" value="@renderToString('body')"/>
 <i:if test="isFilled(label)">
     <div class="bb-gray pb-2 mt-4 mb-3 h5 text-black-75 align-items-end d-flex flex-row @class">
-        <span class="overflow-hidden">@label</span>
-        <span class="ml-auto"><i:render name="actions" /></span>
+        <span class="overflow-hidden">@label<i:render name="label"/></span>
+        <span class="ml-auto"><i:render name="actions"/></span>
     </div>
     <i:else>
         <i:if test="isFilled(markupLabel)">
             <div class="bb-gray pb-2 mt-4 mb-3 h5 text-black-75 align-items-end d-flex flex-row @class">
-                <span class="overflow-hidden"><i:raw>@markupLabel</i:raw></span>
-                <span class="ml-auto"><i:render name="actions" /></span>
+                <span class="overflow-hidden"><i:raw>@markupLabel<i:render name="label"/></i:raw></span>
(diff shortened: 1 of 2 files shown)


src/main/resources/default/taglib/t/heading.html.pasta
```diff
@@ -1,17 +1,17 @@
 <i:arg type="String" name="labelKey" default=""/>
 <i:arg type="String" name="label" default="@i18n(labelKey)"/>
-<i:arg type="String" name="class" default="" />
-<i:local name="markupLabel" value="@renderToString('body')" />
+<i:arg type="String" name="class" default=""/>
+<i:local name="markupLabel" value="@renderToString('body')"/>
 <i:if test="isFilled(label)">
```
jakobvogel, line 5: I personally prefer some blank lines here, but that is – of course – a matter of taste.

```suggestion
<i:arg type="String" name="class" default=""/>

<i:local name="markupLabel" value="@renderToString('body')"/>

<i:if test="isFilled(label)">
```

src/main/resources/default/taglib/t/heading.html.pasta
```diff
@@ -1,17 +1,17 @@
...
     <div class="bb-gray pb-2 mt-4 mb-3 h5 text-black-75 align-items-end d-flex flex-row @class">
-        <span class="overflow-hidden">@label</span>
-        <span class="ml-auto"><i:render name="actions" /></span>
+        <span class="overflow-hidden">@label<i:render name="label"/></span>
```
jakobvogel, line 7: Do we really want to support both? In similar cases, we often seem to check for whether one or the other is filled and only print one.

src/main/resources/default/taglib/t/heading.html.pasta
```diff
@@ -1,17 +1,17 @@
...
+        <span class="ml-auto"><i:render name="actions"/></span>
     </div>
     <i:else>
         <i:if test="isFilled(markupLabel)">
             <div class="bb-gray pb-2 mt-4 mb-3 h5 text-black-75 align-items-end d-flex flex-row @class">
-                <span class="overflow-hidden"><i:raw>@markupLabel</i:raw></span>
-                <span class="ml-auto"><i:render name="actions" /></span>
+                <span class="overflow-hidden"><i:raw>@markupLabel<i:render name="label"/></i:raw></span>
```
jakobvogel, line 13: (s.a.)

src/main/resources/default/taglib/t/tag.html.pasta
```diff
@@ -4,14 +4,15 @@
        default="true"
        description="Determines if the dot has an auto margin at the right"/>
 <i:arg type="String" name="class" default="" description="Defines addition classes to add."/>
+<i:arg type="String" name="id" default="" description="Defines the id of the rendered span."/>
```
jakobvogel, line 7: Let's try not to unnerve @sabieber 😉

```suggestion
<i:arg type="String" name="id" default="" description="Defines the ID of the rendered span."/>
```

src/main/resources/default/taglib/t/heading.html.pasta
cschierle: The use case is to place additional information next to a heading provided by a sirius lib via an extension in a product. This seemed to be the easiest way to achieve this.

Alternatively, I could also try and find another method to append things via JS.

Feel free to comment, which approach I should rather follow.

jakobvogel: That sounds legit. Then I am just wondering whether the block should be named `label-extension` instead.

In the straight-forward use case, I find it a bit strange to do the content as attribute and the extension as nested block in this case. But that is also a matter of taste, of course.

jakobvogel: (Add yes, I would hope that people would use the body version)

sabieber (approved):

idlira (approved):

src/main/resources/default/taglib/t/heading.html.pasta
andyHa: commonly the i:block replaces the parameter. Check card.html.pasta - we ofter renderToString as fallback to the plain parameter..I'd keep it consistent here

andyHa (changes requested):
We should probably stay consistent here...

src/main/resources/default/taglib/t/heading.html.pasta
jakobvogel: For the consistent case, there is already `markupLabel`. @cschierle's use case is truly remarkable 😅

andyHa: Where do we use this at all?

sabieber (approved):

idlira (approved):

jakobvogel (approved):

PR #1237, opened 2023-06-23
Dynamic Template Invocation

jakobvogel:
Enables the `<i:invoke` tag to accept dynamically determined template paths.


jakobvogel:
Not necessary, `<i:dynamicInvoke` already does the job. (Thanks to @fhaScireum for the hint.)

PR #1238, opened 2023-06-23
Improves the t:deleteButton by adding arguments which enables to toggle the button type and label

mko-sci:
Fixes: OX-9962

### DIFF
--- a/src/main/resources/default/taglib/t/deleteButton.html.pasta
+++ b/src/main/resources/default/taglib/t/deleteButton.html.pasta
@@ -2,10 +2,11 @@
 <i:arg type="Page" name="page" default=""/>
 <i:arg name="class" type="String" default="" description="Lists additional CSS classes to apply to the field."/>
 <i:arg name="withLabel" type="boolean" default="false" description="Specifies if the button should have an appropriate label."/>
+<i:arg name="asBtnLink" type="boolean" default="true" description="Specifies if the button should be of type link"/>
 
 <i:pragma name="description" value="Renders a delete button to be used in tables or lists."/>
 
-<a class="btn btn-link btn-danger confirm-link-js @class"
+<a class="btn @if(true) { btn-link } btn-danger confirm-link-js @class"
    title="@i18n('NLS.delete')"
    href="@if(page != null) {@page.linkToCurrentPage(url)} else {@url}">
     <i class="fa-solid fa-trash"></i> @if(withLabel) { @i18n("NLS.delete") }
--- a/src/main/resources/default/taglib/t/deleteButton.html.pasta
+++ b/src/main/resources/default/taglib/t/deleteButton.html.pasta
@@ -2,11 +2,11 @@
 <i:arg name="page" type="Page" default=""/>
 <i:arg name="class" type="String" default="" description="Lists additional CSS classes to apply to the field."/>
 <i:arg name="withLabel" type="boolean" default="false" description="Specifies if the button should have an appropriate label."/>
-<i:arg name="asBtnLink" type="boolean" default="true" description="Specifies if the button should be of type link"/>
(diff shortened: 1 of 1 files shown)


src/main/resources/default/taglib/t/deleteButton.html.pasta
```diff
@@ -1,11 +1,12 @@
 <i:arg type="String" name="url"/>
 <i:arg type="Page" name="page" default=""/>
 <i:arg name="class" type="String" default="" description="Lists additional CSS classes to apply to the field."/>
+<i:arg name="withLabel" type="boolean" default="false" description="Specifies if the button should have an appropriate label."/>
```
idlira, line 4: just `label` ?

idlira (approved):
Screenies would habe been nice 😇

src/main/resources/default/taglib/t/deleteButton.html.pasta
```diff
@@ -1,11 +1,13 @@
-<i:arg type="String" name="url"/>
-<i:arg type="Page" name="page" default=""/>
+<i:arg name="url" type="String" description="Specifies the url which should be called. Is only used when page is left empty."/>
```
jakobvogel, line 1: Have a look at line 11 where the HREF attribute is generated. The URL is used even if there is a `page`.

src/main/resources/default/taglib/t/deleteButton.html.pasta
jakobvogel: I don't like a boolean flag indicating whether to use a certain label. I wonder whether the usual mechanism using an empty string (or an empty) key would be better

src/main/resources/default/taglib/t/deleteButton.html.pasta
```diff
@@ -1,11 +1,13 @@
...
+<i:arg name="page" type="Page" default=""/>
 <i:arg name="class" type="String" default="" description="Lists additional CSS classes to apply to the field."/>
+<i:arg name="withLabel" type="boolean" default="false" description="Specifies if the button should have an appropriate label."/>
+<i:arg name="asBtnLink" type="boolean" default="true" description="Specifies if the button should be of type link."/>
```
jakobvogel, line 5: is this really, really necessary? or is this just for the looks of things? I really don't like such archaic flags 😇

mko-sci: I am not a huge fan either, but I was short on a better solution... 
I wanted a simple toggle to change the type of the button, by also making sure that one would not be able to do too much stuff because the template is called delete button

mko-sci: This is, by the way, the same reason I decided to go with the flag for the label. Its a delete button, I dont want people to screw with the lable text yet.. :D 
But maybe it might be better to support a `labelKey`

jakobvogel (approved):

sabieber (approved):

src/main/resources/default/taglib/t/deleteButton.html.pasta
```diff
@@ -1,11 +1,14 @@
-<i:arg type="String" name="url"/>
-<i:arg type="Page" name="page" default=""/>
+<i:arg name="url" type="String" description="Specifies the url which should be called."/>
+<i:arg name="page" type="Page" default=""/>
 <i:arg name="class" type="String" default="" description="Lists additional CSS classes to apply to the field."/>
+<i:arg name="labelKey" type="String" default="" description="Specifies if the label of the button."/>
```
fhaScireum, line 4:
```suggestion
<i:arg name="labelKey" type="String" default="" description="Specifies the label of the button."/>
```

fhaScireum (approved):

mkeckmkeck (approved):

PR #1245, opened 2023-07-04
Kills the DefaultGlobalContextExtender and marks some methods as deprecated

andyHa:
Requires:
* https://github.com/scireum/sirius-kernel/pull/469


src/main/java/sirius/web/templates/ContentHelper.java
```diff
@@ -61,57 +56,27 @@ public String nl2br(String content) {
...
 
     /**
      * Removes all XML tags from a given content.
      *
      * @param content content to strip XML of
      * @return content without XML tags
+     * @see StringCleanup#replaceXml(String)
      */
     @NoodleSandbox(NoodleSandbox.Accessibility.GRANTED)
     public String stripXML(String content) {
```
jakobvogel, line 78: static as well?

```suggestion
    public static String stripXML(String content) {
```

jakobvogel (approved):

PR #1247, opened 2023-07-05
Continuity: Only calls state handler restore when entry is actually changed

sabieber:
Optimized the 'handleStateChange' function to ensure that the state handler is only called when there is a real change in the state. For this a deep comparison function 'deepEqual' was also added to make sure the changes in the nested properties of objects are correctly detected. This should improve the "optical stability" of the page as state handlers aren't executing code that recreate the currently visible state (e.g. by executing the same search again).

Fixes: [OX-10141](https://scireum.myjetbrains.com/youtrack/issue/OX-10141)

### DIFF
--- a/src/main/resources/default/assets/common/scripts/continuity.js.pasta
+++ b/src/main/resources/default/assets/common/scripts/continuity.js.pasta
@@ -149,7 +149,7 @@ window.sirius.Continuity = (function () {
             if (handler) {
                 const entry = state[entryKey];
                 me.log('[CONTINUITY] [INTERNAL] State handler of type "%s" called for entry: %o', entryKey, entry);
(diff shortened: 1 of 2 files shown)


jakobvogel (approved):

src/main/resources/default/assets/common/scripts/continuity.js.pasta
```diff
@@ -424,5 +430,33 @@ window.sirius.Continuity = (function () {
         }
     }
 
+    function deepEqual(object1, object2) {
```
andyHa, line 433: super optional - we migth make that a public sirius method (?) might be re-usable (??)

andyHa (approved):

PR #1250, opened 2023-07-06
Fixes the sirius.isEmpty method to work properly with js objects

mko-sci:
Fixes: OX-10158

### DIFF
--- a/src/main/resources/default/assets/common/core.js.pasta
+++ b/src/main/resources/default/assets/common/core.js.pasta
@@ -32,7 +32,7 @@ sirius.isEmpty = function (value) {
  * @returns {boolean} true if the given object is empty, false otherwise
  */
 function isEmptyObject (obj) {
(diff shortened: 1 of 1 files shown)


sabieber (approved):

src/main/resources/default/assets/common/core.js.pasta
```diff
@@ -19,9 +19,28 @@ sirius.isEmpty = function (value) {
...
     return !sirius.isDefined(value) || value === '';
 }
 
+/**@
+ * Checks if the given object is empty.
+ * @param obj the object to check
+ * @returns {boolean} true if the given object is empty, false otherwise
+ */
+function isEmptyObject (obj) {
+    for (const prop in obj) {
```
fhaScireum, line 35: Does the `const` in the `for` definition here work in IE11?

PR #1252, opened 2023-07-07
Continuity: Prevents history change handling when back navigation is already queued

sabieber:
Chrome has some weird behaviour where it sometimes completes running all code of the current execution context (everything that wasn't postponed by setTimeout or async calls). This introduced a bug in the continuity history handling where a popStateEntry could actually be reverted automatically as the handleStateChange code ran after back navigation was intended and the back navigation and state change restore would cancel each other. We now use a simple flag to ensure the handleStateChange logic does not get executed when we previously issued a back navigation in popStateEntry.

This is heavily inspired by commit 7c0894290e30e050f276576a5209d7ae52165ce7 which did the same in the tycho history logic.

Fixes: [OX-10192](https://scireum.myjetbrains.com/youtrack/issue/OX-10192)

### DIFF
--- a/src/main/resources/default/assets/common/scripts/continuity.js.pasta
+++ b/src/main/resources/default/assets/common/scripts/continuity.js.pasta
@@ -23,6 +23,9 @@ window.sirius.Continuity = (function () {
         // An entry handler can then be registered by each for `entryType|subScope` to handle the state changes of the specific element.
         this.entryScopeSeparator = '|';
 
+        // This flag is used, to prevent the execution of the history state change logic when we previously issued a back navigation.
+        this.navigationQueued= false;
+
         this.path = window.location.pathname;
         this.searchParams = new URLSearchParams(window.location.search);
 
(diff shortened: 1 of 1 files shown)


src/main/resources/default/assets/common/scripts/continuity.js.pasta
```diff
@@ -23,6 +23,9 @@ window.sirius.Continuity = (function () {
         // An entry handler can then be registered by each for `entryType|subScope` to handle the state changes of the specific element.
         this.entryScopeSeparator = '|';
 
+        // This flag is used, to prevent the execution of the history state change logic when we previously issued a back navigation.
+        this.navigationQueued= false;
```
andyHa, line 27:
```suggestion
        this.navigationQueued = false;
```

andyHa (approved):

idlira (approved):

mkeckmkeck (approved):

jakobvogel (approved):

fhaScireum (approved):

PR #1263, opened 2023-07-27
API Docs: shows method and content type in toc instead of generic icon

cschierle:
Provides a better overview of the service capabilities in the table of contents. Also eliminates name conflicts if documentation exists for the same service with different content types (e.g. json vs xml as in the 2nd screenshot).

Before:
Bildschirmfoto 2023-07-27 um 12 11 19
Bildschirmfoto 2023-07-27 um 12 10 54

After:
Bildschirmfoto 2023-07-27 um 12 17 43

Bildschirmfoto 2023-07-27 um 12 16 53

### DIFF
--- a/src/main/resources/default/templates/system/api.html.pasta
+++ b/src/main/resources/default/templates/system/api.html.pasta
@@ -25,8 +25,9 @@
         <i:block name="sidebar">
             <t:navbox labelKey="template.html.tableOfContent.heading">
                 <i:for type="sirius.web.services.PublicServiceInfo" var="service" items="api.getServices()">
-                    <t:navboxLink url="@apply('#%s', service.getUri())"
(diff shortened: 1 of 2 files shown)


jakobvogel (commented):
Hmm, not sure whether everybody agrees in terms of alignment in the table of contents

cschierle:
> Hmm, not sure whether everybody agrees in terms of alignment in the table of contents

True, I'm also not 100% happy with the alignment. But it also not easy to optimize this with `navboxLink`s.

In my opinion, displaying the tags brings more benefit than perfect alignment though 🤷

sabieber (approved):
Maybe we can optimize it further. I like the suggested change 👍🏻

mkeckmkeck (approved):

PR #1273, opened 2023-08-10
OAuth2 token refreshing util  

mkeckmkeck:
- adds support for our currently needed refresh call format
- parses received tokes from json into object

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -154,6 +154,13 @@
             <artifactId>swagger-annotations</artifactId>
             <version>2.2.6</version>
         </dependency>
+
+        <dependency>
+            <groupId>com.auth0</groupId>
+            <artifactId>java-jwt</artifactId>
+            <version>4.4.0</version>
+        </dependency>
+
     </dependencies>
 
     <dependencyManagement>
--- a/src/main/java/sirius/web/security/oauth/ReceivedTokens.java
+++ b/src/main/java/sirius/web/security/oauth/ReceivedTokens.java
@@ -51,9 +51,9 @@ public record ReceivedTokens(String accessToken, String refreshToken, String typ
                 // No valid JWT, fall back to implementation from OAuth expires_in or the default value
             }
         }
-        long expiresIn = response.path(OAuth.EXPIRES_IN).asLong(0L);
 
         // Check if the 'expires in' field, actually meant for the access token, is better than our default tomorrow.
+        long expiresIn = response.path(OAuth.EXPIRES_IN).asLong(0L);
         if (expiresIn > TWO_DAYS_IN_SECONDS) {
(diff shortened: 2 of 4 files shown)


src/main/java/sirius/web/security/oauth/ReceivedTokens.java
```diff
@@ -0,0 +1,66 @@
...
+        }
+        long expiresIn = response.path(OAuth.EXPIRES_IN).asLong(0L);
+
+        // Check if the 'expires in' field, actually meant for the access token, is better than our default tomorrow.
+        if (expiresIn > TWO_DAYS_IN_SECONDS) {
+            LocalDateTime expiresDate = LocalDateTime.now().plusSeconds(expiresIn);
+            return new ReceivedTokens(accessToken, refreshToken, type, expiresDate);
+        }
+
+        // Use default value tomorrow, we except a refresh token to be valid at least for one more day
```
jakobvogel, line 62:
```suggestion
        // Use default value tomorrow, we expect a refresh token to be valid at least for one more day
```

jakobvogel (approved):

bkositza (approved):

src/main/java/sirius/web/security/oauth/ReceivedTokens.java
```diff
@@ -0,0 +1,66 @@
...
+                LocalDateTime refreshTokenExpiresAt =
+                        JWT.decode(refreshToken).getExpiresAtAsInstant().atZone(ZoneOffset.UTC).toLocalDateTime();
+                return new ReceivedTokens(accessToken, refreshToken, type, refreshTokenExpiresAt);
+            } catch (JWTDecodeException exception) {
+                // No valid JWT, fall back to implementation from OAuth expires_in or the default value
+            }
+        }
+        long expiresIn = response.path(OAuth.EXPIRES_IN).asLong(0L);
+
+        // Check if the 'expires in' field, actually meant for the access token, is better than our default tomorrow.
```
fhaScireum, line 56:
```suggestion

        // Check if the 'expires in' field, actually meant for the access token, is better than our default tomorrow.
        long expiresIn = response.path(OAuth.EXPIRES_IN).asLong(0L);
```
Mimimi code-style: I would put the variable below the comment just above the if-check as it isn't used outside the if-block :)

fhaScireum (approved):

pom.xml
```diff
@@ -154,6 +154,13 @@
             <artifactId>swagger-annotations</artifactId>
             <version>2.2.6</version>
         </dependency>
+
+        <dependency>
+            <groupId>com.auth0</groupId>
+            <artifactId>java-jwt</artifactId>
```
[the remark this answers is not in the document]
mkeckmkeck: MIT license. Works for us. The lib was previously part of S2

sabieber (approved):

PR #1280, opened 2023-08-25
History Handling Bugfix

jakobvogel:
There has been a case where the history was non-null, but the URL within was. We now make sure that both exist before doing history magic.


mko-sci (approved):

bkositza (approved):

fhaScireum (approved):

ymo-sci (approved):

mkeckmkeck (approved):

jakobvogel:
@sabieber FYI — This PR only fixes the symptoms of some history hiccup. In case you want to check on the underlying problem.

PR #1286, opened 2023-09-15
Fixes a Chrome only issue where click-spaming on a button lead to multiple execution of the attached click-handlers although the button had the pending class attached

mko-sci:
The issue was solved by setting the useCapture option to true which controls whether events of this type will be dispatched to the registered listener before being dispatched to any EventTarget beneath it in the DOM tree.

https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener

### DIFF
--- a/src/main/resources/default/assets/tycho/scripts/form.js.pasta
+++ b/src/main/resources/default/assets/tycho/scripts/form.js.pasta
@@ -7,7 +7,7 @@ sirius.ready(function () {
                 return true;
             }
             _node.classList.add('single-click-pending-js');
-        });
+        }, true);
     });
 
     document.querySelectorAll('.submit-link-js').forEach(function (_node) {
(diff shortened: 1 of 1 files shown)


src/main/resources/default/assets/tycho/scripts/form.js.pasta
```diff
@@ -3,6 +3,7 @@ sirius.ready(function () {
         _node.addEventListener('click', function (e) {
             if (_node.classList.contains('single-click-pending-js')) {
                 e.preventDefault();
+                e.stopPropagation();
```
sabieber, line 6: **Optional:** `e` -> `event`

sabieber (approved):

fhaScireum (approved):

jakobvogel (approved):

sabieber (approved):

PR #1289, opened 2023-09-21
Tycho: Adds input addon with copy button

sabieber:
This allows to render an addon button next to a (most probably read-only) which copies the current value of the field into the clipboard on click.
It also updates the tooltip when this action worked.

image

image

**Sample:**
```html
<t:textfield name="id"
             readonly="true"
             value="@token.getId()"
             label="ID"
             helpKey="PortalApiToken.id.help">
    <i:block name="addon">
        <t:copyToClipboardAddon fieldName="id"/>
    </i:block>
</t:textfield>
```


jakobvogel (approved):

mko-sci (approved):

src/main/resources/default/assets/common/core.js.pasta
```diff
@@ -691,3 +691,64 @@ sirius.extractFileExtension = function (filename) {
     }
     return '.' + nameAndExtension.pop();
 }
+
+/**@
```
jmuscireum, line 695: What does this solve that https://github.com/scireum/sirius-web/blob/develop/src/main/resources/default/assets/tycho/scripts/clipboard.js.pasta can't already do (except the scoping)?

mkeckmkeck (approved):

fhaScireum (approved):

src/main/resources/default/assets/common/core.js.pasta
sabieber: Just my opinion:
* Better scoping (along the other helper methods)
* More robust fake element / fallback handling
* Tried and tested code that ran a long time in OX (I can't find a call to the function from clipboard js)
* Marking the element with a class (which is a required use case for OX

jmuscireum (approved):

PR #1299, opened 2023-10-09
Fixes sirius.isEmpty not working for dom elements

mko-sci:
sirius.isEmpty returned always ture if a dom element was provided. This clearly was a bug and not intended

Fixes: SIRI-886

### DIFF
--- a/src/main/resources/default/assets/common/core.js.pasta
+++ b/src/main/resources/default/assets/common/core.js.pasta
@@ -8,7 +8,7 @@ sirius.isDefined = function (value) {
     return typeof value !== 'undefined' && value !== null && value !== undefined;
 }
 
-/**
+/**@
  * Checks if a given DOM node is empty. An empty node has no child nodes
  * and its text content is empty or contains only whitespace characters.
  *
--- a/src/main/resources/default/assets/common/core.js.pasta
+++ b/src/main/resources/default/assets/common/core.js.pasta
@@ -32,6 +32,10 @@ sirius.isEmptyNode = function (node) {
 
 /**@
  * Checks if the given value is empty or null.
+ * <p>
+ * Note: Regarding dom elements the method only checks if the given value is a dom element and therefore is defined.
+ * It does not specify if the provided dom element is empty in a literal sense (has no children) or not.
+ * Please use the method sirius.isEmptyNode for this purpose.
  *
  * @param value the value to check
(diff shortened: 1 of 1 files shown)


src/main/resources/default/assets/common/core.js.pasta
```diff
@@ -20,6 +20,11 @@ sirius.isEmpty = function (value) {
     }
 
     if (sirius.isDefined(value) && typeof value === 'object') {
+        // Check if it's a DOM object (has a nodeType property)
+        if (value.nodeType !== undefined) {
+            return false;
```
sabieber, line 25: A meaningful empty check for DOM elements could be implemented here. Otherwise I would actually prefer to throw an error as always returning `false` isn't correct either.

A meaningful logic could be to check whether the list of child Nodes and the textContent are empty

src/main/resources/default/assets/common/core.js.pasta
```diff
@@ -8,8 +8,34 @@ sirius.isDefined = function (value) {
     return typeof value !== 'undefined' && value !== null && value !== undefined;
 }
 
+/**
```
sabieber, line 11:
```suggestion
/**@
```
Otherwise Tagliatelle breaks as `@param` is no Variable in the Scope

src/main/resources/default/assets/common/core.js.pasta
```diff
@@ -8,8 +8,34 @@ sirius.isDefined = function (value) {
...
+        return textContent === '';
+    }
+
+    return false;
+}
+
 /**@
  * Checks if the given value is empty or null.
+ * <p>
+ * Note: Regarding dom elements the method only checks if the given value is a dom element and therefore is defined.
```
sabieber, line 36:
```suggestion
 * Note: Regarding DOM elements the method only checks if the given value is a DOM element and therefore is defined.
```

src/main/resources/default/assets/common/core.js.pasta
```diff
@@ -8,8 +8,34 @@ sirius.isDefined = function (value) {
...
+ * It does not specify if the provided dom element is empty in a literal sense (has no children) or not.
```
sabieber, line 37:
```suggestion
 * It does not specify if the provided DOM element is empty in a literal sense (has no children) or not.
```

sabieber (approved):

jmuscireum (approved):

PR #1314, opened 2023-11-02
Supports If-Modified-Since HTTP header delivering assets

idlira:
If the asset has been modified before the [If-Modified-Since](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Modified-Since) requested date, we return a 304 NOT_MODIFIED return code with am empty body.

This logic was implemented for tagliatelle templates, but not for static assets or compiled CSS files, although a bug has been fixed here and the portion of code never really worked. See 1st. commit for details.

Fixes: [SIRI-729](https://scireum.myjetbrains.com/youtrack/issue/SIRI-729)

### DIFF
--- a/src/main/java/sirius/web/dispatch/AssetsDispatcher.java
+++ b/src/main/java/sirius/web/dispatch/AssetsDispatcher.java
@@ -222,7 +222,9 @@ public class AssetsDispatcher implements WebDispatcher {
             }
         }
 
-        response.named(uri.substring(uri.lastIndexOf('/') + 1)).file(file);
+        if (!handleUnmodified(file, response)) {
+            response.named(uri.substring(uri.lastIndexOf('/') + 1)).file(file);
+        }
         return DispatchDecision.DONE;
     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/web/dispatch/AssetsDispatcher.java
```diff
@@ -105,18 +105,18 @@ public DispatchDecision dispatch(WebContext ctx) throws Exception {
     private DispatchDecision tryStaticResource(WebContext ctx, String uri, Response response)
             throws URISyntaxException, IOException {
         Optional<Resource> res = resources.resolve(uri).filter(resource -> !resource.getPath().endsWith(PASTA_SUFFIX));
```
sabieber, line 108: **Optional:** `res` -> `resource`, `ctx` -> `webContext`

src/main/java/sirius/web/dispatch/AssetsDispatcher.java
```diff
@@ -179,6 +186,14 @@ private boolean handleUnmodified(Template template, Response response) {
         return response.handleIfModifiedSince(template.getCompilationTimestamp());
     }
 
+    private boolean handleUnmodified(URLConnection urlConnection, Response response) {
+        return response.handleIfModifiedSince(urlConnection.getLastModified());
+    }
+
+    private boolean handleUnmodified(File file, Response response) {
+        return response.handleIfModifiedSince(file.lastModified());
+    }
```
sabieber, line 195: **Optional:** In my opinion these don't really have a benefit as they just shadow a call and use the same number of arguments, thus they don't simplify the code

sabieber (approved):

jakobvogel (approved):

bkositza (approved):

mko-sci (approved):

sabieber (approved):

PR #1316, opened 2023-11-02
Slider: Multi-Column View (OX-10502)

jakobvogel:
Extends the _Slider_ to support multiple columns, i.e. to slide piecewise.

### DIFF
--- a/src/main/resources/default/assets/common/scripts/slider.js.pasta
+++ b/src/main/resources/default/assets/common/scripts/slider.js.pasta
@@ -32,6 +32,18 @@
         options.dots = !!options.dots;
         options.interrupt = !!options.interrupt;
 
+        if (options.columns <= 0) {
+            console.warn('Invalid number of columns, defaulting to 1.');
+            options.columns = 1;
+        }
+        if (options.columnsSm <= 0) {
+            console.warn('Invalid number of columns on small screens, defaulting to 1.');
+            options.columnsSm = 1;
+        }
+        if (options.columnsXs <= 0) {
+            console.warn('Invalid number of columns on tiny screens, defaulting to 1.');
+            options.columnsXs = 1;
+        }
         if (options.animation <= 0) {
             console.warn('Invalid animation timing, defaulting to ' + (DEFAULT_ANIMATION / 1000) + ' seconds.');
             options.animation = DEFAULT_ANIMATION;
@@ -39,6 +51,15 @@
         if (options.transition && options.transition !== TRANSITION_BLEND && options.transition !== TRANSITION_SLIDE) {
             console.warn('Invalid transition identifier, disabling transition.');
             options.transition = undefined;
+        }
+
+        if (options.columnsSm > options.columns) {
+            console.warn('Number of columns on small screen must not be larger than the number of columns on normal screens. Resetting to ' + options.columns + '.');
+            options.columnsSm = options.columns;
+        }
+        if (options.columnsXs > options.columnsSm) {
+            console.warn('Number of columns on tiny screen must not be larger than the number of columns on small screens. Resetting to ' + options.columnsSm + '.');
+            options.columnsXs = options.columnsSm;
         }
 
         // we remember the number of columns in a local variable, as the value may change according to the screen size,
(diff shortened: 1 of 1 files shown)


mkeckmkeck (approved):

src/main/resources/default/assets/common/scripts/slider.js.pasta
```diff
@@ -248,6 +258,16 @@
         };
 
         const fixView = function () {
+            if (window.innerWidth) {
+                if (window.innerWidth < 992) {
+                    columns = options.columnsXs;
+                } else if (window.innerWidth < 1200) {
```
sabieber, line 336: **Question:** Are these magic numbers or are they trying to mirror the bootstrap breakpoints? If its the later they dont match the naming of the sirius or bootstrap breakpoints at all. Can we at least re-use one of the naming schemes to not have a third? 😅

https://getbootstrap.com/docs/5.0/layout/breakpoints/
https://github.com/scireum/sirius-web/blob/e7b0112d7c90ff719f381afcefef1af3c3364320/src/main/resources/assets/design-system/design-system.scss#L1

sabieber (changes requested):

src/main/resources/default/assets/common/scripts/slider.js.pasta
jakobvogel: That is the scheme we use within Stories. Of course, we can also use another one.

src/main/resources/default/assets/common/scripts/slider.js.pasta
```diff
@@ -9,6 +9,9 @@
      * @param target  css selector selecting the outer html node containing all slides as child nodes
```
mko-sci, line 9: please add the type of each `@param` to the documentation.

src/main/resources/default/assets/common/scripts/slider.js.pasta
```diff
@@ -9,6 +9,9 @@
...
      * @param options optional object specifying the behaviour of the slider (see constants for possible values):
      *                 - animation:  Time in ms between sliding (can be undefined for no animation)
```
mko-sci, line 11: if `options` is an object, and therefore `animation` is a member it could be documented as:
`@param {number} options.animation Time in ms between sliding (....)` also if `options.animation` is optional it could be wrapped in brackets `[options.animation]`. If there is a default value it could be specified with an equals such as: `[options.animation=100]`

src/main/resources/default/assets/common/scripts/slider.js.pasta
```diff
@@ -234,17 +332,40 @@
         };
 
         const fixView = function () {
+            if (window.innerWidth) {
+                if (window.innerWidth < 992) {
```
mko-sci, line 336: why 992? Exp?

mko-sci (approved):

src/main/resources/default/assets/common/scripts/slider.js.pasta
```diff
@@ -6,16 +6,30 @@
...
+     * @param {number} [options.columnsSm=columnsMd]
+     *        number of columns to show at a time on small-sized screens, by default equal to `columnsMd`; must not be
+     *        larger than `columnsMd`
+     * @param {boolean} [options.interrupt=false]
+     *        if **true,** automatic sliding will be suspended while the mouse pointer hovers over the slider
+     * @param {boolean} [options.arrows=false]
+     *        if **true,** clickable arrows for the previous and next views are rendered on the left and on the right of
+     *        the content, respectively
+     * @param {boolean} [options.dots]
+     *        if **true,** clickable dots showing the currently selected view are rendered below the content
```
mko-sci, line 32: Was the place of the comma intentionally?

mko-sci: Or should it be **true**, ...

jakobvogel: I usually put the comma into the emphasis, but we can also change this

sabieber (approved):

PR #1318, opened 2023-11-03
JS: Adds helper for checking if a DOM node is probably visible

sabieber:
This is not a definitive way (as there are many more ways to manipulate a node to not show up, like size, opacity, ...) but its a useful way to determine whether a node is shown or hidden via our most used measures:

- manual display none
- sci-d-none

Fixes: [OX-10422](https://scireum.myjetbrains.com/youtrack/issue/OX-10422)

### DIFF
--- a/src/main/resources/default/assets/common/core.js.pasta
+++ b/src/main/resources/default/assets/common/core.js.pasta
@@ -31,6 +31,22 @@ sirius.isEmptyNode = function (node) {
 }
 
 /**@
+ * Checks if a given DOM node is visible.
+ * <p>
+ * Note: This is not a perfect way to check for visibility, but it's a good approximation in controlled circumstances.
+ *
+ * @param {Node} node the DOM node to check for visibility
+ * @returns {boolean} true if the node is most probably visible
+ */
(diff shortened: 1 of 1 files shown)


mkeckmkeck (approved):

src/main/resources/default/assets/common/core.js.pasta
```diff
@@ -30,6 +30,22 @@ sirius.isEmptyNode = function (node) {
     return false;
 }
 
+/**@
+ * Checks if a given DOM node is visible.
```
jakobvogel, line 34: Hmm, I would like to have the documentation extended a bit, to describe what visibility is in this context. Apparently, it does not mean "visible in the viewport“ here, i.e. the method checks only for whether a node _could_ be visible, not whether it is currently shown to the user.

jakobvogel (approved):

src/main/resources/default/assets/common/core.js.pasta
sabieber: Better?

jakobvogel (approved):

ymo-sci (approved):

PR #1320, opened 2023-11-06
Revamps asset caching

idlira:
# Makes the default client cache duration configurable
Key: `http.response.defaultClientCacheTTL`
This is basically what lands in Cache-Control's max-age HTTP header.
**BREAKING**: The `Response.HTTP_CACHE` is gone. Use `Response.obtainClientDurationInSeconds()` instead.

# Increases the cache of constant assets to 6h
Configurable with the key: `http.response.defaultStaticAssetTTL`

# Introduces the `X-Custom-TTL` HTTP response header
This header can be used by reverse proxies, e.g [Varnish](varnish-cache.org), to set a different TTL (time-to-leave) value than the one provided with the standard HTTP headers, such as `Cache-Control`.

# Sets the X-Custom-TTL HTTP response header of constant assets to 30d
Configurable with the key: `http.response.defaultCustomProxyTTL`

Fixes: [SIRI-728](https://scireum.myjetbrains.com/youtrack/issue/SIRI-728)

### DIFF
--- a/src/main/java/sirius/web/dispatch/AssetsDispatcher.java
+++ b/src/main/java/sirius/web/dispatch/AssetsDispatcher.java
@@ -71,7 +71,8 @@ public class AssetsDispatcher implements WebDispatcher {
     private String cacheDir;
 
     @ConfigValue("http.response.defaultStaticAssetTTL")
-    private Duration defaultStaticAssetTTL;
+    private static Duration defaultStaticAssetTTL;
+
     private File cacheDirFile;
 
     @Part
@@ -143,7 +144,7 @@ public class AssetsDispatcher implements WebDispatcher {
             return 0;
         }
 
-        return constantAsset ? (int) defaultStaticAssetTTL.getSeconds() : Response.fetchDefaultClientTTL();
+        return constantAsset ? (int) defaultStaticAssetTTL.getSeconds() : Response.obtainClientDurationInSeconds();
     }
 
     private String computeEffectiveCustomProxyTTL(String uri, boolean constantAsset) {
--- a/src/main/java/sirius/web/dispatch/AssetsDispatcher.java
+++ b/src/main/java/sirius/web/dispatch/AssetsDispatcher.java
@@ -118,6 +118,7 @@ public class AssetsDispatcher implements WebDispatcher {
 
         webContext.enableTiming(ASSETS_PREFIX);
         response.cachedForSeconds(computeEffectiveCacheTTL(webContext.getRequestedURI(), true));
+        response.customProxyCached(computeEffectiveCustomProxyTTL(webContext.getRequestedURI(), true));
         URL url = optionalResource.get().getUrl();
         if ("file".equals(url.getProtocol())) {
             File file = new File(url.toURI());
@@ -145,6 +146,13 @@ public class AssetsDispatcher implements WebDispatcher {
         return constantAsset ? (int) defaultStaticAssetTTL.getSeconds() : Response.fetchDefaultClientTTL();
     }
 
+    private String computeEffectiveCustomProxyTTL(String uri, boolean constantAsset) {
+        if (uri.startsWith("/assets/no-cache/") || !constantAsset) {
+            return "0s";
+        }
+        return Response.fetchDefaultCustomProxyTTL();
+    }
+
     private String computeEffectiveURI(WebContext webContext) {
         String uri = webContext.getRequestedURI();
         if (uri.startsWith("/assets/dynamic/")) {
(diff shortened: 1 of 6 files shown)


src/test/java/sirius/web/dispatch/AssetsDispatcherSpec.groovy
```diff
@@ -40,7 +51,8 @@ class AssetsDispatcherSpec extends BaseSpecification {
         StringWriter writer = new StringWriter()
         gen.generate(new Output(writer, true))
         then:
-        writer.toString() == "test { background: url(data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAADIAAAAyCAIAAACRXR/mAAAACXBIWXMAAAsTAAALEwEAmpwYAAAAB3RJTUUH4QsODw4S4KU/XgAAAB1pVFh0Q29tbWVudAAAAAAAQ3JlYXRlZCB3aXRoIEdJTVBkLmUHAAAAdElEQVRYw+3Q3Q1AMAAA4ar+qHhghq7SBQ3CPsQO5a1JS0wh+nA3wZdr5hBEfUlRZbBgwYIFCxYsWLBgwYIFCxYsWLBgwYIFCxYsWLBgfZSqDRRtWv1eC2vx2zFGV7S5pX2U+n3M2SWX1ZCNv6a+aHNL/bQvbxUXkThEKBQAAAAASUVORK5CYII=); }\n "
+        writer
+                .toString() == "test { background: url(data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAADIAAAAyCAIAAACRXR/mAAAACXBIWXMAAAsTAAALEwEAmpwYAAAAB3RJTUUH4QsODw4S4KU/XgAAAB1pVFh0Q29tbWVudAAAAAAAQ3JlYXRlZCB3aXRoIEdJTVBkLmUHAAAAdElEQVRYw+3Q3Q1AMAAA4ar+qHhghq7SBQ3CPsQO5a1JS0wh+nA3wZdr5hBEfUlRZbBgwYIFCxYsWLBgwYIFCxYsWLBgwYIFCxYsWLBgfZSqDRRtWv1eC2vx2zFGV7S5pX2U+n3M2SWX1ZCNv6a+aHNL/bQvbxUXkThEKBQAAAAASUVORK5CYII=); }\n "
```
jakobvogel, line 55: Intentional?

jakobvogel (approved):

src/test/java/sirius/web/dispatch/AssetsDispatcherSpec.groovy
idlira: Formatter's BS. But the TC runs, would not bother as we will convert this to Kotlin anyways

src/main/java/sirius/web/dispatch/AssetsDispatcher.java
```diff
@@ -128,25 +134,46 @@ private DispatchDecision tryStaticResource(WebContext webContext, String uri, Re
...
+            return;
+        }
+
+        response.cachedForSeconds(Response.obtainClientDurationInSeconds());
+    }
+
+    private String computeEffectiveURI(WebContext webContext) {
         String uri = webContext.getRequestedURI();
         if (uri.startsWith("/assets/dynamic/")) {
             uri = uri.substring(16);
```
mko-sci, line 160: could we introduce a constant for the 16 which also documents what its for?
Same applies to 17

mko-sci (approved):

mkeckmkeck (approved):

fhaScireum (approved):

src/main/java/sirius/web/dispatch/AssetsDispatcher.java
idlira: Well, this is the length of the string just above. Its removing the prefix so /assets/dynamic/foo -> /foo
I think is straight forward, otherwise must put the string in a constant and get the .length of it

src/main/java/sirius/web/http/Response.java
```diff
@@ -678,6 +702,27 @@ public Response infinitelyCached() {
...
+
+    /**
+     * Sets the default value for the {@link #CUSTOM_PROXY_CACHE_TTL_HEADER} header.
+     *
+     * @return <tt>this</tt> to fluently create the response
+     */
+    public Response withDefaultCustomProxyTTL() {
+        this.customProxyTTL = defaultCustomProxyTTL;
+        return this;
+    }
```
sabieber, line 724: **Optional:** Correct Camel-Casing would be `withCustomProxyTtl`

sabieber (approved):

src/main/java/sirius/web/http/Response.java
idlira: It is, but looks *ugly* 🤓

PR #1323, opened 2023-11-09
Adds multiple sci-icons 

mko-sci:
Adds the following css icon classes:

Bildschirmfoto 2023-11-09 um 13 55 46

Fixes: OX-10532

### DIFF
--- a/src/main/resources/assets/design-system/icons.scss
+++ b/src/main/resources/assets/design-system/icons.scss
@@ -100,3 +100,7 @@
 .sci-icon-product-cart-gray {
   background-image: url(svgResource('/assets/images/icons/cart.svg', #212529));
 }
+
+.sci-icon-download-gray {
+  background-image: url(svgResource('/assets/images/icons/download.svg', #212529));
+}
--- a/src/main/resources/assets/design-system/icons.scss
+++ b/src/main/resources/assets/design-system/icons.scss
@@ -108,3 +108,7 @@
 .sci-icon-catalog-gray {
   background-image: url(svgResource('/assets/images/icons/catalog.svg', #212529));
 }
+
+.sci-icon-magnifier-gray {
+    background-image: url(svgResource('/assets/images/icons/magnifier.svg', #212529));
+}
--- a/src/main/resources/assets/design-system/icons.scss
+++ b/src/main/resources/assets/design-system/icons.scss
@@ -104,3 +104,7 @@
 .sci-icon-download-gray {
   background-image: url(svgResource('/assets/images/icons/download.svg', #212529));
 }
+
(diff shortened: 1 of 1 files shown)


src/main/resources/assets/design-system/icons.scss
```diff
@@ -96,3 +96,19 @@
...
+
+.sci-icon-product-cart-gray {
+  background-image: url(svgResource('/assets/images/icons/cart.svg', #212529));
+}
+
+.sci-icon-download-gray {
+  background-image: url(svgResource('/assets/images/icons/download.svg', #212529));
+}
+
+.sci-icon-catalog-gray {
```
idlira, line 108: Maybe `document`, `book` or `open-book` better shows what the icon really is.

idlira (approved):

src/main/resources/assets/design-system/icons.scss
idlira: Just out of curiosity: shall colors be part of the icons or should it be added via extra classes?

jakobvogel (approved):

src/main/resources/assets/design-system/icons.scss
```diff
@@ -96,3 +96,7 @@
 .sci-icon-product-cart-blue {
   background-image: url(svgResource('/assets/images/icons/cart.svg', #3465CC));
 }
+
+.sci-icon-product-cart-gray {
```
sabieber, line 100: Would use the same naming as the sirius-colors or otherwise it gets confusing as sirius-gray = #ACACAC. So maybe call this -muted

src/main/resources/assets/design-system/icons.scss
sabieber: But we would have to rename the svg itself in that case, which could be breaking

sabieber (changes requested):

src/main/resources/assets/design-system/icons.scss
mko-sci: The color comes from `sci-text-soft`.
The class `sci-muted-link` does not modify the color:
```
.sci-muted-link {
  text-decoration: none;
  cursor: pointer;

  &:hover {
    text-decoration: underline;
  }
}
```

mko-sci: Should I replaces `-gray` with `-soft` then?

sabieber: Sorry yes, thats what I meant. Or `-text-soft`

idlira (approved):

sabieber (approved):

jakobvogel (approved):

PR #1324, opened 2023-11-09
Adds a neat little helper which returns a value as a List of Strings

mko-sci:
...we have multiple parameter where multiple values can be provided by separating them either by a comma or pipe. This little helper allows to directly receive a List of Strings for a certain key.

Fixes: OX-10532

### DIFF
--- a/src/main/java/sirius/web/http/WebContext.java
+++ b/src/main/java/sirius/web/http/WebContext.java
@@ -591,6 +591,18 @@ public class WebContext implements SubContext {
         return Value.EMPTY;
     }
 
+    /**
+     * Returns a value or parameter supplied by the request as List of Strings.
+     * <p>
+     * The individual values must be separated by a comma or a pipe character.
+     *
+     * @param key the key used to look for the value
+     * @return a List of Strings representing the provided data.
+     */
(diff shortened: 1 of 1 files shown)


src/main/java/sirius/web/http/WebContext.java
```diff
@@ -591,6 +591,18 @@ public Value get(String key) {
...
+     * Returns a value or parameter supplied by the request as List of Strings.
+     * <p>
+     * The individual values must be separated by a comma or a pipe character.
+     *
+     * @param key the key used to look for the value
+     * @return a List of Strings representing the provided data.
+     */
+    public List<String> getAsStringList(String key) {
+        return Arrays.stream(get(key).asString().split("[,|]")).filter(Strings::isFilled).toList();
+    }
```
jakobvogel, line 604: Wouldn't something like that fit better into `sirius.kernel.commons.Value`? And maybe use a CSV parser if escaping is supposed to be supported.

fhaScireum: This implementation seems oddly specific. A WebContext already supports a parameter being set multiple times and all set values can be fetched via `WebContext#getParameters`.

PR #1330, opened 2023-11-27
Adds the possibility to create dependencies between two SingleSelects

bkositza:
If the selection options of a SingleSelect are dependent on the selection of another SingleSelect, this can now be specified via a dependencyField and the dependent fields can be determined via autocomplete.

image

Whenever we select a classification system, only the classes that belong to it will be specified as an option.

### DIFF
--- a/src/main/resources/default/taglib/t/singleSelect.html.pasta
+++ b/src/main/resources/default/taglib/t/singleSelect.html.pasta
@@ -12,6 +12,7 @@
 <i:arg name="optional" type="boolean" default="false"/>
 <i:arg name="allowCustomEntries" type="boolean" default="false"/>
 <i:arg name="suggestionUri" type="String" default=""/>
+<i:arg name="dependencyField" type="String" default=""/>
 
 <i:pragma name="description">
     Renders a select field for a single value using the given parameters. Note that additional flags can be set via CSS
@@ -51,6 +52,9 @@
 </div>
 <script>
     sirius.ready(function () {
+        const _dependencyField = document.getElementById('@dependencyField');
+        const _dependencyFieldSelect = document.getElementById('@dependencyField' + '-select');
+
         const autocomplete = new TokenAutocomplete(
             Object.assign({
                 selector: '#@id',
@@ -63,8 +67,16 @@
                 noMatchesText: '@i18n("template.html.selects.noMatches")',
                 noMatchesCustomEntriesDescription: '@i18n("template.html.selects.useCustomEntry")',
                 placeholderText: '@placeholder',
-                suggestionsUri: '@suggestionUri'
+                suggestionsUri: '@suggestionUri',
+                suggestionsUriBuilder: function (query) {
+                    return this.suggestionsUri + (_dependencyFieldSelect ? "/" + _dependencyFieldSelect.value : '') + '?query=' + query;
+                }
             }, <i:render name="autocompleteOptions"/>));
+        if (_dependencyField) {
+            _dependencyField.addEventListener("tokens-changed", function() {
+                autocomplete.autocomplete.loadSuggestions();
+            });
+        }
 
         <i:render name="script"/>
     });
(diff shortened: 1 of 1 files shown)


src/main/resources/default/taglib/t/singleSelect.html.pasta
```diff
@@ -12,6 +12,7 @@
 <i:arg name="optional" type="boolean" default="false"/>
 <i:arg name="allowCustomEntries" type="boolean" default="false"/>
 <i:arg name="suggestionUri" type="String" default=""/>
+<i:arg name="dependencyField" type="String" default=""/>
```
sabieber, line 15: **Optional:** `dependencyFieldId` would be more descriptive

src/main/resources/default/taglib/t/singleSelect.html.pasta
```diff
@@ -51,6 +52,9 @@
 </div>
 <script>
     sirius.ready(function () {
+        const _dependencyField = document.getElementById('@dependencyField');
+        const _dependencyFieldSelect = document.getElementById('@dependencyField' + '-select');
```
sabieber, line 56: **Required:** If the other field is also a TokenAutocomplete the [val](https://github.com/scireum/token-autocomplete/blob/master/lib/token-autocomplete.ts#L288C5-L288C5) function should be used instead of relying on the internal structure of the component. This could change in the future and break this functionality.

src/main/resources/default/taglib/t/singleSelect.html.pasta
```diff
@@ -63,8 +67,16 @@
...
-                suggestionsUri: '@suggestionUri'
+                suggestionsUri: '@suggestionUri',
+                suggestionsUriBuilder: function (query) {
+                    return this.suggestionsUri + (_dependencyFieldSelect ? "/" + _dependencyFieldSelect.value : '') + '?query=' + query;
+                }
             }, <i:render name="autocompleteOptions"/>));
+        if (_dependencyField) {
+            _dependencyField.addEventListener("tokens-changed", function() {
+                autocomplete.autocomplete.loadSuggestions();
+            });
```
sabieber, line 78: **Question:** Why is this necessary? Suggestions should be loaded when the input field gains focus and it should not have focus when another input field changes (as in the user selects one of the suggestions of another field).

src/main/resources/default/taglib/t/singleSelect.html.pasta
```diff
@@ -63,8 +67,16 @@
                 noMatchesText: '@i18n("template.html.selects.noMatches")',
                 noMatchesCustomEntriesDescription: '@i18n("template.html.selects.useCustomEntry")',
                 placeholderText: '@placeholder',
-                suggestionsUri: '@suggestionUri'
+                suggestionsUri: '@suggestionUri',
+                suggestionsUriBuilder: function (query) {
+                    return this.suggestionsUri + (_dependencyFieldSelect ? "/" + _dependencyFieldSelect.value : '') + '?query=' + query;
```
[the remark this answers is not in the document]
idlira: I vote for parameter.

mko-sci (approved):

src/main/resources/default/taglib/t/singleSelect.html.pasta
bkositza: It was just a thought that the dropdown opens as soon as a value is selected in the dependent field. A kind of hint for the user, but of course you're right, we can probably leave that out.

mkeckmkeck (approved):

idlira (changes requested):

src/main/resources/default/taglib/t/singleSelect.html.pasta
```diff
@@ -69,14 +68,13 @@
                 placeholderText: '@placeholder',
                 suggestionsUri: '@suggestionUri',
                 suggestionsUriBuilder: function (query) {
-                    return this.suggestionsUri + (_dependencyFieldSelect ? "/" + _dependencyFieldSelect.value : '') + '?query=' + query;
+                    let baseUri = this.suggestionsUri + '?query=' + query;
+                    if (_dependencyField) {
+                        return baseUri + '&dependentValue=' + _dependencyField.tokenAutocomplete.val().join();
```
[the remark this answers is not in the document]
bkositza: Good point, i encoded both now, the dependentValue and the query

idlira (approved):

sabieber (approved):

PR #1337, opened 2023-12-05
JUNIT: Migrate iprange test to kotlin

MOOOOOSER:


mko-sci (approved):

mkeckmkeck (approved):
I would prefer migrating spock/groovy tests first. There isn't much change and improvement in the migration here.

PR #1354, opened 2024-01-08
JUnit: Migrate CSRFTokenTest to Kotlin

scireum-mbo:
Fixes: OX-10357


src/test/kotlin/sirius/web/http/CSRFTokenTest.kt
```diff
@@ -0,0 +1,181 @@
...
+import java.net.URL
+
+import java.nio.charset.StandardCharsets
+import kotlin.test.assertEquals
+
+@ExtendWith(SiriusExtension::class)
+class CSRFTokenTest {
+
+    @Test
+    fun `safePOST() works correctly if token is missing via GET`() {
```
mkeckmkeck, line 27: Maybe add more meaningful description than "works correctly" :-)
E.g.: "safePOST() fails if token is missing via GET" or "safePOST() fails on GET with missing token"

mkeckmkeck (approved):

src/test/kotlin/sirius/web/http/CSRFTokenTest.kt
```diff
@@ -27,7 +28,7 @@ class CSRFTokenTest {
 
         val result = TestRequest.GET("/test/fake-delete-data").execute()
 
-        result.getStatus() == HttpResponseStatus.INTERNAL_SERVER_ERROR
+        assertEquals(result.getStatus(), HttpResponseStatus.INTERNAL_SERVER_ERROR)
```
sabieber, line 31: **Optional:** Change the order of the parameters. The first parameter should contain the expected value and the second one the actual value.
This improves the clarity of the error message when a test fails.

sabieber (approved):

src/test/kotlin/sirius/web/http/CSRFTokenTest.kt
```diff
@@ -0,0 +1,181 @@
...
+import java.net.URL
+
+import java.nio.charset.StandardCharsets
+import kotlin.test.assertEquals
+
+@ExtendWith(SiriusExtension::class)
+class CSRFTokenTest {
+
+    @Test
+    fun `safePOST() passes test if token is missing via GET`() {
```
mkeckmkeck, line 27: This is misleading. Actually safePOST() does not pass. But every test should pass, so providing that information is not relevant. 
The whole setup in here isn't to elegant, e.g. we do blacklist tests on web api level and test description mentions internal method that isn't visible in test.
The test endpoint expecting valid csrf token should also not be called "fake-delete-data" but maybe "csrf-protected". But works for me like this. I don't won't to slow down kotlin migration to much

idlira (approved):

mkeckmkeck (approved):

PR #1373, opened 2024-01-26
Fixes the incorrect calculation of the dimensions

bkositza:
It was noticeable with particularly high images with a narrow width.

Some of the images ran across several pages of the PDF.

Some images also overlapped areas of the PDF that should not be overlapped:

Before:
image
image

After:
image
image


sabieber (approved):
After comparison image would be nice :)

idlira (approved):

fhaScireum (approved):

mkeckmkeck (approved):

PR #1375, opened 2024-01-29
Fixes scaling of images

bkositza:
For very large images that would fall out of the box in both height and width, both must be scaled down.

### DIFF
--- a/src/main/java/sirius/web/templates/pdf/handlers/PdfReplaceHandler.java
+++ b/src/main/java/sirius/web/templates/pdf/handlers/PdfReplaceHandler.java
@@ -129,7 +129,7 @@ public abstract class PdfReplaceHandler implements Priorized {
             imageWidth = cssWidth;
         }
 
-        // Depending on the image aspect ration, the height might still be higher than the defined limit, so
+        // Depending on the image aspect ratio, the height might still be larger than the defined limit, so
         // we scale down further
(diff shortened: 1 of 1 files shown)


sabieber (approved):

src/main/java/sirius/web/templates/pdf/handlers/PdfReplaceHandler.java
```diff
@@ -120,10 +120,22 @@ private Tuple<Integer, Integer> computeResizeBox(int cssWidth, int cssHeight, FS
...
+        int imageHeight = fsImage.getHeight();
+
+        // First, check if we need to scale down the width
+        if (imageWidth > cssWidth) {
+            imageHeight = cssWidth * imageHeight / imageWidth;
+            imageWidth = cssWidth;
         }
-        return Tuple.create((cssHeight * fsImage.getWidth()) / fsImage.getHeight(), cssHeight);
+
+        // Sometimes we also need to scale down the height - even if the width is already scaled down
```
idlira, line 132:
```suggestion
        // Depending on the image aspect ration, the height might still be higher than the defined limit, so
        // we scale down further
```

idlira (approved):

jakobvogel (approved):

fhaScireum (approved):

PR #1376, opened 2024-01-31
Checks if we need to downscale the height after upscaling the width

bkositza:
Depending on how the maximum permitted sizes for an image are specified, it may be that both the width and height are too small and therefore need to be scaled up. If we scale up the width first, it may well be that the height is too high for the specified maximum height.

Example:
Maximum width: 100px
Maximum height: 20px

Image:
50px wide and 15px high

After we have scaled up the width, the image would be :
100px wide and 30px high

The height of the image is now too high and we need to scale it down once.

### DIFF
--- a/src/main/java/sirius/web/templates/pdf/handlers/PdfReplaceHandler.java
+++ b/src/main/java/sirius/web/templates/pdf/handlers/PdfReplaceHandler.java
@@ -108,20 +108,18 @@ public abstract class PdfReplaceHandler implements Priorized {
         }
 
         if (fsImage.getWidth() > cssWidth || fsImage.getHeight() > cssHeight) {
-            return downscaleResource(cssWidth, cssHeight, fsImage);
+            return downscaleResource(cssWidth, cssHeight, fsImage.getWidth(), fsImage.getHeight());
         }
 
         if (cssWidth > fsImage.getWidth() && cssHeight > fsImage.getHeight()) {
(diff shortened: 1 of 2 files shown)


sabieber (approved):

ymo-sci (approved):

fhaScireum (commented):
Don't we also need to upscale the height? Wouldn't we need to upscale the width if needed, upscale the height if needed, and then down scale both to fit?

bkositza:
> Don't we also need to upscale the height? Wouldn't we need to upscale the width if needed, upscale the height if needed, and then down scale both to fit?

Do you have an example of when we need to scale up the height and width? In my opinion, it is sufficient if we scale one of them up and then scale the other down.

sabieber:
Langsam wären ein paar Testcases für die Logik glaub garnicht so verkehrt

fhaScireum:
> > Don't we also need to upscale the height? Wouldn't we need to upscale the width if needed, upscale the height if needed, and then down scale both to fit?
> 
> Do you have an example of when we need to scale up the height and width? In my opinion, it is sufficient if we scale one of them up and then scale the other down.

Now that I think about it a bit more, the behaviour now should be correct.

fhaScireum (approved):

bkositza:
> Langsam wären ein paar Testcases für die Logik glaub garnicht so verkehrt

Done ;)

mko-sci (approved):

mko-sci (approved):

PR #1379, opened 2024-02-08
Bootstrap 5

jakobvogel:
### Description

Ports Tycho to Bootstrap 5.

### Additional Notes

- This PR fixes or works on following ticket(s): [SIRI-924](https://scireum.myjetbrains.com/youtrack/issue/SIRI-924)

### Checklist


src/main/resources/default/assets/tycho/lib.js.pasta
```diff
@@ -2,13 +2,10 @@
 // A list of all licenses can be obtained via /licenses
 //
 
-___include("/assets/tycho/libs/jquery/jquery-3.5.1.min.js")
-___include("/assets/tycho/libs/popper/popper-1.14.3.min.js")
+___include("/assets/tycho/libs/popper/popper.min.js")
 ___include("/assets/tycho/libs/bootstrap/js/bootstrap.min.js")
```
jakobvogel, line 6: We could also directly include such files in the HTML page. That would enable the browsers to load source maps, thus facilitating debugging. Merging everything into one file is primarily for SEO, as far as I understand it, which is not so relevant on backend pages.

src/main/resources/default/assets/wondergem/wondergem.css.pasta
```diff
@@ -2,8 +2,6 @@
  * A list of all licenses can be obtained via /licenses
  */
 
-<i:include name="/assets/wondergem/bootstrap/css/bootstrap.min.css"/>
-
```
sabieber, line 6: Is this change intended?

src/main/resources/default/assets/tycho/scripts/form.js.pasta
```diff
@@ -25,46 +25,56 @@ sirius.ready(function () {
...
+                _modalElement.addEventListener('shown.bs.modal', () => {
+                    const _cancelButton = _modalElement.querySelector('.btn-cancel-js');
+                    if (_cancelButton != null) {
+                        _cancelButton.focus();
+                    }
+                }, {
+                    once: true,
+                });
+
+                modal.show(null);
```
sabieber, line 54:
```suggestion
                modal.show();
```
Or does this not work? normally a not given parameter in JS should be undefined in the called method, so most probably the behaviour is the same

sabieber (commented):
Is an up-to-date LICENCE.TXT available for all updated libs?

mko-sci (approved):

src/main/resources/default/taglib/t/colorpicker.html.pasta
```diff
@@ -42,9 +49,17 @@
...
-        $(function () {
-            $('#@id').parent().colorpicker({
-                format: 'hex'
+        sirius.ready(() => {
+            const _textInput = document.querySelector('#@textId');
+            const _colorInput = document.querySelector('#@colorId');
+
+            _textInput.addEventListener('input', () => {
+                const color = _textInput.value;
+                _colorInput.value = color.match(/^#([0-9a-f]{6})$/i) ? color : '#000000';
```
ymo-sci, line 58: also allow shorthands like `#fff`?

jakobvogel: According to mozdev, the `value` of a `color` typed `<input>` is always a 7-character string, see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/color#value … I did not test all browsers, but i fear we would need to expand shorthand hexadecimal ourselves to be absolutely sure.

The main input element of the entire template is the text field anyway. If the shorthand or anything else is entered, the color preview will not be updated. (It will display black.) Still, upon submit, the raw value should be transmitted to the server.

If a color is picked via the auxiliary color input, though, we always obtain the 7-character string.

So, is there really a use-case for shorthand hexadecimal?

src/main/resources/default/assets/tycho/scripts/form.js.pasta
jakobvogel: I agree that yours is the better code, and it works without an issue…

However, (faulty?) Code Inspection wants to have the superfluous `null` there for some reason. I decided to put it in to suppress a pointless warning.

Screenshot 2024-02-09 at 18 22 37

In the light of this, which version do you prefer?

src/main/resources/default/taglib/t/colorpicker.html.pasta
jakobvogel: I have pushed a proposal that uses the browser itself for parsing the colors. This will lead the color preview to update for other valid color strings as well. _However, exactly this raw string is sent to the server, and not the computed `#rrggbb` value!_

https://github.com/scireum/sirius-web/assets/36069303/88109679-c6ec-48c4-9915-c4410966f34d

src/main/resources/default/assets/wondergem/wondergem.css.pasta
jakobvogel: Yes, the same file is included another time below in line 11 (previously line 13)

sabieber (approved):

src/main/resources/default/taglib/t/colorpicker.html.pasta
sabieber: > However, exactly this raw string is sent to the server, and not the computed #rrggbb value!
Are we sure this does not break server-side functionality then?
As in the code that uses the values stored via this input

fhaScireum (approved):

ymo-sci (approved):

src/main/resources/default/taglib/t/colorpicker.html.pasta
jakobvogel: This is correct, @sabieber. At the end of the day, it really depends on the use. Let's compare the two scenarios:

1. We treat the color input as primary source and always use/transmit the `#rrggbb` value:
   \+ We use a standardised format
   \+ We have some basic validation
   \+ The text field is just auxiliary input

2. We use the raw format as provided in the text field:
   \+ We support transparency
   \+ We support other color models such as CMYK, HSL, etc. and different notations
   \+ We support color names where applicable
   \- We rely on the user to provide useful input
   \- The color field is just auxiliary input that may show the wrong color (particularly transparency will not be shown)

Considering that the server should never trust client-side validation and thus validate itself, I took the second approach as it seems to be more versatile while relying on standard components as much as possible. If we need to support more complex scenarios (like different notations, different validated color models, etc.), we would need to define them more precisely.

sabieber: But we should keep the current uses of the field in mind while doing so. Don't know of the top of the top of my head which approach was used before, but we have a bunch of those fields and it would be kind of sad to allow users "too much" there which could break stuff.
Examples in OX are: portal design colors, stories design colors, brand colors, ...
And maybe S2 has current use cases as well.

jakobvogel: I have checked the previous implementation. It always sends hex colors. Therefore, I yield and restrict the code to also only send hex RGB strings. By the way, entering color strings manually in the previous implementation was outright unpleasant, unless one was only pasting values. The new component will now act as follows:

https://github.com/scireum/sirius-web/assets/36069303/a02abd2e-112b-4632-b647-3fa3bed65382

sabieber (approved):

PR #1383, opened 2024-02-20
Datasheet: Acquires a token before starting the thread

bkositza:
To ensure that users with smaller requests are not blocked by users with large requests, the system waits for a token from the semaphore before starting a new thread. 

This prevents countless threads from running and simply waiting.

### DIFF
--- a/src/main/java/sirius/web/templates/pdf/SemaphoreContext.java
+++ b/src/main/java/sirius/web/templates/pdf/SemaphoreContext.java
@@ -8,6 +8,8 @@
 
 package sirius.web.templates.pdf;
 
+import sirius.kernel.async.SubContext;
+
 import java.util.concurrent.Semaphore;
 
 /**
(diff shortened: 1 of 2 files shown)


sabieber (approved):

ymo-sci (approved):

src/main/java/sirius/web/templates/pdf/SemaphoreContext.java
```diff
@@ -8,34 +8,42 @@
...
 package sirius.web.templates.pdf;
 
+import sirius.kernel.async.SubContext;
+
 import java.util.concurrent.Semaphore;
 
 /**
  * This class is used to limit the number of concurrent resource handling operations.
  * <p>
  * This is used to avoid overloading the system with too many concurrent resource handling operations.
```
mkeckmkeck, line 18:
```suggestion
 * This class is used to limit the number of concurrent resource handling operations per belonging CallContext.
 * <p>
 *  The usecase is that we want to control the number of spawned threads to avoid overloading the 
 *  system with too many concurrent operations.
```

mkeckmkeck (approved):

sabieber (approved):

idlira (approved):

fhaScireum (approved):

ymo-sci (approved):

PR #1404, opened 2024-04-11
Fixes user scope in virtual thread

bkositza:
The UserScope in a virtual thread is always the default. We must therefore specify the ScopeID.


mko-sci (approved):

mkeckmkeck (commented):
Why not simply setting the outer Scope inside virtual thread execution? Might we even require a user?

bkositza:
> Why not simply setting the outer Scope inside virtual thread execution? Might we even require a user?

That is much simpler, thanks :)

mkeckmkeck (approved):

sabieber (approved):

PR #1407, opened 2024-04-19
Retries resolving image if any exception occurred

bkositza:
In some cases, the image we want to resolve is not yet available. It is therefore important to try again after a short waiting time.

This is the case, for example, when we request a variant of an image that has not yet been created.

### DIFF
--- a/src/main/java/sirius/web/templates/pdf/AsyncLoadedImageElement.java
+++ b/src/main/java/sirius/web/templates/pdf/AsyncLoadedImageElement.java
@@ -16,6 +16,7 @@ import org.xhtmlrenderer.render.BlockBox;
 import org.xhtmlrenderer.render.RenderingContext;
 import sirius.kernel.async.CallContext;
 import sirius.kernel.commons.Tuple;
+import sirius.kernel.commons.Wait;
 import sirius.kernel.health.Exceptions;
 import sirius.web.security.ScopeInfo;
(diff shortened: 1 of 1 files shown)


src/main/java/sirius/web/templates/pdf/AsyncLoadedImageElement.java
```diff
@@ -62,20 +64,44 @@ public AsyncLoadedImageElement(PdfReplaceHandler handler, String uri, int cssWid
...
+        }
+    }
+
+    private void retryOrFail(String uri, PdfReplaceHandler handler, Semaphore semaphore, int attempt, Exception exception) {
+        if (exception instanceof InterruptedException) {
+            // we should accept that the thread is interrupted and don't try again
+            Exceptions.handle(exception);
+            return;
+        }
+        if (attempt <= MAX_ATTEMPTS) {
```
sabieber, line 95:
```suggestion
        if (attempt < MAX_ATTEMPTS) {
```
I think this is an off by one error where we actually try 4 times if my thinking is correct

sabieber (approved):
Small tips for the next PR:
- Put valuable explanation also in the commit message body, not just the PR description, so it is available in the IDE
- Put method extractions and logic alterations in two separate commits, so the logic changes are more obvious in the diff

And thanks for your work on this 👍🏻

PR #1408, opened 2024-04-19
Retries resolving image if any exception occurred

bkositza:
In some cases, the image we want to resolve is not yet available. It is therefore important to try again after a short waiting time.

This is the case, for example, when we request a variant of an image that has not yet been created.


jakobvogel (approved):
Assuming the code is the same as in https://github.com/scireum/sirius-web/pull/1407

sabieber (approved):

sabieber:
> Assuming the code is the same as in #1407

Mainly, but the off-by-one-error seems to be fixed :)

fhaScireum (approved):

jmuscireum (approved):

mko-sci (approved):

PR #1412, opened 2024-04-25
Tooltips for `<t:iconInlineInfo>`

jakobvogel:
### Description

When using `<t:iconInlineInfo>` within `<t:datacard>`, users may want to know more about the information displayed, depending on how obvious the icon is. The new `title` property allows to set such information, which will be displayed as BS tooltip and rendered as ARIA label.

In order to make this work, the tooltip technically needs to become a link, thus overriding the card's own link. It is thus convenient to also pass the link property to allow seamless navigation.

### Screenie

Screenshot 2024-04-25 at 15 05 17

### DIFF
--- a/src/main/resources/default/taglib/t/iconInlineInfo.html.pasta
+++ b/src/main/resources/default/taglib/t/iconInlineInfo.html.pasta
@@ -2,23 +2,36 @@
 <i:arg type="String" name="value" default=""/>
 <i:arg type="Object" name="effectiveValue" default="isFilled(value) ? value : renderToString('body')"/>
 <i:arg type="String" name="class" default=""/>
+<i:arg name="title" type="String" default=""/>
+<i:arg name="link" type="String" default=""/>
 
 <i:pragma name="description">
     Renders an icon + value pair which is rendered as a single block.
 </i:pragma>
 
+<i:local name="wrapper"
+         value="@isFilled(link) ? 'a' : 'span'"/>
+
 <i:if test="isFilled(effectiveValue)">
     <span class="d-flex flew-row text-small mb-2 me-2 @class">
-        <span class="text-sirius-gray-dark ">
-            <i class="@icon"></i>
-        </span>
-        <span class="ps-1 overflow-hidden text-break">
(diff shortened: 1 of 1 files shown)


src/main/resources/default/taglib/t/iconInlineInfo.html.pasta
```diff
@@ -2,23 +2,36 @@
...
 <i:arg type="String" name="class" default=""/>
+<i:arg name="title" type="String" default=""/>
+<i:arg name="link" type="String" default=""/>
 
 <i:pragma name="description">
     Renders an icon + value pair which is rendered as a single block.
 </i:pragma>
 
+<i:local name="wrapper"
+         value="@isFilled(link) ? 'a' : 'span'"/>
```
sabieber, line 13: **Optional:** According to the BS5 [docs](https://getbootstrap.com/docs/5.0/components/tooltips/#disabled-elements) it should also work on spans or buttons, but I am fine with it becoming an anchor.

sabieber (approved):

idlira (approved):

src/main/resources/default/taglib/t/iconInlineInfo.html.pasta
jakobvogel: That is correct, the tooltip works on the span alone. However, the scenario is a bit more complex in our case:

A datacard with a link has a `<a class="stretched link">` in its body. This link covers the entire card, including the spans of this element, and thus disables the tooltip at first (or rather prevents its activation). In order to make the tooltip work, I need to lift up the span by increasing the `z-index` via the `card-link` class. That, on the other hand, makes the element no longer clickable, so basically punches out a hole where the card cannot be clicked. My solution to that is to allow attaching a link to the icon info. If that is set to the card's link, the tooltip works, and the entire card can be clicked like before.

Of course, this could be extended to a JS-based automatic solution, to figure out the link target *in-situ*.

sabieber: Thanks for the explanation 👌🏻

PR #1418, opened 2024-05-07
Resolving URLs that uses http but redirects to https

bkositza:
### Description

The newly implemented option of resolving images that is very efficiently does not work completely reliably in a few cases. For example, if we have URLs that are specified with http but automatically redirect to https. Although this happens very rarely, it should still work. The old, more expensive method via the UserAgentCallback handles this very well. As we only need this in a few exceptions, we only use the fallback if it has not worked twice with the new method.

### Additional Notes

- This PR fixes or works on following ticket(s): https://scireum.myjetbrains.com/youtrack/issue/SIRI-961/URI-Resolve-Fehler-bei-http-https
- This PR is related to PR: https://github.com/scireum/sirius-web/pull/1371

### Checklist

### DIFF
--- a/src/main/java/sirius/web/templates/pdf/ImageReplacedElementFactory.java
+++ b/src/main/java/sirius/web/templates/pdf/ImageReplacedElementFactory.java
@@ -69,7 +69,7 @@ public class ImageReplacedElementFactory extends ITextReplacedElementFactory {
         try {
             String protocol = Strings.split(src, "://").getFirst();
             PdfReplaceHandler handler = findHandler(protocol);
-            return new AsyncLoadedImageElement(handler, src, cssWidth, cssHeight);
+            return new AsyncLoadedImageElement(handler, uac, src, cssWidth, cssHeight);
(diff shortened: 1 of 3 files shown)


sabieber (approved):

src/main/java/sirius/web/templates/pdf/ImageReplacedElementFactory.java
```diff
@@ -69,7 +69,7 @@ public ReplacedElement createReplacedElement(LayoutContext c,
         try {
             String protocol = Strings.split(src, "://").getFirst();
             PdfReplaceHandler handler = findHandler(protocol);
-            return new AsyncLoadedImageElement(handler, src, cssWidth, cssHeight);
+            return new AsyncLoadedImageElement(handler, uac, src, cssWidth, cssHeight);
```
idlira, line 72: Can we rename these the variables c and uac to something more descriptive?

idlira (approved):

PR #1454, opened 2024-09-04
Returns the oauth access token "expires at" date

mkeckmkeck:
### Description
Returns the oauth access token "expires at" date in case the server did return it. 
Providing this information e.g. gives the API caller the possibility to refresh the token if required. The usage is optional, if not set, the implemention in sirius will trigger a refresh if the provided access token could not be used to authenticate at server. 

Technically this is breaking. But it should have no impact on API users as only the framework should use the constructor with the changed parameters.

### Additional Notes

- This PR fixes or works on following ticket(s): [SE-13502](https://scireum.myjetbrains.com/youtrack/issue/SE-13502)

### Checklist

### DIFF
--- a/src/main/java/sirius/web/security/oauth/ReceivedTokens.java
+++ b/src/main/java/sirius/web/security/oauth/ReceivedTokens.java
@@ -44,7 +44,8 @@ public record ReceivedTokens(String accessToken, String refreshToken, String typ
         String refreshToken = response.required(OAuth.REFRESH_TOKEN).asText("");
         String type = response.required(OAuth.TOKEN_TYPE).asText("");
         long accessTokenExpiresIn = response.path(OAuth.EXPIRES_IN).asLong(0L);
-        LocalDateTime accessTokenExpiresAt = LocalDateTime.now().plusSeconds(accessTokenExpiresIn);
+        LocalDateTime accessTokenExpiresAt =
+                accessTokenExpiresIn > 0 ? LocalDateTime.now().plusSeconds(accessTokenExpiresIn) : null;
         if (OAuth.TOKEN_TYPE_BEARER.equalsIgnoreCase(type)) {
             try {
                 // Try to read the exact refresh token expiration date from the JWT token itself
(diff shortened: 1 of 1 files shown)


src/main/java/sirius/web/security/oauth/ReceivedTokens.java
```diff
@@ -22,10 +22,12 @@
...
  * @param refreshToken          the refresh token received from the authorization server
  * @param type                  the type of the tokens received from the authorization server, e.g. "Bearer" or "MAC"
+ * @param accessTokenExpiresAt  the date at which the access token expires, might be null if the server response
+ *                              contains no information
  * @param refreshTokenExpiresAt the date at which the refresh token expires, might be estimated if no JWT bearer
  *                              token is given
  */
-public record ReceivedTokens(String accessToken, String refreshToken, String type,
+public record ReceivedTokens(String accessToken, String refreshToken, String type, LocalDateTime accessTokenExpiresAt,
                              LocalDateTime refreshTokenExpiresAt) {
```
sabieber, line 31: **Optional:** This formatting of the constructor parameters looks a bit weird. Is this autoformatted with our default formatter settings?

sabieber (approved):

src/main/java/sirius/web/security/oauth/ReceivedTokens.java
```diff
@@ -22,10 +22,12 @@
  * @param accessToken           the access token received from the authorization server
  * @param refreshToken          the refresh token received from the authorization server
  * @param type                  the type of the tokens received from the authorization server, e.g. "Bearer" or "MAC"
+ * @param accessTokenExpiresAt  the date at which the access token expires, might be null if the server response
+ *                              contains no information
```
jakobvogel, line 26: Are you sure this field is `null` if the server response does not contain such information? At a quick glance, your code below seems to fall back to _expires in_ = 0, leading to _expires at_ to equal "now". Or am I wrong?

src/main/java/sirius/web/security/oauth/ReceivedTokens.java
mkeckmkeck: This is autoformatted. I cannot see an issue on my formatter settings. I case it is wrong, please ping me.

src/main/java/sirius/web/security/oauth/ReceivedTokens.java
mkeckmkeck: You are right. Thanks for the finding!

jakobvogel (approved):

sabieber (approved):

jmuscireum (approved):

PR #1455, opened 2024-09-04
HTML Input for PDF Generation

jakobvogel:
### Description

The _Flying Saucer_ PDF renderer requires XHTML as input, and we rely on the invoker to provide valid content. However, luckily, the world has moved on from XHTML. In order to relax the requirements and facilitate reuse of HTML templates, this PR introduces an additional processing step that removes problematic elements and ensures XHTML automatically.

Marked as **breaking** as PDF rendering should be checked after upgrading.

### Additional Notes

- This PR fixes or works on following ticket(s): [OX-10783](https://scireum.myjetbrains.com/youtrack/issue/OX-10783)

### Checklist

### DIFF
--- a/src/main/java/sirius/web/templates/pdf/TagliatellePDFContentHandler.java
+++ b/src/main/java/sirius/web/templates/pdf/TagliatellePDFContentHandler.java
@@ -8,6 +8,8 @@
 
 package sirius.web.templates.pdf;
 
+import org.jsoup.Jsoup;
+import org.jsoup.nodes.Document;
 import org.xhtmlrenderer.pdf.ITextRenderer;
 import sirius.kernel.di.std.Part;
 import sirius.kernel.di.std.Register;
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/web/templates/pdf/TagliatellePDFContentHandler.java
```diff
@@ -53,14 +55,37 @@ public boolean generate(Generator generator, OutputStream out) throws Exception
         ITextRenderer renderer = new ITextRenderer();
         renderer.getSharedContext()
                 .setReplacedElementFactory(new ImageReplacedElementFactory(renderer.getOutputDevice()));
-        renderer.setDocumentFromString(content);
+        renderer.setDocumentFromString(cleanHtml(content));
```
sabieber, line 58: **Optional:** Maybe this could be enabled optionally? Don't know if it could break stuff in S2

jakobvogel: Hmm, how? As an additional pseudo-parameter in the context? As a system-wide setting? Both do not seem to be particularly elegant 😇

Let's see what S2 is going to say about this PR, okay? We can also flag it as _breaking_, but it should process all previously valid input as it was, and only affect the behaviour in case of faulty input.

mkeckmkeck (approved):

PR #1457, opened 2024-09-05
Filter out hidden elements in sirius.addElementVisibleListener

jmuscireum:
### Description
Hidden elements have a bounding client rect at the top of the viewport and thus the listener is triggered immediately after the page is loaded.

This adds a simple check for visibility by checking the width and height of the bounding client rect.

### Additional Notes

- This PR fixes or works on following ticket(s): [SE-13968](https://scireum.myjetbrains.com/youtrack/issue/SE-13968)

### Checklist

### DIFF
--- a/src/main/resources/default/assets/common/core.js.pasta
+++ b/src/main/resources/default/assets/common/core.js.pasta
@@ -381,6 +381,12 @@ sirius.addElementVisibleListener = function (selector, listener, distanceFactor)
         const factor = distanceFactor || 1;
         const rect = element.getBoundingClientRect();
 
+        // Simple visibility check to ignore hidden elements
+        if (rect.height === 0 || rect.width === 0) {
+            return false;
+        }
+
+        // Check the position relative to the view port
         const verticalMargin = rect.height * factor;
         const horizontalMargin = rect.width * factor;
         const clientHeight = document.documentElement.clientHeight;


fhaScireum (approved):

mkeckmkeck (approved):

src/main/resources/default/assets/common/core.js.pasta
```diff
@@ -381,6 +381,12 @@ sirius.addElementVisibleListener = function (selector, listener, distanceFactor)
         const factor = distanceFactor || 1;
         const rect = element.getBoundingClientRect();
 
+        // Simple visibility check to ignore hidden elements
+        if (rect.height === 0 || rect.width === 0) {
```
sabieber, line 385: **Optional:** Wouldn't [checkVisibility](https://developer.mozilla.org/en-US/docs/Web/API/Element/checkVisibility) be a saver approach for modern browsers?
But I am also fine with the current fix until we see a false-positive (a zero size element that isn't actually hidden).

sabieber (approved):

jakobvogel (approved):

src/main/resources/default/assets/common/core.js.pasta
sabieber: Would be possible if a function like `sirius.isElementVisible` (or something like that) is added with something like this:

```js
sirius.isElementVisible = function (_element) {
	if (_element.checkVisibility) {
		return _element.checkVisibility();
	} else {
		const rect = _element.getBoundingClientRect();
		return rect.height > 0 || rect.width > 0;
	}
}
```

Not sure if `||` or `&&` would be better in the last condition

mkeckmkeck (approved):

sabieber (approved):
Thx 👌🏻

scireum-mbo (approved):

ymo-sci (approved):

mko-sci (approved):

PR #1464, opened 2024-10-01
SE-13618: Extend the checks of the TemplateCompiler to include missing required attributes

oolscireum:
### Description

The TemplateCompiler currently ignores if there are missing attributes that are required because of a missing default value. This circumstance ist only detected during the rendering of the page.

The changes to the TemplateCompiler is now checking for these missing attributes as an early warning to the programmer. Especialy if the templates are overwritten by an artefact.

### Additional Notes

- This PR fixes or works on following ticket(s): [SE-13618](https://scireum.myjetbrains.com/youtrack/issue/SE-13618)

### Checklist

### DIFF
--- a/src/main/java/sirius/pasta/tagliatelle/compiler/TemplateCompiler.java
+++ b/src/main/java/sirius/pasta/tagliatelle/compiler/TemplateCompiler.java
@@ -434,7 +434,7 @@ public class TemplateCompiler extends InputProcessor {
      * @param attributes the set of attributes which were parsed
      */
     private void checkMissingAttributes(TagHandler handler, Set<String> attributes) {
-        Set<String> missingAttributes = handler.getRequiredAttributeNames();
+        Set<String> missingAttributes = new HashSet<>(handler.getRequiredAttributeNames());
         missingAttributes.removeAll(attributes);
         if (!missingAttributes.isEmpty()) {
             missingAttributes.forEach(attr -> context.error(reader.current(),
--- a/src/main/java/sirius/pasta/tagliatelle/compiler/TemplateCompiler.java
+++ b/src/main/java/sirius/pasta/tagliatelle/compiler/TemplateCompiler.java
@@ -8,14 +8,14 @@
 
 package sirius.pasta.tagliatelle.compiler;
 
-import sirius.kernel.tokenizer.Char;
-import sirius.kernel.tokenizer.ParseError;
-import sirius.kernel.tokenizer.Position;
 import sirius.kernel.commons.Explain;
 import sirius.kernel.commons.Strings;
 import sirius.kernel.di.std.PriorityParts;
 import sirius.kernel.health.Exceptions;
 import sirius.kernel.health.HandledException;
+import sirius.kernel.tokenizer.Char;
+import sirius.kernel.tokenizer.ParseError;
+import sirius.kernel.tokenizer.Position;
 import sirius.pasta.Pasta;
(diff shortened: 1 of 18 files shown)


src/main/java/sirius/pasta/tagliatelle/tags/ArgTag.java
```diff
@@ -157,4 +159,9 @@ public Class<?> getExpectedAttributeType(String name) {
 
         return super.getExpectedAttributeType(name);
     }
+
+    @Override
+    public Set<String> getRequiredAttributeNames() {
+        return new HashSet<>(List.of(PARAM_NAME, PARAM_TYPE));
```
sabieber, line 165: **Question:** Why is `Set.of` not used here instead of creating a list and then turning that list into a set?

sabieber (approved):

src/main/java/sirius/pasta/tagliatelle/compiler/TemplateCompiler.java
```diff
@@ -379,6 +383,8 @@ private void parseAttributes(TagHandler handler) {
             verifyAttributeNameAndType(handler, name, attributeType);
             if (attributeType == null) {
                 attributeType = String.class;
+            } else {
```
idlira, line 386: Question: why an else? An attribute without a type just defaults to String, but its still an attribute or am I missing something?

oolscireum: When there is no attributeType it already creates the error "Unknown attribute." That means, the attribute is not defined in the tag. Therefore i think the attribute is not needed in the Set.

src/main/java/sirius/pasta/tagliatelle/tags/ArgTag.java
oolscireum: The modifiable list is needed because the calling method is removing the recognized attributes. I redo the the methods in the tags with Set.of(...) and make the new HashSet(...) in the check method.

scireum-mbo (approved):

idlira (approved):

idlira (approved):

sabieber (approved):

src/main/java/sirius/pasta/tagliatelle/compiler/TemplateCompiler.java
```diff
@@ -417,6 +425,25 @@ private void parseAttributes(TagHandler handler) {
...
+     * @param attributes the set of attributes which were parsed
+     */
+    private void checkMissingAttributes(TagHandler handler, Set<String> attributes) {
+        Set<String> missingAttributes = new HashSet<>(handler.getRequiredAttributeNames());
+        missingAttributes.removeAll(attributes);
+        if (!missingAttributes.isEmpty()) {
+            missingAttributes.forEach(attr -> context.error(reader.current(),
+                                                            "Missing required attribute. %s missing the required attribute '%s'.",
+                                                            handler.getTagName(),
+                                                            attr));
```
mkeckmkeck, line 445: Style: würd mir hier die mutability sparen und direkt nen stream drüber laufen lassen: 
So ungefähr: requiredAttribute.stream.filter(predicate.not(foundAttributes::contains).foreach...

mkeckmkeck (commented):
Importe noch aufräumen

fhaScireum (approved):

mkeckmkeck (approved):

PR #1469, opened 2024-10-23
CSS Design System - Switch from rem to px based CSS variables

sabieber:
### Description

Reason:
We use the CSS design system for our integrations into other websites, where we don't have control over the base font size. Because of this, a bunch of customers have problems with small text / elements / margins, as they specify a base font size of 10 px instead of 16 px.
To mitigate this, we switch to pixels and allow customers to easily alter the used sizes.

The following CSS variables can be used to very easily scale and alter the UI:

`--sci-base-unit` -> allows to scale pretty much anything of the UI (font-sizes, paddings, borders, ...)
`--sci-font-size-base` -> allows to scale all text while keeping the ratios between different size classes (like `-smaller` or `-h1`)
`--sci-border-radius` -> for example allows to disable roundings on cards, buttons and inputs when set to `0`
...

With this, we also provide a polyfill for IE 11 that should make most if not all of this work again.

**Without polyfill:**

Bildschirmfoto 2024-10-23 um 22 31 02

**With polyfill:**

Bildschirmfoto 2024-10-23 um 22 30 38

### Additional Notes

- This PR fixes or works on following ticket(s): [SIRI-](https://scireum.myjetbrains.com/youtrack/issue/SIRI-)
- This PR is related to PR: 

### Checklist

- [ ] Code change has been tested and works locally


src/main/resources/assets/design-system/misc.scss
```diff
@@ -227,10 +227,10 @@
 }
 
 .sci-pill {
-  padding: 0.25rem 0.5rem;
-  border-radius: 1rem;
+  padding: calc(0.25 * var(--sci-base-unit)) calc(0.5 * var(--sci-base-unit));
+  border-radius: calc(4* var(--sci-border-radius));
```
jakobvogel, line 231:
```suggestion
  border-radius: calc(4 * var(--sci-border-radius));
```

jakobvogel (approved):
Very nice so far 🤩

idlira (approved):

mko-sci (approved):

PR #1477, opened 2024-11-13
Implements icons for the new watchlist feature in OX

scireum-mbo:
### Additional Notes

- This PR fixes or works on following ticket(s): [OX-11593](https://scireum.myjetbrains.com/youtrack/issue/OX-11593)
- This PR is related to PR: https://github.com/scireum/oxomi/pull/11727

### Checklist

### DIFF
--- a/src/main/resources/assets/design-system/icons.scss
+++ b/src/main/resources/assets/design-system/icons.scss
@@ -125,3 +125,11 @@
 .sci-icon-magnifier-text-soft {
   background-image: url(svgResource('/assets/images/icons/magnifier.svg', #212529));
 }
+
+.sci-icon-trash {
+  background-image: url(svgResource('/assets/images/icons/trash.svg', #53595D));
+}
+
+.sci-icon-close {
+  background-image: url(svgResource('/assets/images/icons/navigation/close.svg', #53595D));
+}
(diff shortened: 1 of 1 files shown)


src/main/resources/assets/design-system/icons.scss
```diff
@@ -125,3 +125,11 @@
 .sci-icon-magnifier-text-soft {
   background-image: url(svgResource('/assets/images/icons/magnifier.svg', #212529));
 }
+
+.sci-icon-trash {
+  background-image: url(svgResource('/assets/images/icons/trash.svg', #53595D));
```
jmuscireum, line 130: Shouldn't this by default use the same colour as the other icons? Can you use for example opacity to make it lighter in OXOMI?

jakobvogel (approved):

mko-sci (approved):

src/main/resources/assets/design-system/icons.scss
sabieber: Yes, this also doesn't seem to be a "sirius color".
Would prefer to use a color suffix appended to the class like the one above.
I think using text-soft should also be fine for the icons in the watchlist, but feel free to post a screenshot after the change. Its no problem to add additional color variants of the icons with different color suffixes.

scireum-mbo:
Before color change:
image

After:
image

sabieber (approved):

jmuscireum (approved):

PR #1480, opened 2024-11-19
Prevents the row from being selected when clicked on an i-tag

mko-sci:
### Additional Notes

- This PR fixes or works on following ticket(s): [SIRI-11094](https://scireum.myjetbrains.com/youtrack/issue/SIRI-11094)

### Checklist

### DIFF
--- a/src/main/resources/default/assets/tycho/scripts/selectableTable.js.pasta
+++ b/src/main/resources/default/assets/tycho/scripts/selectableTable.js.pasta
@@ -32,6 +32,7 @@ sirius.ready(function () {
             _row.addEventListener('click', function (event) {
                 if (event.target.tagName.toLowerCase() !== 'input' &&
                     event.target.tagName.toLowerCase() !== 'a' &&
+                    event.target.tagName.toLowerCase() !== 'i' &&
                     event.target.tagName.toLowerCase() !== 'button') {
                     _checkbox.checked = !_checkbox.checked;
                     _checkbox.dispatchEvent(new Event('change'));


src/main/resources/default/assets/tycho/scripts/selectableTable.js.pasta
```diff
@@ -32,6 +32,7 @@ sirius.ready(function () {
             _row.addEventListener('click', function (event) {
                 if (event.target.tagName.toLowerCase() !== 'input' &&
                     event.target.tagName.toLowerCase() !== 'a' &&
+                    event.target.tagName.toLowerCase() !== 'i' &&
                     event.target.tagName.toLowerCase() !== 'button') {
```
sabieber, line 36: Wouldn't it be better to maintain a whitelist or use a marker class that possible click-targets get assigned?

sabieber (approved):

src/main/resources/default/assets/tycho/scripts/selectableTable.js.pasta
jakobvogel: > Icons are often wrapped in `<i>`-tags. If you add an `<a>` which has an `<i>` inside and you manage to click directly on the `<i>` (or the icon) you would select the row. In this case however the intention was to click on the `<a>`-button.

In case of FA, the `<i>` tag only serves as a marker and is replaced by a `<svg>` tag if i am not wrong.

scireum-mbo (approved):

mko-sci:
Merged because This PR just extends the current mechanism.

Obvoiusly we need to rework the mechansim but there is currently no time for that since we are close the release.

PR #1483, opened 2024-11-27
Lokalise: Translations update

scireumskip:
This pull request was initiated by Lokalise (user Marcus Korinth) at 2024-11-27 09:41:26


src/main/resources/web_en.properties
```diff
@@ -44,10 +44,14 @@ fileUpload.URL = URL
...
 help = Help
+selectableTable.all.deselection = Deselect all
+selectableTable.all.selection = Select all
+selectableTable.row.deselection = Deselect
+selectableTable.row.selection = Select
 table.html.empty = No contents found
 template.html.additionalActions = Additional Actions
 template.html.confirmHeader = Are you sure?
-template.html.confirmText = Press 'Yes' to continue or Cancel to go back.
+template.html.confirmText = Press 'Yes' to continue or 'Cancel' to go back.
```
jakobvogel, line 54: This line was only adapted in English, not in the other languages. What about the German version?

jakobvogel (approved):

scireum-mbo (approved):

idlira (approved):

src/main/resources/web_en.properties
mko-sci: `Drücken Sie 'Ja', um fortzufahren oder 'Abbrechen', um zurückzugehen.`

mko-sci: The english version has been enhanced by the quotes. The German original already came with quotes (although not the correct ones... but we need to fix the quotation in sirius anyway as this is not the only key)

mkeckmkeck (approved):

src/main/resources/web_en.properties
jakobvogel: Look at _all_ the other languages as well, plz. There, the new quotes seem to be missing.

mko-sci: Yes, they are missing there indeed.

PR #1495, opened 2025-01-15
Design System additions

jmuscireum:
### Description
This adds additional classes to the scireum Design System.

#### Breaking
Uses of `--sci-margin-05` need to be replaced with `--sci-margin-0_5`.

### Additional Notes

- This PR fixes or works on following ticket(s): [SE-13735](https://scireum.myjetbrains.com/youtrack/issue/SE-13735)

### Checklist

### DIFF
--- a/src/main/resources/assets/design-system/margin-padding.scss
+++ b/src/main/resources/assets/design-system/margin-padding.scss
@@ -5,9 +5,6 @@
   --sci-margin-2: calc(2 * var(--sci-margin-1)); // Default: 16px / 1rem
   --sci-margin-3: calc(3 * var(--sci-margin-1)); // Default: 24px / 1.5rem
   --sci-margin-4: calc(4 * var(--sci-margin-1)); // Default: 32px / 2rem
-
-  // Alternative variable for backwards compatability
-  --sci-margin-05: --sci-margin-0_5; // Default: 4px / 0.25rem
 }
 
 .sci-m-0.5 {
--- a/src/main/resources/assets/design-system/margin-padding.scss
+++ b/src/main/resources/assets/design-system/margin-padding.scss
@@ -1,13 +1,22 @@
 :root {
   --sci-margin-1: calc(0.5 * var(--sci-base-unit)); // Default: 8px / 0.5rem
+  --sci-margin-15: calc(1.5 * var(--sci-margin-1)); // Default: 12px / 0.75rem
   --sci-margin-05: calc(0.5 * var(--sci-margin-1)); // Default: 4px / 0.25rem
   --sci-margin-2: calc(2 * var(--sci-margin-1)); // Default: 16px / 1rem
   --sci-margin-3: calc(3 * var(--sci-margin-1)); // Default: 24px / 1.5rem
   --sci-margin-4: calc(4 * var(--sci-margin-1)); // Default: 32px / 2rem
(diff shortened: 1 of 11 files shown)


mko-sci (approved):

src/main/resources/assets/design-system/margin-padding.scss
```diff
@@ -1,15 +1,24 @@
 :root {
   --sci-margin-1: calc(0.5 * var(--sci-base-unit)); // Default: 8px / 0.5rem
+  --sci-margin-15: calc(1.5 * var(--sci-margin-1)); // Default: 12px / 0.75rem
```
sabieber, line 3: **Optional:** Actually I hate this 😅 What differentiates this from a real `-15`?
I know I introduced this with the `05` below (inspired by SCSS vars I think) but think we should change the pattern for decimal values

sabieber (approved):

idlira (approved):

src/main/resources/assets/design-system/margin-padding.scss
jmuscireum: Do you have a suggestion?

sabieber: Tailwind for example simply includes Dots in the Class-Names, like `mt-1.5`.
Maybe we could use `_` as a separator if Dots are not allowed in CSS-Variable Names?

jakobvogel (approved):

src/main/resources/assets/design-system/margin-padding.scss
jmuscireum: I adjusted the class and variable names for 0.5 and 1.5
For the margin variable I kept the old name as well, so it's non-breaking. All other names were only used in S2 so far.

src/main/resources/assets/design-system/margin-padding.scss
```diff
@@ -1,15 +1,27 @@
...
   --sci-margin-1: calc(0.5 * var(--sci-base-unit)); // Default: 8px / 0.5rem
-  --sci-margin-05: calc(0.5 * var(--sci-margin-1)); // Default: 4px / 0.25rem
+  --sci-margin-0_5: calc(0.5 * var(--sci-margin-1)); // Default: 4px / 0.25rem
+  --sci-margin-1_5: calc(1.5 * var(--sci-margin-1)); // Default: 12px / 0.75rem
   --sci-margin-2: calc(2 * var(--sci-margin-1)); // Default: 16px / 1rem
   --sci-margin-3: calc(3 * var(--sci-margin-1)); // Default: 24px / 1.5rem
   --sci-margin-4: calc(4 * var(--sci-margin-1)); // Default: 32px / 2rem
+
+  // Alternative variable for backwards compatability
+  --sci-margin-05: --sci-margin-0_5; // Default: 4px / 0.25rem
```
sabieber, line 10: Can also live with removing this var, marking it as breaking and renaming the few places where we use it in OX

sabieber (approved):
Thanks 👌🏻

src/main/resources/assets/design-system/margin-padding.scss
jmuscireum: Alighty

sabieber (approved):

PR #1508, opened 2025-02-27
Adds multiple icon classes and allows changing the color

mko-sci:
### Description

These icons are used in the new OX-Watchlist feature.

### Additional Notes

- This PR fixes or works on following ticket(s): [OX-11901](https://scireum.myjetbrains.com/youtrack/issue/OX-11901)

### Checklist


sabieber (approved):

src/main/resources/assets/design-system/icons.scss
```diff
@@ -157,3 +157,99 @@
 .sci-icon-zoom-out-blue {
   background-image: url(svgResource('/assets/images/icons/navigation/zoom-out.svg', #3465CC));
 }
+
+.sci-icon-bars-gray-dark {
+  background-image: url(svgResource('/assets/images/icons/fontawesome/bars.svg', $sirius-gray-dark));
+}
+
+.sci-icon-bars-blue {
+  background-image: url(svgResource('/assets/images/icons/fontawesome/bars.svg', #3465CC));
```
jakobvogel, line 166: I see that we use the explicit color code elsewhere as well – but do we need to? Is there no matching colour in the palette of predefined colours? If not, why do we need particularly that blue here then?

jakobvogel (approved):

src/main/resources/assets/design-system/icons.scss
mko-sci: Thats exactly what I thought, but no - there is no predefined color for the blue - and also when you take a look into the file - there are a lot of custom colors in there already. 
I am wondering why we even started with the `sirius-colors` when no one really cares..

mko-sci: > If not, why do we need particularly that blue here then?
I dont know - its the blue we use in OX within the Datasheet and Infoplay dialogs and other places aswell...

jakobvogel: `$sirius-primary` looks close enough, doesn't it? – if we want to port OX to use that color as well, of course

mko-sci: yup - but I think this should be done in a separate ticket - especially since we would neet to change a lot of places in OX as well

sabieber: @mko-sci Will you create a ticket for this?

PR #1641, opened 2026-05-28
BREAKING: Checks for CSRF token for All POST Requests by default

fhaScireum:
### Description

POST requests that can be secured by CSRF tokens should be secured by these by default from a security standpoint.

Exceptions are POST requests without an existing session (e.g. login requests) or requests from external systems (e.g. API routes invoked by our clients). These must now be specifically exempt. **This means, the logic which routes check for a CSRF token and which don't has now been inverted!**

---

**BREAKING CHANGES:**
- ALL routes that MAY be invoked via a POST request MUST now either ensure that a CSRF token is sent as part of the request (by providing it in the templates) or they must exempt the route (via `skipCsrfValidation=true` in the `Routed` annotation or by exempting all routes in a controller or via the system or user config).
- `WebContext#hidePost()` was renamed to `WebContext#hidePostRequest`.
- `WebContext#isUnsafePOST()` was renamed to `WebContext#isPostRequest`.
- `WebContext#isSafePOST()` was removed. The CSRF validation is now handled automatically by the dispatcher, so you can be sure that if `WebContext#isPostRequest` returns true and the route is not exempt, the CSRF token will have been validated already and deemed valid.
- `WebContext#ensureSafePOST` was removed. It essentially worked the same as `WebContext#isSafePOST()` but would throw in case the request is a POST request but without providing a CSRF token. Now invoking `WebContext#isPostRequest` effectively does the same thing, althouth the exception in case of a missing CSRF token will be thrown earlier by the dispatcher.
- `BasicController#enforceMethodPost` was removed. Declare `methods = HttpMethod.POST` in the `Routed` annotation to ensure it is only called via POST.
- Controllers must now implement the isSkipCsrfValidation(), if they do not extend BasicController. However, most (if not all) controllers inherit it from BasicController.
- t:dropdownDeleteItem with `withConfirm=false` now uses POST (with a CSRF token) instead of GET.

---

Migration help:

1) Methods that would throw on non-POST requests like this
```
if (!ctx.isUnsafePOST()) {
  throw Exceptions.createHandled().withNLSKey("MyController.notPOST").handle();
}
```
should now instead declare `methods = HttpMethod.POST` in the Routed annotation. Alternatively they can check `WebContext#isPostRequest` in case they want to throw a very specific error. However, enforcing it via the annotation is preferable as that will result in a proper 403 HTTP status and in most cases, invoking POST routes via GET is only done by accident or by attackers or bots.

2) Methods that do something extra when the request is a POST request without a CSRF token like this
```
if (ctx.isUnsafePOST()) {
  loadConfiguration(ctx, configuration);
}
```
should now instead check `WebContext#isPostRequest`. Additional, it must now be checked whether the call happens from inside a template, meaning code we control, or from an external system. In the case of the former, the template must now provide a CSRF token via `part(sirius.web.http.CSRFHelper.class).getCSRFToken()`, in the case of the later the CSRF check must be skipped via declaring `skipCsrfValidation = true` in the Routed annotation.

3) Methods that do something extra when the request is a safe POST request like this
```
@Routed("/foo-bar")
public void sendOfferMail(WebContext ctx) {
    if (ctx.isSafePOST()) {
        sendMails(ctx);
    }

    ctx.respondWith().template("/templates/xyz.html.pasta");
}
```
or like this
```
if (webContext.ensureSafePOST()) {
    somethingExtra();
}
webContext.respondWith().redirectToGet("/xyz");
```
were already secured by a CSRF token check. Therefore they can now simply check WebContext#isPostRequest instead of the now removed WebContext#isSafePOST and WebContext#ensureSafePOST methods to do the extra actions only when a CSRF-protected POST request is being processed.

4) Methods that do something wild where they check for a safe POST request (which includes the CSRF token check) but don't throw if it fails like this
```
try {
    if (ctx.ensureSafePOST() && checkParameters(ctx) && checkValidation(ctx)) {
        ctx.respondWith().redirectToGet("/start");
        return;
    }
} catch (HandledException e) {
    UserContext.message(Message.error(e));
}
ctx.respondWith().template("xyz.html.pasta");
```
should now skip the automatic check by declaring `skipCsrfValidation = true` in the Routed annotation. Instead, they must manually check for `WebContext#isPostRequest` and invoke the check via `csrfHelper.hasValidCsrfToken(webContext)`. (Note: CSRFHelper is despite of its name a part class and not a common helper!)

5) Note, that some some elements like the `t:dropdownDeleteItem` tag or or the "confirm-link-js" css class already ensured that the request would be a POST including a CSRF token. Hoewever, it should still be checked whether the route could be closed down to only allow POST requests.

---

Because this has a massive impact on potentially hundreds of routes, there are many different ways to skip the check:
- Declare `skipCsrfValidation = true` in the Routed annotation of a method.
- Override `isSkipCsrfValidation` method in a controller and return true to skip the check for all of its routes.
- Exempt the route via system settings (for ease of use, the raw routed string is now propagated so no re-writes and co. must be considered). This however requires a system restart.
- System wide kill switch for all routes is still supported.

---

sirius-web routes overview: Routes with no CSRF info **CAN ONLY** be accessed via GET.
```
| Route                   | CSRF  | Notes                                                   |
|-------------------------|-------|---------------------------------------------------------|
| /api                    |       |                                                         |
| /barcode                |       |                                                         |
| /captcha-challenge      |       |                                                         |
| /licenses               |       |                                                         |
| /qr                     |       |                                                         |
| /system/api             |       |                                                         |
| /system/api/:1          |       |                                                         |
| /system/console         |       |                                                         |
| /system/console/api     | added | POST $.post with CSRFToken in console UI                |
| /system/fail            |       |                                                         |
| /system/info            |       |                                                         |
| /system/load            |       |                                                         |
| /system/metric/:1       |       |                                                         |
| /system/metrics         |       |                                                         |
| /system/monitor         |       |                                                         |
| /system/ok              |       |                                                         |
| /system/reset           |       |                                                         |
| /system/scope-config    |       |                                                         |
| /system/scope-config/:1 |       |                                                         |
| /system/state           |       |                                                         |
| /system/tag/:1/:2       |       |                                                         |
| /system/tags            |       |                                                         |
| /system/tags.md         |       |                                                         |
| /system/tags/state      |       | state change still reachable via GET ?reset             |
| /system/timing          |       | state change still reachable via GET ?enable / ?disable |
```

---

See commits for details :)

### Additional Notes

- This PR fixes or works on following ticket(s): [SIRI-1216](https://scireum.myjetbrains.com/youtrack/issue/SIRI-1216)

### Checklist

### DIFF
--- a/src/main/java/sirius/web/controller/BasicController.java
+++ b/src/main/java/sirius/web/controller/BasicController.java
@@ -265,6 +265,11 @@ public class BasicController implements Controller {
         out.endResult();
     }
 
+    @Override
+    public boolean isSkipCsrfValidation() {
+        return false;
+    }
+
     private StructuredOutput createStructuredOutput(WebContext webContext,
                                                     Format format,
                                                     HttpResponseStatus status,
--- a/src/main/java/sirius/web/controller/BasicController.java
+++ b/src/main/java/sirius/web/controller/BasicController.java
@@ -248,11 +248,6 @@ public class BasicController implements Controller {
         out.endResult();
     }
 
-    @Override
-    public boolean isSkipCsrfValidation() {
-        return false;
-    }
-
     private StructuredOutput createStructuredOutput(WebContext webContext,
                                                     Format format,
                                                     HttpResponseStatus status,
--- a/src/main/java/sirius/web/controller/BasicController.java
+++ b/src/main/java/sirius/web/controller/BasicController.java
@@ -111,23 +111,6 @@ public class BasicController implements Controller {
     }
 
     /**
-     * Throws an error which yields an HTTP 405 (Method Not Allowed) status if the request method isn't a POST.
-     * <p>
-     * Note that this doesn't enforce CSRF tokens.
-     *
-     * @see WebContext#isUnsafePOST()
-     * @see WebContext#ensureSafePOST()
-     */
-    protected void enforceMethodPost(WebContext webContext) {
-        if (!webContext.isUnsafePOST()) {
-            throw Exceptions.createHandled()
-                            .withDirectMessage("A POST request is expected.")
-                            .hint(HTTP_STATUS, HttpResponseStatus.METHOD_NOT_ALLOWED.code())
-                            .handle();
-        }
-    }
-
-    /**
      * Asserts that the given object is non-null.
      * <p>
      * Throws an appropriate error if the object is <tt>null</tt>.
--- a/src/main/java/sirius/web/controller/ControllerDispatcher.java
+++ b/src/main/java/sirius/web/controller/ControllerDispatcher.java
@@ -11,7 +11,6 @@ package sirius.web.controller;
 import io.netty.handler.codec.http.HttpHeaderNames;
 import io.netty.handler.codec.http.HttpMethod;
 import io.netty.handler.codec.http.HttpResponseStatus;
-import sirius.kernel.Sirius;
 import sirius.kernel.async.CallContext;
 import sirius.kernel.async.Promise;
 import sirius.kernel.async.TaskContext;
@@ -66,6 +65,9 @@ public class ControllerDispatcher implements WebDispatcher {
 
     @ConfigValue("http.skipCSRFTokens")
     private static boolean skipCsrfTokens;
+
+    @ConfigValue("http.csrfExemptions")
+    private static Set<String> csrfExemptions;
 
     /**
      * Used to log general controller related activities.
@@ -505,6 +507,6 @@ public class ControllerDispatcher implements WebDispatcher {
     }
 
     private boolean isExemptFromCsrfValidation(Route route) {
-        return Sirius.getSettings().getStringList("http.csrfExemptions").contains(route.getUri());
+        return csrfExemptions.contains(route.getUri());
     }
 }
--- a/src/main/java/sirius/web/controller/ControllerDispatcher.java
+++ b/src/main/java/sirius/web/controller/ControllerDispatcher.java
@@ -11,6 +11,7 @@ package sirius.web.controller;
 import io.netty.handler.codec.http.HttpHeaderNames;
 import io.netty.handler.codec.http.HttpMethod;
 import io.netty.handler.codec.http.HttpResponseStatus;
+import sirius.kernel.Sirius;
 import sirius.kernel.async.CallContext;
 import sirius.kernel.async.Promise;
 import sirius.kernel.async.TaskContext;
@@ -29,6 +30,7 @@ import sirius.kernel.health.Exceptions;
 import sirius.kernel.health.HandledException;
 import sirius.kernel.health.Log;
 import sirius.kernel.xml.StructuredOutput;
+import sirius.web.http.CSRFHelper;
 import sirius.web.http.Firewall;
 import sirius.web.http.InputStreamHandler;
 import sirius.web.http.Limited;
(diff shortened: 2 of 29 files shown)


src/main/java/sirius/web/controller/BasicController.java
```diff
@@ -265,6 +248,11 @@ public void onApiError(WebContext webContext, HandledException error, Format for
         out.endResult();
     }
 
+    @Override
+    public boolean isSkipCsrfValidation() {
+        return false;
+    }
+
```
sabieber: **Optional:** Why not use a default implementation in the `Controller` interface for this?

src/main/java/sirius/web/controller/ControllerDispatcher.java
```diff
@@ -469,4 +485,27 @@ private Route compileMethod(Routed routed, final Controller controller, final Me
...
+                            .handle();
+        }
+    }
+
+    private boolean isCsrfValidatedMethod(WebContext webContext) {
+        return CSRF_VALIDATED_METHODS.contains(webContext.getRequest().method());
+    }
+
+    private boolean isExemptFromCsrfValidation(Route route) {
+        return Sirius.getSettings().getStringList("http.csrfExemptions").contains(route.getRawRoute());
```
sabieber: **Optional:** Should we bind this to a variable after startup so we don't have to resolve it for each call?

src/main/java/sirius/web/controller/BasicController.java
fhaScireum: Because this was the way to go so far: `Controller` would declare methods, `BasicController` would provide the actual default implementation.

sabieber: Yes, because `Controller` pre-dates the introduction of default methods in interfaces I think

sabieber (approved):

mkeckmkeck (approved):

jmuscireum (approved):

src/main/java/sirius/web/controller/Routed.java
```diff
@@ -96,8 +96,7 @@
      * @return <tt>true</tt> if the method is used to create a JSON response for an AJAX call, <tt>false</tt> otherwise
      * @deprecated Use {@link InternalService} or {@link PublicService}
      */
-    @Deprecated(since = "2021/07/01")
-    boolean jsonCall() default false;
+    @Deprecated(since = "2021/07/01") boolean jsonCall() default false;
```
jakobvogel, line 99: Sollten wir das mal wegwerfen? Sind ja eh schon super-breaking.

jakobvogel (approved):

idlira (approved):

src/main/java/sirius/web/controller/Routed.java
fhaScireum: Ich denke auch, wir sollten das mal entfernen. Nutzen es noch im StorageController sowie in den Produkten. Aber das glatt zu ziehen ist ja sehr einfach, einfach stattdessen die `@InternalService` Annotation angeben.

fhaScireum: Ich habe Ticket SIRI-1226 angelegt. Wenn noch Luft ist könnte man das noch reinnehmen. Ich spreche es morgen an.

idlira (approved):

jakobvogel (approved):

PR #1662, opened 2026-06-24
Add AES-256-GCM client session encryption

bkositza:
### Breaking Changes

Client session cookies are now encrypted by default (`http.sessionCookie.encrypt = true`). As a consequence, a fixed `http.sessionSecret` is mandatory: if encryption is enabled and no secret is configured, the application fails on startup (previously an empty secret silently generated a random one each start).

Before deploying, make sure a stable, sufficiently long and random `http.sessionSecret` is configured (the same value across all cluster nodes).

### Description

Introduces client session cookie encryption and secret rotation.

ClientSessionCrypto encrypts and decrypts the client session cookie payload using AES-256 in GCM mode. GCM is an authenticated encryption mode, so it provides confidentiality and integrity in one step.

The symmetric key is derived from the session secret via SHA-256. A fresh random IV is generated per encryption and prepended to the cipher text, so encrypting the same payload twice yields different cookies.
Encrypted values are marked with an E1: prefix so the reading side can distinguish them from legacy (unencrypted) cookies and support both during migration.
decrypt() returns null on any failure (wrong secret, tampering, malformed input) so the caller can simply reset the session instead of failing the request.
ClientSessionSecrets manages the secrets used to sign and encrypt sessions. It resolves all accepted secrets on startup, verifies that a fixed secret is present when encryption is enabled (failing startup otherwise), and exposes the primary secret plus all configured legacy secrets.

Secret rotation is supported via http.legacySessionSecrets: old secrets are still accepted when reading, but never used for writing. This allows rotating http.sessionSecret without logging everyone out — add the old secret to the legacy list, set a new primary secret, and remove the old one once all cookies signed/encrypted with it have expired.

WebContext wires this in: it writes sessions in the encrypted format, transparently reads both encrypted and legacy cookies (trying the primary and all legacy secrets), and eagerly upgrades a legacy unencrypted cookie to the encrypted format on read, instead of waiting for the next session modification.

Configuration: new `http.sessionCookie.encrypt` (default `true`) and `http.legacySessionSecrets` (default `[]`).

### Additional Notes

- This PR fixes or works on following ticket(s): [SIRI-1208](https://scireum.myjetbrains.com/youtrack/issue/SIRI-1208)
- This PR is related to PR: 

### Checklist

### DIFF
--- a/src/main/java/sirius/web/http/ClientSessionCrypto.java
+++ b/src/main/java/sirius/web/http/ClientSessionCrypto.java
@@ -34,6 +34,8 @@ import java.util.Base64;
  * <p>
  * Encrypted values are marked with {@link #ENCRYPTION_PREFIX} so that the reading side can distinguish them from
  * legacy (unencrypted) cookies and transparently support both formats during a migration period.
+ *
+ * @see ClientSessionSecrets for the management of the secrets used here
  */
 class ClientSessionCrypto {
 
--- a/src/main/java/sirius/web/http/ClientSessionCrypto.java
+++ b/src/main/java/sirius/web/http/ClientSessionCrypto.java
@@ -1,0 +1,138 @@
+/*
+ * Made with all the love in the world
+ * by scireum in Stuttgart, Germany
+ *
+ * Copyright by scireum GmbH
+ * https://www.scireum.de - info@scireum.de
+ */
+
+package sirius.web.http;
+
+import sirius.kernel.commons.Hasher;
+import sirius.kernel.commons.Strings;
+import sirius.kernel.health.Exceptions;
+
+import javax.annotation.Nullable;
+import javax.crypto.Cipher;
+import javax.crypto.spec.GCMParameterSpec;
+import javax.crypto.spec.SecretKeySpec;
+import java.nio.charset.StandardCharsets;
+import java.security.SecureRandom;
+import java.util.Arrays;
+import java.util.Base64;
+
+/**
+ * Encrypts and decrypts the payload stored within the client session cookie.
+ * <p>
+ * The client session is stored on the client side within a cookie. In order to keep its contents confidential
+ * (and not just tamper-proof), the payload is encrypted using <b>AES-256 in GCM mode</b>. GCM is an authenticated
+ * encryption mode and therefore also guarantees the integrity of the encrypted data.
+ * <p>
+ * The symmetric key is derived from the configured session secret (see {@code http.sessionSecret}) by hashing it
+ * with SHA-256. A fresh random initialization vector (IV) is generated for every encryption and prepended to the
+ * cipher text so that encrypting the same payload twice yields different cookies.
+ * <p>
+ * Encrypted values are marked with {@link #ENCRYPTION_PREFIX} so that the reading side can distinguish them from
+ * legacy (unencrypted) cookies and transparently support both formats during a migration period.
+ */
+class ClientSessionCrypto {
+
+    /**
+     * Marks a cookie value as being encrypted using version 1 of this scheme.
+     * <p>
+     * Legacy cookies start with a hex encoded SHA-512 hash and therefore never collide with this prefix.
+     */
+    static final String ENCRYPTION_PREFIX = "E1:";
+
+    private static final String KEY_ALGORITHM = "AES";
+    private static final String CIPHER_TRANSFORMATION = "AES/GCM/NoPadding";
+    private static final int GCM_IV_LENGTH = 12;
+    private static final int GCM_TAG_LENGTH_BITS = 128;
+
+    private static final SecureRandom SECURE_RANDOM = new SecureRandom();
+
+    private ClientSessionCrypto() {
+    }
+
+    /**
+     * Determines if the given cookie value is an encrypted payload created by {@link #encrypt(String, String)}.
+     *
+     * @param cookieValue the raw cookie value to inspect
+     * @return <tt>true</tt> if the value is encrypted, <tt>false</tt> if it is a legacy plain text value
+     */
+    static boolean isEncrypted(@Nullable String cookieValue) {
+        return cookieValue != null && cookieValue.startsWith(ENCRYPTION_PREFIX);
+    }
+
+    /**
+     * Encrypts the given plain text using a key derived from the given secret.
+     *
+     * @param plainText the payload to encrypt
+     * @param secret    the secret used to derive the symmetric key
+     * @return the encrypted payload, marked with {@link #ENCRYPTION_PREFIX} and Base64 encoded
+     */
+    static String encrypt(String plainText, String secret) {
+        try {
+            byte[] iv = new byte[GCM_IV_LENGTH];
+            SECURE_RANDOM.nextBytes(iv);
+
+            Cipher cipher = Cipher.getInstance(CIPHER_TRANSFORMATION);
+            cipher.init(Cipher.ENCRYPT_MODE, deriveKey(secret), new GCMParameterSpec(GCM_TAG_LENGTH_BITS, iv));
+            byte[] cipherText = cipher.doFinal(plainText.getBytes(StandardCharsets.UTF_8));
+
+            byte[] combined = new byte[iv.length + cipherText.length];
+            System.arraycopy(iv, 0, combined, 0, iv.length);
+            System.arraycopy(cipherText, 0, combined, iv.length, cipherText.length);
+
+            return ENCRYPTION_PREFIX + Base64.getEncoder().encodeToString(combined);
+        } catch (Exception exception) {
+            throw Exceptions.handle()
+                            .to(WebServer.LOG)
+                            .error(exception)
+                            .withSystemErrorMessage("Failed to encrypt the client session: %s (%s)")
+                            .handle();
+        }
+    }
+
+    /**
+     * Decrypts a payload previously created by {@link #encrypt(String, String)}.
+     * <p>
+     * If the value cannot be decrypted (e.g. because it has been tampered with or was encrypted using a different
+     * secret), <tt>null</tt> is returned so that the caller can reset the session instead of failing the request.
+     *
+     * @param encryptedValue the encrypted payload, including the {@link #ENCRYPTION_PREFIX}
+     * @param secret         the secret used to derive the symmetric key
+     * @return the decrypted plain text or <tt>null</tt> if decryption failed
+     */
+    @Nullable
+    static String decrypt(String encryptedValue, String secret) {
+        if (!isEncrypted(encryptedValue)) {
+            return null;
+        }
+        try {
+            byte[] combined = Base64.getDecoder().decode(encryptedValue.substring(ENCRYPTION_PREFIX.length()));
+            if (combined.length <= GCM_IV_LENGTH) {
+                return null;
+            }
+
+            byte[] iv = Arrays.copyOfRange(combined, 0, GCM_IV_LENGTH);
+            byte[] cipherText = Arrays.copyOfRange(combined, GCM_IV_LENGTH, combined.length);
+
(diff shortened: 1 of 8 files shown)


src/test/kotlin/sirius/web/http/WebContextTest.kt
```diff
@@ -119,13 +119,29 @@ class WebContextTest {
...
-        val connection = URI("http://localhost:9999/test/session-test").toURL().openConnection() as HttpURLConnection
-        connection.setRequestMethod("GET")
-        connection.connect()
-
-        assertEquals(200, connection.responseCode)
-        assertTrue { connection.headerFields[HttpHeaderNames.SET_COOKIE.toString()]!![0].contains("test1=test") }
-        assertFalse { connection.headerFields[HttpHeaderNames.SET_COOKIE.toString()]!![0].contains("test2=") }
+        // The first request stores the session values and returns the (encrypted) session cookie.
+        val writeConnection =
+            URI("http://localhost:9999/test/session-test").toURL().openConnection() as HttpURLConnection
```
sabieber, line 124: **Question:** Is client session encryption really enabled in the tests? I don't see it configured in the `test.conf`

src/main/java/sirius/web/http/ClientSessionSecrets.java
```diff
@@ -0,0 +1,109 @@
...
+
+    /**
+     * Returns all secrets which are accepted when reading a client session.
+     * <p>
+     * The primary secret (used for writing) comes first, followed by all configured legacy secrets in the given
+     * order. This is used to support secret rotation: a cookie written with a previous secret can still be read.
+     *
+     * @return the list of accepted secrets, the primary one first
+     */
+    public List<String> getAllSessionSecrets() {
```
fhaScireum, line 107: Optional: Store result in field as it can never change without a system restart.

src/main/java/sirius/web/http/WebContext.java
```diff
@@ -944,39 +947,90 @@ private void installSessionPinningCookieIfRequired() {
...
+                String payload = ClientSessionCrypto.decrypt(encodedSession, secret);
+                if (payload != null) {
+                    Map<String, String> decodedSession =
+                            parseSessionPayload(payload, Collections.singletonList(secret));
+                    if (decodedSession != null) {
+                        return decodedSession;
+                    }
+                }
+            }
+        } else {
```
fhaScireum, line 976: If cookie starts with `E...:`, shouldn't we always throw an exception here instead of silently dropping the if- branch? We know the cookie is encrypted, so parsing it as a non-encrypted cookie is sure to fail.

src/test/kotlin/sirius/web/http/WebContextTest.kt
bkositza: It is enabled by default, the secret is set in the component-test-web.conf

src/main/java/sirius/web/http/WebContext.java
bkositza: We must continue to support the scenario where a cookie is not encrypted. For example, if we set the encryption to ‘false’ for a certain period of time. Or even now, whilst all existing cookies are still unencrypted.

If a cookie starts with "E1:" ClientSessionCrypto.isEncrypted will be true otherwise it is not encrypted and we should more or less handle it the old way

src/main/java/sirius/web/http/ClientSessionSecrets.java
bkositza: I don’t think I quite understand what you mean. We want to have a list of old secrets so that we can still decrypt older cookies – even after a rotation. Of course, we only store the old cookies for a certain period of time. At some point, the older secrets are removed from the list again.

src/test/kotlin/sirius/web/http/WebContextTest.kt
sabieber: Okay, don't really know why we split up the test config that way but not a topic for this PR :)

src/main/java/sirius/web/http/ClientSessionSecrets.java
sabieber: He just means that it makes no sense to compute the list every time the method is called. It could be calculated at system startup and stored in a class-level field

src/main/resources/component-065-web.conf
```diff
@@ -164,14 +164,28 @@ http {
         # You probably want this, especially if you do not expect unsecure http connections without SSL.
         # Possible values are: ALWAYS_SECURE, NEVER, IF_SSL
         secure = "IF_SSL"
+
+        # Determines whether the client session payload is encrypted (AES-256/GCM) within the cookie instead of
+        # being stored as plain text. The encryption key is derived from #sessionSecret, therefore a fixed secret
+        # should be configured - otherwise sessions cannot be decrypted across a cluster or a server restart.
+        # When reading, legacy unencrypted cookies are still accepted, so this can be enabled without logging
+        # everyone out. Disable only for debugging or to temporarily roll back the behaviour.
+        encrypt = true
```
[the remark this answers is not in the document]
sabieber: > and also conflicts with the comment below which says a random secret is created on startup if none is present

Documentation should be updated. But disregarding that I think the setup is fine.

src/main/java/sirius/web/http/ClientSessionCrypto.java
```diff
@@ -0,0 +1,140 @@
...
+
+    /**
+     * Marks a cookie value as being encrypted using version 1 of this scheme.
+     * <p>
+     * Legacy cookies start with a hex encoded SHA-512 hash and therefore never collide with this prefix.
+     */
+    static final String ENCRYPTION_PREFIX = "E1:";
+
+    private static final String KEY_ALGORITHM = "AES";
+    private static final String CIPHER_TRANSFORMATION = "AES/GCM/NoPadding";
```
sabieber, line 50: **Optional:** There are no general constants we can reuse here, right?

src/main/java/sirius/web/http/WebContext.java
sabieber: > Or even now, whilst all existing cookies are still unencrypted.

Isn't it okay for all users to have to re-login to secure their sessions?

src/test/java/sirius/web/http/ClientSessionCryptoTest.java
```diff
@@ -0,0 +1,72 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
+ * Provides tests for the {@link ClientSessionCrypto} used to protect the client session cookie.
+ */
+class ClientSessionCryptoTest {
```
sabieber, line 22: **Optional:** No Kotlin 😿 But its fine

sabieber (approved):

idlira (approved):

src/main/java/sirius/web/http/ClientSessionSecrets.java
bkositza: That makes sense, I’ve done it

src/main/java/sirius/web/http/WebContext.java
bkositza: I think it would be better if we avoided mass logouts. However, I’ve added that if an unencrypted cookie is read, we should upgrade it straight away to an encrypted version.

src/main/java/sirius/web/http/ClientSessionCrypto.java
bkositza: Unfortunately, no

src/main/java/sirius/web/http/ClientSessionSecrets.java
```diff
@@ -95,15 +96,19 @@ public String requireSessionSecret() {
...
      */
     public List<String> getAllSessionSecrets() {
-        List<String> secrets = new ArrayList<>();
-        secrets.add(requireSessionSecret());
-        if (legacySessionSecrets != null) {
-            for (String legacySecret : legacySessionSecrets) {
-                if (Strings.isFilled(legacySecret)) {
-                    secrets.add(legacySecret);
+        if (effectiveSessionSecrets == null) {
+            List<String> secrets = new ArrayList<>();
```
idlira: **Optional**: no need to create a temporary secrets and copy it over. Just initialize effectiveSessionSecrets with a new ArrayList and fill it up would do the job.
**Question**: this method can be called in multiple threads right? If so, consider initializing it once in the started() method.

idlira (approved):

src/main/java/sirius/web/http/ClientSessionSecrets.java
bkositza: That’s a good idea – if we do that on startup, we can simplify the method a bit too.

sabieber: This can now actually have a race condition where another thread accesses the list in the middle if initialization. Unlikely, but could cause undefined behaviour.

I think Getter and Initializer should be separated.

fhaScireum (approved):

src/main/java/sirius/web/http/ClientSessionSecrets.java
```diff
@@ -97,16 +98,13 @@ public String requireSessionSecret() {
...
     public List<String> getAllSessionSecrets() {
         if (effectiveSessionSecrets == null) {
-            List<String> secrets = new ArrayList<>();
-            secrets.add(requireSessionSecret());
-            if (legacySessionSecrets != null) {
-                for (String legacySecret : legacySessionSecrets) {
-                    if (Strings.isFilled(legacySecret)) {
-                        secrets.add(legacySecret);
-                    }
+            effectiveSessionSecrets = new ArrayList<>();
```
idlira: **Optional suggestion**: initializing on start actually requires no lazy load. The list can be defined empty and initialized on start. After initialization, a boolean can be set to indicate its fully initialized.
On the other hand, getAllSessionSecrets should enter a small loop until initialization is complete, so callers will have no race-condition, and simply return the unmodified version of the list.

idlira (approved):

src/main/java/sirius/web/http/ClientSessionSecrets.java
bkositza: Is that what you were thinking of? Check the last commit https://github.com/scireum/sirius-web/pull/1662/changes/5d4f7492866b98afa288694bd8de79b19631e2fc

sabieber (approved):

idlira (approved):

ymo-sci (commented):
Wie ist die Ausgabe in `/templates/system/info.html.pasta` jetzt?

bkositza:
> Wie ist die Ausgabe in `/templates/system/info.html.pasta` jetzt?

Guter Hinweis, zeigt den Verschlüsselten Text des Cookies (unproblemantisch) aber leider entschlüsselt es dort auch das Cookie und zeigt dann die Werte im Klartext. Das ist noch unschön gewesen. Ich hab diesen Punkt daher aus dem Template entfernt.

sabieber (approved):

mkeckmkeck (approved):
Wird das verschlüsselte Cookie noch angezeigt? Das wäre vll nicht schlecht. Dann könnte man das als supporter kopieren und auf einem abgesichterten Endpunkt decrypten lassen

bkositza:
> Wird das verschlüsselte Cookie noch angezeigt? Das wäre vll nicht schlecht. Dann könnte man das als supporter kopieren und auf einem abgesichterten Endpunkt decrypten lassen

Jap, das Cookie wird angezeigt mit verschlüsseltem Inhalt.

PR #1668, opened 2026-07-02
Add system tool to decode client session cookies

bkositza:
### Description

Adds a system administrator page at /system/session-cookie that decrypts a submitted session cookie value and displays its contents in a table. Guarded by permission-system-console.

image

### Additional Notes

- This PR fixes or works on following ticket(s): [SIRI-1208](https://scireum.myjetbrains.com/youtrack/issue/SIRI-1208)
- https://github.com/scireum/sirius-biz/pull/2338

### Checklist

### DIFF
--- a/src/main/java/sirius/web/http/SessionCookieController.java
+++ b/src/main/java/sirius/web/http/SessionCookieController.java
@@ -1,0 +1,107 @@
+/*
+ * Made with all the love in the world
+ * by scireum in Stuttgart, Germany
+ *
+ * Copyright by scireum GmbH
+ * https://www.scireum.de - info@scireum.de
+ */
+
+package sirius.web.http;
+
+import io.netty.handler.codec.http.QueryStringDecoder;
+import sirius.kernel.commons.Hasher;
+import sirius.kernel.commons.Strings;
+import sirius.kernel.commons.Tuple;
+import sirius.kernel.commons.Values;
+import sirius.kernel.di.std.Part;
+import sirius.kernel.di.std.Register;
+import sirius.web.controller.BasicController;
+import sirius.web.controller.Routed;
+import sirius.web.security.Permission;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Provides a system tool which decodes the contents of a client session cookie for diagnostic purposes.
+ * <p>
+ * The client session is stored on the client side within an encrypted cookie (see {@link ClientSessionCrypto}). To
+ * support debugging, this controller accepts the raw cookie value, decrypts it using the configured session secrets
+ * (see {@link ClientSessionSecrets}) and displays the contained key/value pairs in a table.
+ */
+@Register
+public class SessionCookieController extends BasicController {
+
+    private static final String TEMPLATE = "/templates/system/session-cookie.html.pasta";
+
+    @Part
+    private ClientSessionSecrets sessionSecrets;
+
+    /**
+     * Renders the tool and, on a POST request, the decoded contents of the submitted cookie value.
+     *
+     * @param webContext the current request
+     */
+    @Routed("/system/session-cookie")
+    @Permission("permission-system-console")
+    public void sessionCookie(WebContext webContext) {
+        String cookieValue = webContext.get("cookieValue").asString();
+        if (!webContext.isPostRequest() || Strings.isEmpty(cookieValue)) {
+            webContext.respondWith().template(TEMPLATE, cookieValue, false, false, false, null);
+            return;
+        }
+
+        String rawValue = cookieValue.trim();
+        boolean encrypted = ClientSessionCrypto.isEncrypted(rawValue);
+        String readablePayload = null;
+        boolean integrityValid = false;
+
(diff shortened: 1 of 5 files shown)


src/main/resources/default/templates/system/session-cookie.html.pasta
```diff
@@ -0,0 +1,80 @@
+<i:arg type="String" name="cookieValue" default=""/>
+<i:arg type="boolean" name="submitted" default="false"/>
+<i:arg type="boolean" name="encrypted" default="false"/>
+<i:arg type="boolean" name="integrityValid" default="false"/>
+<i:arg type="java.util.List" name="entries" default="null"/>
```
fhaScireum: We don't need default values here

fhaScireum (approved):

ymo-sci (commented):
should we link it somewhere in the menu/other page with permission?

bkositza:
> should we link it somewhere in the menu/other page with permission?

Do you have any specific requests regarding where the menu item should go and what it should be called? Personally, I don’t need a menu item, as I’m sure it won’t be used very often, but if there’s a need for it, I can include it.

mkeckmkeck (approved):

ymo-sci:
> > should we link it somewhere in the menu/other page with permission?
> 
> Do you have any specific requests regarding where the menu item should go and what it should be called? Personally, I don’t need a menu item, as I’m sure it won’t be used very often, but if there’s a need for it, I can include it.

Not sure, but i think /system/info below cookies at the bottom would be the best place, cause here the encoded value is displayed

src/main/java/sirius/web/http/SessionCookieController.java
```diff
@@ -0,0 +1,107 @@
...
+    @Part
+    private ClientSessionSecrets sessionSecrets;
+
+    /**
+     * Renders the tool and, on a POST request, the decoded contents of the submitted cookie value.
+     *
+     * @param webContext the current request
+     */
+    @Routed("/system/session-cookie")
+    @Permission("permission-system-console")
```
sabieber: **Optional:** I don't like the permission re-use

sabieber (approved):

src/main/java/sirius/web/http/SessionCookieController.java
idlira: **Question**: why this specific permission instead of system administrato?

idlira (approved):
Without a menu, we will forget this route even exists :-)

src/main/java/sirius/web/http/SessionCookieController.java
bkositza: changed it to system-administrator

bkositza: Changed it to system-administrator

bkositza:
> > > should we link it somewhere in the menu/other page with permission?
> > 
> > 
> > Do you have any specific requests regarding where the menu item should go and what it should be called? Personally, I don’t need a menu item, as I’m sure it won’t be used very often, but if there’s a need for it, I can include it.
> 
> Not sure, but i think /system/info below cookies at the bottom would be the best place, cause here the encoded value is displayed

I grouped client-info and session-cookie decryption tool into a new section that uses a sidebar, this way, we can easily access the page via system/info

image

idlira (approved):

src/main/java/sirius/web/http/SessionCookieController.java
sabieber: This breaks the separation of roles and permissions that was introduced in biz

idlira (approved):

src/main/java/sirius/web/http/SessionCookieController.java
```diff
@@ -0,0 +1,107 @@
...
+    @Part
+    private ClientSessionSecrets sessionSecrets;
+
+    /**
+     * Renders the tool and, on a POST request, the decoded contents of the submitted cookie value.
+     *
+     * @param webContext the current request
+     */
+    @Routed("/system/session-cookie")
+    @Permission("flag-system-administrator")
```
sabieber: As stated above user roles, tenant features and flags should not be used directly for permission checks but rather be used to grant specific permissions via profiles.
See also here: https://github.com/scireum/sirius-biz/blob/main/src/main/resources/component-070-biz.conf#L1677

This leaves much greater control for the systems using sirius (as permissions can also be revoked via other profiles)

src/main/java/sirius/web/http/SessionCookieController.java
idlira: True... I will not mention again that we should have one single sirius lib 😇

sabieber: You can, but in the sirius Kränzchen :) I am supporting that request ;)

But this does not block introducing a ner permission for the route and defining a profile in the config here in web

sabieber (approved):

idlira (approved):

jmuscireum (approved):

jakobvogel (approved):

PR #1680, opened 2026-07-14
Adds Partitioned (CHIPS) support for the session cookie

jakobvogel:
### Description

Adds an opt-in `http.sessionCookie.partitioned` (CHIPS) config that marks the client session cookie **and its session-pin cookie** with the `Partitioned` attribute. A partitioned cookie is stored and sent under a per-top-level-site partition, which lets the client session work inside a **cross-site (third-party) iframe** — e.g. an embedded support widget — that browsers (Safari ITP, Firefox TCP, Chrome 3PCD) would otherwise block or partition away.

- **Opt-in, defaults to `false`** → behaviour is unchanged for every existing product unless it enables the flag. When `false`, `setPartitioned(false)` is a genuine no-op (the encoder only emits `; Partitioned` when the flag is set), so no cookie changes on the wire.
- Only honored by browsers on a `Secure` cookie; for the cross-site iframe use case it additionally has to be combined with `SameSite=None`.
- Writing of the session and pin cookies is now funnelled through a single `setSessionScopedCookie` helper, so both consistently carry the SameSite, Secure and Partitioned attributes (the pin cookie must share the session cookie's partition, or pinning would break when CHIPS is enabled).

No breaking changes. Operational note when enabling: a partitioned session is scoped per top-level site (a session established under one embedding site is not shared with another) — this is the intended CHIPS behaviour.

### Additional Notes

- This PR fixes or works on following ticket(s): [MIO-6543](https://scireum.myjetbrains.com/youtrack/issue/MIO-6543) (memoio support widget needs its session cookie to survive a cross-site iframe)
- This PR is related to PR: 

### Checklist

- [ ] Patch Tasks: Is local execution of Patch Tasks necessary?

### DIFF
--- a/src/main/java/sirius/web/http/WebContext.java
+++ b/src/main/java/sirius/web/http/WebContext.java
@@ -360,6 +360,17 @@ public class WebContext implements SubContext {
     private static boolean sessionCookieEncryption;
 
     /**
+     * Determines if the session cookie (and its pin cookie) is marked as
+     * <a href="https://developer.mozilla.org/en-US/docs/Web/Privacy/Guides/Privacy_sandbox/Partitioned_cookies">Partitioned</a>
+     * (CHIPS). A partitioned cookie is stored and sent under a per-top-level-site partition, which lets it work in
+     * cross-site (third-party) iframe contexts that would otherwise block or partition it away. Browsers only honor the
+     * attribute on a {@code Secure} cookie; and for the cross-site iframe use case it additionally has to be combined
+     * with {@code SameSite=None} (otherwise the cookie is not sent cross-site in the first place).
+     */
+    @ConfigValue("http.sessionCookie.partitioned")
+    private static boolean sessionCookiePartitioned;
+
+    /**
      * Determines the domain set for all cookies. If empty no domain will be set.
      * If a cookie's domain attribute is not set, the cookie is only applicable to the domain of the originating request, EXCLUDING all its subdomains.
      * (However in IE 9 and older versions, a cookie made for abc.com is also sent in requests to xyz.abc.com)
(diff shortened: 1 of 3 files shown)


src/main/java/sirius/web/http/WebContext.java
```diff
@@ -1563,16 +1570,29 @@ private void buildClientSessionCookie() {
...
+     * @param name          the cookie name
+     * @param value         the cookie value
+     * @param maxAgeSeconds the max age in seconds, or {@link Long#MIN_VALUE} for a browser-session cookie
+     */
+    private void setSessionScopedCookie(String name, String value, long maxAgeSeconds) {
+        DefaultCookie cookie =
+                createCookie(name, value, maxAgeSeconds, determineSessionCookieSameSite(), sessionCookieSecurity);
+        cookie.setHttpOnly(true);
+        cookie.setPartitioned(sessionCookiePartitioned);
+        setCookie(cookie);
```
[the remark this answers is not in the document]
jakobvogel: Good catch — addressed in a6ff917c. The `Partitioned` attribute is now only emitted when the resulting cookie is both `Secure` and `SameSite=None`; otherwise it is silently skipped, so enabling `http.sessionCookie.partitioned` in a non-secure or non-cross-site context can no longer produce a rejected cookie / break the session.

src/main/java/sirius/web/http/WebContext.java
```diff
@@ -1566,12 +1572,33 @@ private void buildClientSessionCookie() {
```
sabieber, line 1566: **Question:** This has no usage left now, right?

sabieber (approved):

fhaScireum (approved):

idlira (approved):
I would have done it the other way around, since this is needed in a HF: PR against main, later branch from your desired source tag and cherry-pick there, resolving any conflict that might arrive.
Another reason for that: since compile fail due to conflicts, test cases don't even run.

src/main/java/sirius/web/http/WebContext.java
jakobvogel: yes, line 1533

PR #1683, opened 2026-07-21
Supports pojo mapping api routes

mkeckmkeck:
### Description

Adds support for **mapped request/response payloads** in JSON services: a `@Routed` service method can now receive the request body as a POJO parameter and return a result object (or `Promise`/`Future`) which is serialized directly into the JSON response — no more manual `JSONStructuredOutput` plumbing:

```java
@InternalService
@Routed("/greet")
public GreetResult greet(WebContext webContext, GreetInput input) {
    return new GreetResult("Hello " + input.getName());
}
```

How it works:

- A JSON route without a leading JSONStructuredOutput parameter is treated as a mapped service. The request body POJO is detected via the body-only binding rule (the method declares exactly one parameter more than the route has path parameters) and is bound right after the WebContext.
- The new ApiPayloadCodec SPI performs the Jackson-based (de-)serialization. The result is serialized before the response is opened, so serialization failures still produce a proper JSON error envelope instead of an empty 200. Async services returning Promise/Future are supported; their serialization errors are routed through handleFailure as well.
- Invalid signatures fail fast at route compile time with explanatory messages: simple value types / collections / arrays as body, POJOs in path parameter positions and body POJOs on pre-dispatchable routes are all rejected on startup instead of failing per request.
- The /system/api explorer auto-documents the input/output POJOs of mapped services (via @Schema field annotations, including nested POJOs, collections and generic superclasses) as request/response body tables — unless explicit @RequestBody/@ApiResponse annotations are present, which take precedence.
- Each public API can now be exported as an OpenAPI 3.1 document via a download action on /system/api (/system/api/:name/openapi.json), deriving paths, parameters and components/schemas (with $refs, self-referential and generic types handled) from the same metadata.

No breaking changes: legacy streaming services (JSONStructuredOutput / XMLStructuredOutput as second parameter) keep working unchanged and keep their manually annotated documentation. XML mapped payloads are rejected at compile time until an XML codec exists.

The PR is organized in 8 self-contained commits (codec → route validation → dispatcher → tests → schema introspection → explorer docs → OpenAPI generator → download endpoint) — reviewing commit by commit is recommended.

Additional Notes

- This PR fixes or works on following ticket(s): SE-15156
- Currently only Json is supported. If required XML can be implemented as a separate codec later. 

Checklist

- [x] Code was formatted via IntelliJ and follows SonarLint & best practices

### DIFF
--- a/src/main/java/sirius/web/controller/ControllerDispatcher.java
+++ b/src/main/java/sirius/web/controller/ControllerDispatcher.java
@@ -23,6 +23,7 @@ import sirius.kernel.commons.Strings;
 import sirius.kernel.di.Injector;
 import sirius.kernel.di.std.ConfigValue;
 import sirius.kernel.di.std.Part;
+import sirius.kernel.di.std.Parts;
 import sirius.kernel.di.std.PriorityParts;
 import sirius.kernel.di.std.Register;
 import sirius.kernel.health.Exceptions;
@@ -40,6 +41,7 @@ import sirius.web.http.WebDispatcher;
 import sirius.web.security.MaintenanceInfo;
 import sirius.web.security.UserContext;
 import sirius.web.security.UserInfo;
+import sirius.web.services.ApiPayloadCodec;
 import sirius.web.services.Format;
 import sirius.web.templates.ClosedChannelHelper;
 
@@ -48,6 +50,7 @@ import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.time.Duration;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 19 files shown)


sabieber:
> Each public API can now be exported as an OpenAPI 3.1 document via a download action on /system/api (/system/api/:name/openapi.json), deriving paths, parameters and components/schemas (with $refs, self-referential and generic types handled) from the same metadata.

Does this mean that existing public APIs writing JSON directly provide an incomplete schema download?

mkeckmkeck:
> > Each public API can now be exported as an OpenAPI 3.1 document via a download action on /system/api (/system/api/:name/openapi.json), deriving paths, parameters and components/schemas (with $refs, self-referential and generic types handled) from the same metadata.
> 
> Does this mean that existing public APIs writing JSON directly provide an incomplete schema download?

No, the OpenAPI schema also gets filled up from the informations provided by the existing swagger annotations. Here are some example files: 
[health-openapi.json](https://github.com/user-attachments/files/30215138/health-openapi.json)
[cluster-openapi.json](https://github.com/user-attachments/files/30215140/cluster-openapi.json)
They look fine when put into an online editor like https://editor.swagger.io/

mkeckmkeck:
And here a screenshot of the download button: 
grafik

idlira (approved):
🤩

fhaScireum (approved):

sabieber (approved):
The OpenAiGenerator is a bit large, but feels like an acceptable tradeoff.

PR #1689, opened 2026-07-24
SIRI-1256 sass modern css support

ymo-sci:
### Description

Extends the Sirius SASS compiler (sirius-web) with modern CSS constructs that
previously failed with "Unexpected Token" or silently produced wrong CSS.

**Selectors**
- Attribute selectors on the parent reference: `&[hidden]`, `&[data-x="y"]`
- Chained/parametrized pseudo-classes on `&`: `&:hover:focus`,
  `&:not(.a):hover` (previously silently wrong or a parse error)

**At-rules**
- `@container` (incl. name, `and`/`or`/`not`, range syntax) and `@supports`
- Nesting: consecutive `@media` rules keep merging into `@media a and b`;
  any other nesting (e.g. `@container` inside `@media`) is emitted as properly
  nested blocks
- `container` is intentionally not a keyword, so the CSS property
  `container: …` keeps working

**Tests:** new SassTest cases (`testSelectors` extended, `testContainer`,
`testSupports`, `testConditionalNesting`); added `@ExtendWith(SiriusExtension)`
so `SassTest` runs in isolation. Existing `@media` behaviour is unchanged.
### Additional Notes

- This PR fixes or works on following ticket(s): [SIRI-1256](https://scireum.myjetbrains.com/youtrack/issue/SIRI-1256)
- This PR is related to PR: 

### Checklist


src/main/java/sirius/web/sass/Parser.java
```diff
@@ -405,9 +423,27 @@ private String readRawCondition() {
                     break;
                 }
             }
-            parts.add(tokenizer.consume().getSource());
+            String token = tokenizer.consume().getSource();
+            if (needsSpaceBetween(previous, token)) {
+                builder.append(' ');
+            }
```
fhaScireum: `needsSpaceBetween` keeps colons tight, which is right for `selector(:focus-visible)` but wrong for `selector(a :hover)` as it results in `selector(a:hover)`.

src/main/java/sirius/web/sass/Parser.java
```diff
@@ -353,8 +361,16 @@ private void parseConditionalGroupSelector(String keyword, Section result) {
...
+                String word = tokenizer.consume().getSource();
+                if (tokenizer.current().isSymbol("(") && CONDITION_FUNCTIONS.contains(word.toLowerCase())) {
+                    // A functional condition like "selector(:focus-visible)" or "style(--x: y)": the name is
+                    // glued directly to its argument (no separating space).
+                    tokenizer.consumeExpectedSymbol("(");
+                    result.addMediaQuery(new Value(word + readRawCondition()));
+                } else {
+                    // A bare word: a '@container' name or a connector like 'and' / 'or' / 'not' / 'only'.
+                    result.addMediaQuery(new Value(word));
+                }
```
fhaScireum: Problem: Wenn der Container so benannt wird wie eine `CONDITION_FUNCTIONS`, dann wird es fälschlicherweise nicht als Container-Name sondern als eine Condition erkannt (also an die Bedingung "gepappt":

Vorschlag der KI:
```
    Token wordToken = tokenizer.consume();
    String word = wordToken.getSource();
    if (isGluedTo(wordToken, tokenizer.current())) {
        // A functional condition like "selector(:focus-visible)" or "style(--x: y)": the name is
        // glued directly to its argument. As the tokenizer drops whitespace, we compare the source
        // positions to tell "style(...)" (a style query) from "style (...)" (a container named "style").
        tokenizer.consumeExpectedSymbol("(");
        result.addMediaQuery(new Value(word + readRawCondition()));
    } else {
        // A bare word: a '@container' name or a connector like 'and' / 'or' / 'not' / 'only'.
        result.addMediaQuery(new Value(word));
    }
}
```
Dazu die Hilfsmethode
```
/*
 * Determines if the given tokens were written without any separating whitespace. The tokenizer discards
 * whitespace, so "style(--x: y)" and "style (--x: y)" yield identical token streams - only the source
 * positions tell them apart.
 */
private static boolean isGluedTo(Token first, Token second) {
    return second.isSymbol("(")
           && second.getLine() == first.getLine()
           && second.getPos() == first.getPos() + first.getSource().length();
}
```

Das CONDITION_FUNCTIONS Set kann dann weg.

jmuscireum (approved):

fhaScireum (approved):

mkeckmkeck (approved):

src/main/java/sirius/web/sass/Parser.java
```diff
@@ -305,6 +338,114 @@ private void parseSectionSelector(boolean mediaQuery, Section result) {
...
+     * Reads a parenthesized condition verbatim (the opening '(' has already been consumed), balancing nested
+     * parentheses and reconstructing the original whitespace between tokens from their source positions (see
+     * {@link #areAdjacent}). This keeps significant spacing intact, e.g. "a:hover" vs "a :hover" (compound vs
+     * descendant selector) or "--x: y". Used for conditions which are not of the simple '(feature: value)' shape,
+     * e.g. the range syntax '(width > 400px)' or a functional condition's argument.
+     *
+     * @return the condition including the surrounding parentheses
+     */
+    private String readRawCondition() {
+        StringBuilder builder = new StringBuilder("(");
```
sabieber: **Optional:** Especially the brackets are used multiple times (i think stuff like `@`, too). Doesn't SonarQube report something there?

sabieber (approved):

src/main/java/sirius/web/sass/Parser.java
```diff
@@ -385,6 +526,14 @@ private List<String> parseSelector() {
                 parseFilterInSelector(builder);
                 parseOperatorInSelector(builder);
                 selector.add(builder.toString());
+            } else if (tokenizer.current().isSymbol("[")) {
```
jakobvogel, line 544: > 🤖 Automated review by Claude (Claude Code), posted under Jakob's account — the findings below are Claude's, not Jakob's own.

This branch produces `["&", "[hidden]"]` regardless of whether the `[` was glued to the `&`, so a descendant selector gets compiled as a compound one:

```scss
.a { & [hidden] { color: green; } }
```
* before: parse error `Unexpected Token: [`
* now: `.a[hidden] { … }` — no warning

(Verified against the merge base; `.a { [hidden] { … } }` correctly yields `.a [hidden]`, so it's only the `&`-form that diverges.)

Since input that used to be rejected now silently compiles to a different selector, this looks worth fixing before merge — and the machinery is already in this PR: gating the branch on `areAdjacent(previousToken, current)` and otherwise keeping it a separate descendant part should do it. A case in `selectors.scss` next to the `&[hidden]` ones would lock it in.

src/main/java/sirius/web/sass/Parser.java
```diff
@@ -305,6 +338,114 @@ private void parseSectionSelector(boolean mediaQuery, Section result) {
...
+    /**
+     * Reads a parenthesized condition verbatim (the opening '(' has already been consumed), balancing nested
+     * parentheses and reconstructing the original whitespace between tokens from their source positions (see
+     * {@link #areAdjacent}). This keeps significant spacing intact, e.g. "a:hover" vs "a :hover" (compound vs
+     * descendant selector) or "--x: y". Used for conditions which are not of the simple '(feature: value)' shape,
+     * e.g. the range syntax '(width > 400px)' or a functional condition's argument.
+     *
+     * @return the condition including the surrounding parentheses
+     */
+    private String readRawCondition() {
```
jakobvogel: > 🤖 Automated review by Claude (Claude Code), posted under Jakob's account — the findings below are Claude's, not Jakob's own.

Because this path is verbatim, variables aren't evaluated inside a raw condition:

```scss
$bp: 400px;
@container (width > $bp) { .a { color: red; } }
  ==> @container (width > $bp) { … }   // literal "$bp" reaches the CSS, no warning
```

`(min-width: $bp)` works fine via the `MediaFilter` branch, so the two shapes behave differently — and range syntax is the form this PR just made available. Resolving `VariableReference` tokens here, or at least warning when a `$` shows up in a raw condition, would avoid the silent broken output.

src/main/java/sirius/web/sass/Parser.java
```diff
@@ -45,6 +45,12 @@ public class Parser {
     private static final String KEYWORD_EXTEND = "extend";
     private static final String KEYWORD_MEDIA = "media";
 
+    // Additional conditional group at-rules. These are intentionally NOT registered as tokenizer keywords:
+    // "container" is also a regular CSS property (e.g. "container: name / inline-size;"), so registering it as a
```
jakobvogel: > 🤖 Automated review by Claude (Claude Code), posted under Jakob's account — the findings below are Claude's, not Jakob's own.

Small heads-up on the example cited here: the slash form doesn't actually survive the compiler, on this branch or on main:

```scss
.a { container: layout / inline-size; }
  ==> .a { container: layoutinline-size; }
```

That's pre-existing `/`-handling in expression parsing and clearly outside this PR's scope — but `container.scss` only covers the slash-free `container: layout`, so the comment currently promises a bit more than the tests show. Maybe reword to the shorthand-without-slash case (or add the slash form as a documented limitation) so the next reader isn't misled.

src/main/java/sirius/web/sass/Generator.java
```diff
@@ -275,31 +279,41 @@ private void transferImplicitAttributes(String mediaQueryPath, Section section,
...
+     * single "@media a and b" segment (the established SASS shortcut, kept for backwards compatibility); every
+     * other rule - and every @media following a non-@media rule - starts a new nesting level, so that mixed
+     * nesting (e.g. @container inside @media) is emitted as nested blocks as required by CSS.
+     */
+    private List<String> expandConditionPath(List<String> conditionPath, Section section) {
+        String keyword = section.getConditionKeyword() != null ? section.getConditionKeyword() : "media";
+        String query = section.getMediaQuery(scope, this);
+        List<String> result = conditionPath == null ? new ArrayList<>() : new ArrayList<>(conditionPath);
+
+        if ("media".equals(keyword) && !result.isEmpty() && result.getLast().startsWith("@media ")) {
```
jakobvogel: > 🤖 Automated review by Claude (Claude Code), posted under Jakob's account — the findings below are Claude's, not Jakob's own.

Nit, no behavioural problem found: the merge decision re-derives the keyword by sniffing the rendered prefix (`startsWith("@media ")`), and `"media"` now appears as a literal here and in `Section.getMediaQuery` while `Parser` keeps it as `KEYWORD_MEDIA`. Keeping the path as segments of `(keyword, query)` instead of pre-rendered strings would drop the prefix check and the duplicated literal. Fine to leave as is if you'd rather keep the diff small.

jakobvogel (commented):
> 🤖 **Automated review by Claude (Claude Code)**, posted under Jakob's account. The analysis and all inline comments are Claude's, not Jakob's own.

Nice piece of work — the design calls look right to me: keeping `container` out of the tokenizer keywords, preserving the `@media` "and"-merging shortcut for backwards compatibility, deriving adjacency from source positions instead of a name list, and adding `@ExtendWith(SiriusExtension.class)` (`SassTest` was quietly relying on another test to boot the `Resources` part).

**What I verified**

* `SassTest` passes, 19/19.
* Compiled all 62 `.scss` files under `src/main/resources` with this branch and with the merge base, diffing full output plus warnings — **byte-identical**. No drift on existing stylesheets.
* The headline fix is real: `&:hover:focus` used to compile to `.a:hover :focus`, now `.a:hover:focus`.
* The new fixtures match the generated output exactly (worth knowing: `compare()` only iterates *expected* lines, so extra trailing output would go unnoticed — there is none here).

**Findings** (details inline)

1. `& [attr]` with a space now compiles to `.a[attr]` instead of a descendant selector, silently. Previously a parse error. This is the one I'd suggest fixing before merge — small change, and `areAdjacent` from this PR already does the work.
2. Variables aren't resolved inside range-syntax conditions: `@container (width > $bp)` emits `$bp` literally, while `(min-width: $bp)` resolves.
3. The `container: name / inline-size` example in the new comment doesn't survive the compiler (pre-existing `/` handling), and the test only covers the slash-free form.
4. Nit on the stringly-typed condition path in `Generator`.

Not blocking, just noting for possible follow-up tickets: `@media` still rejects range syntax, `or`/`not`, and comma-separated media lists, so `@container` is now ahead of it; `@container`/`@supports` inside `@mixin` fail to parse (as `@media` already does); empty rule blocks can still be emitted inside conditional groups; and an unterminated raw condition swallows the rest of the file before reporting `Expected: '{'` at EOF.

jakobvogel (approved):

jakobvogel (approved):
Hast Du selbst auch schon eine Claude Review laufen lassen?

mkeckmkeck (approved):

PR #1698, opened 2026-08-03
Make CORS `Access-Control-Allow-Origin` resolution extensible via `Interceptor`

hso-scireum:
### Description

This PR refactors CORS `Access-Control-Allow-Origin` handling out of Response into a dedicated, extensible mechanism, and lets Interceptor implementations decide which origins are allowed on a per-route basis instead of always mirroring the request's Origin header.

Previously, CORS origin handling was hardcoded in `Response` and only ever mirrored the request's `Origin` header when `http.firewall.corsAllowAll` was enabled - there was no way for a controller/route to have a different, narrower CORS policy. This introduces a pluggable strategy so interceptors can restrict allowed origins per route (e.g. an allow-list) without touching core response/dispatch code.

In controller-based routes dispatched via `ControllerDispatcher`, the CORS origin behavior can now be centrally configured via `Interceptor.determineAllowedCorsOrigin(WebContext, Collection<Route>)` and the `AllowedOrigin` abstraction. In all other dispatchers, the header can either be set manually or, recommended, via `CorsAllowOriginResolver`, which respects the CORS cofnguration and is used by `ControllerDisaptcher` as well.

> [!CAUTION]
> __Breaking config change affecting CORS handling:__
> The config key`http.corsAllowAll` has been renamed to `http.enableCors` to better reflect the updated behavior.
> If you previously enabled CORS handling via the flag, your yonfig will need to be adjusted or CORS handling will be disabled by default.

> [!WARNING]
> __Also breaking:__
> `WebContext.isCorsAllowAll()` is no longer `static` and has been renamed to `webContext.isCorsEnabled()`.

### Additional Notes

- This PR fixes or works on following ticket(s): [SIRI-1241](https://scireum.myjetbrains.com/youtrack/issue/SIRI-1241)

### Checklist

### DIFF
--- a/src/main/java/sirius/web/controller/ControllerDispatcher.java
+++ b/src/main/java/sirius/web/controller/ControllerDispatcher.java
@@ -31,7 +31,7 @@ import sirius.kernel.health.HandledException;
 import sirius.kernel.health.Log;
 import sirius.kernel.xml.StructuredOutput;
 import sirius.web.cors.AllowedOrigin;
-import sirius.web.cors.CorsAllowOriginHelper;
+import sirius.web.cors.CorsAllowOriginResolver;
 import sirius.web.http.CSRFHelper;
 import sirius.web.http.Firewall;
 import sirius.web.http.InputStreamHandler;
@@ -107,7 +107,7 @@ public class ControllerDispatcher implements WebDispatcher {
     private CSRFHelper csrfHelper;
 
     @Part
-    private CorsAllowOriginHelper corsOriginHelper;
+    private CorsAllowOriginResolver corsOriginResolver;
 
     @Part
     @Nullable
@@ -307,14 +307,14 @@ public class ControllerDispatcher implements WebDispatcher {
      * <p>
      * The stored origin can then be used in order to set the {@code Access-Control-Allow-Origin} header in the
      * response in the relevant places by using
-     * {@link CorsAllowOriginHelper#applyResolvedOriginHeader(Response)}.
+     * {@link CorsAllowOriginResolver#applyResolvedOriginHeader(Response)}.
      * </p>
      *
      * <p>
      * <b>Note:</b> The strategy is only resolved and stored when {@code webContext.isCorsEnabled()} is set for the
      * request's scope - in that case the interceptors decide based on the routes matching the URI (falling back to a
      * {@link AllowedOrigin.MatchRequest} reflecting the request's origin if none does). If it is not set,
-     * {@link CorsAllowOriginHelper#tryResolveAndStoreOrigin(WebContext, java.util.function.Supplier)} stores nothing
+     * {@link CorsAllowOriginResolver#tryResolveAndStoreOrigin(WebContext, java.util.function.Supplier)} stores nothing
      * (and does not even consult the interceptors), so no CORS headers are emitted.
      * </p>
      *
@@ -322,7 +322,7 @@ public class ControllerDispatcher implements WebDispatcher {
      * @param uri        the effective request URI
      */
     private void determineAndStoreAllowedOriginForUri(WebContext webContext, String uri) {
-        corsOriginHelper.tryResolveAndStoreOrigin(webContext, () -> {
+        corsOriginResolver.tryResolveAndStoreOrigin(webContext, () -> {
             Collection<Route> routes = collectMatchingRoutes(webContext, uri);
             return determineAllowedOriginForRoutes(webContext, routes);
         });
@@ -445,7 +445,7 @@ public class ControllerDispatcher implements WebDispatcher {
             response.setHeader(HttpHeaderNames.ACCESS_CONTROL_ALLOW_METHODS, allowedMethods)
                     .setHeader(HttpHeaderNames.ACCESS_CONTROL_ALLOW_HEADERS,
                                requestedHeaders == null ? "" : requestedHeaders);
-            corsOriginHelper.applyResolvedOriginHeader(response);
+            corsOriginResolver.applyResolvedOriginHeader(response);
         }
 
         response.status(HttpResponseStatus.OK);
--- a/src/main/java/sirius/web/controller/ControllerDispatcher.java
+++ b/src/main/java/sirius/web/controller/ControllerDispatcher.java
@@ -237,7 +237,14 @@ public class ControllerDispatcher implements WebDispatcher {
         // preferred allowed origin and store it in the WebContext.
         // It may then be used in the different relevant places: In this class in order to answer preflight requests
         // and OPTIONS requests, as well as the Response class to attach it to all remaining responses.
-        determineAndStoreAllowedOriginForUri(webContext, uri);
+        //
+        // The preferred origin is only attempted to resolve if the request is a preflight request or an `Origin` header
+        // has been sent by the client.
+        // No such header means the browser won't verify CORS for the given request, so we can omit the extra steps and
+        // iterations to resolve it.
+        if (isCorsPreflightRequest(webContext) || Strings.isFilled(webContext.getHeader(HttpHeaderNames.ORIGIN))) {
+            determineAndStoreAllowedOriginForUri(webContext, uri);
+        }
 
         // A genuine CORS preflight request must always be answered centrally with a method list derived from the
         // routes registered for this path. It must never reach any business logic - not even a controller which
@@ -300,47 +307,48 @@ public class ControllerDispatcher implements WebDispatcher {
      * <p>
      * The stored origin can then be used in order to set the {@code Access-Control-Allow-Origin} header in the
      * response in the relevant places by using
-     * {@link CorsAllowOriginHelper#applyHeaderFromWebContext(WebContext, Response)}.
+     * {@link CorsAllowOriginHelper#applyResolvedOriginHeader(Response)}.
      * </p>
      *
      * <p>
-     * <b>Note:</b> If {@code WebContext.isCorsAllowAll()} is set, a {@link AllowedOrigin.MatchRequest} is used.
-     * Otherwise, the interceptors decide based on the routes matching the URI.
+     * <b>Note:</b> The strategy is only resolved and stored when {@code webContext.isCorsAllowAll()} is set for the
+     * request's scope - in that case the interceptors decide based on the routes matching the URI (falling back to a
+     * {@link AllowedOrigin.MatchRequest} reflecting the request's origin if none does). If it is not set,
+     * {@link CorsAllowOriginHelper#tryResolveAndStoreOrigin(WebContext, java.util.function.Supplier)} stores nothing
+     * (and does not even consult the interceptors), so no CORS headers are emitted.
      * </p>
      *
      * @param webContext the current request
      * @param uri        the effective request URI
      */
     private void determineAndStoreAllowedOriginForUri(WebContext webContext, String uri) {
-        Collection<Route> routes = collectMatchingRoutes(webContext, uri);
-
-        Optional<AllowedOrigin> origin = WebContext.isCorsAllowAll() ?
-                                         Optional.of(new AllowedOrigin.MatchRequest()) :
-                                         determineAllowedOriginForRoutes(webContext, routes);
-
-        origin.ifPresentOrElse(actual -> corsOriginHelper.tryResolveOriginAndStoreInWebContext(webContext, actual),
-                               () -> LOG.WARN("CORS: No allowed origin could be determined for '%s'!", uri, routes));
+        corsOriginHelper.tryResolveAndStoreOrigin(webContext, () -> {
+            Collection<Route> routes = collectMatchingRoutes(webContext, uri);
+            return determineAllowedOriginForRoutes(webContext, routes);
+        });
     }
 
     /**
      * Asks the interceptors to determine the {@link AllowedOrigin} for the given routes.
      * <p>
      * The first interceptor returning a non-empty result via
-     * {@link Interceptor#determineAllowedCorsOrigin(WebContext, Collection)} wins.
+     * {@link Interceptor#determineAllowedCorsOrigin(WebContext, Collection)} wins. If no interceptor makes a decision,
+     * a {@link AllowedOrigin.MatchRequest} is used as fallback, reflecting the origin of the current request.
      * </p>
      *
      * @param webContext the current request
      * @param routes     the routes matching the requested URI
-     * @return the preferred CORS allowed origin, or an empty optional if none could be determined
+     * @return the CORS allowed origin as decided by the first interceptor making a decision, or a
+     * {@link AllowedOrigin.MatchRequest} if none does
      */
-    private Optional<AllowedOrigin> determineAllowedOriginForRoutes(WebContext webContext, Collection<Route> routes) {
+    private AllowedOrigin determineAllowedOriginForRoutes(WebContext webContext, Collection<Route> routes) {
         for (var interceptor : interceptors) {
             Optional<AllowedOrigin> origin = interceptor.determineAllowedCorsOrigin(webContext, routes);
             if (origin.isPresent()) {
-                return origin;
+                return origin.get();
             }
         }
-        return Optional.empty();
+        return new AllowedOrigin.MatchRequest();
     }
 
     /**
@@ -351,9 +359,7 @@ public class ControllerDispatcher implements WebDispatcher {
      * @return the routes matching the URI
      */
     private Collection<Route> collectMatchingRoutes(WebContext webContext, String uri) {
-        return getRoutes().stream()
-                          .filter(route -> route.matches(webContext, uri, route.isPreDispatchable()) != Route.NO_MATCH)
-                          .collect(Collectors.toSet());
+        return getRoutes().stream().filter(route -> route.matchesUri(uri)).collect(Collectors.toSet());
     }
 
     /**
@@ -439,7 +445,7 @@ public class ControllerDispatcher implements WebDispatcher {
             response.setHeader(HttpHeaderNames.ACCESS_CONTROL_ALLOW_METHODS, allowedMethods)
                     .setHeader(HttpHeaderNames.ACCESS_CONTROL_ALLOW_HEADERS,
                                requestedHeaders == null ? "" : requestedHeaders);
-            corsOriginHelper.applyHeaderFromWebContext(webContext, response);
+            corsOriginHelper.applyResolvedOriginHeader(response);
         }
 
         response.status(HttpResponseStatus.OK);
--- a/src/main/java/sirius/web/controller/ControllerDispatcher.java
+++ b/src/main/java/sirius/web/controller/ControllerDispatcher.java
@@ -311,7 +311,7 @@ public class ControllerDispatcher implements WebDispatcher {
      * </p>
      *
      * <p>
-     * <b>Note:</b> The strategy is only resolved and stored when {@code webContext.isCorsAllowAll()} is set for the
+     * <b>Note:</b> The strategy is only resolved and stored when {@code webContext.isCorsEnabled()} is set for the
      * request's scope - in that case the interceptors decide based on the routes matching the URI (falling back to a
      * {@link AllowedOrigin.MatchRequest} reflecting the request's origin if none does). If it is not set,
      * {@link CorsAllowOriginHelper#tryResolveAndStoreOrigin(WebContext, java.util.function.Supplier)} stores nothing
--- a/src/main/java/sirius/web/controller/ControllerDispatcher.java
+++ b/src/main/java/sirius/web/controller/ControllerDispatcher.java
@@ -30,6 +30,8 @@ import sirius.kernel.health.Exceptions;
 import sirius.kernel.health.HandledException;
 import sirius.kernel.health.Log;
 import sirius.kernel.xml.StructuredOutput;
+import sirius.web.cors.AllowedOrigin;
+import sirius.web.cors.CorsAllowOriginHelper;
 import sirius.web.http.CSRFHelper;
 import sirius.web.http.Firewall;
 import sirius.web.http.InputStreamHandler;
@@ -103,6 +105,9 @@ public class ControllerDispatcher implements WebDispatcher {
 
     @Part
     private CSRFHelper csrfHelper;
+
+    @Part
+    private CorsAllowOriginHelper corsOriginHelper;
 
     @Part
     @Nullable
@@ -228,6 +233,12 @@ public class ControllerDispatcher implements WebDispatcher {
     public DispatchDecision dispatch(WebContext webContext) throws Exception {
         String uri = determineEffectiveURI(webContext);
 
+        // In order to be able to send a CORS `Access-Control-Allow-Origin` header, we determine the
+        // preferred allowed origin and store it in the WebContext.
+        // It may then be used in the different relevant places: In this class in order to answer preflight requests
+        // and OPTIONS requests, as well as the Response class to attach it to all remaining responses.
+        determineAndStoreAllowedOriginForUri(webContext, uri);
+
         // A genuine CORS preflight request must always be answered centrally with a method list derived from the
         // routes registered for this path. It must never reach any business logic - not even a controller which
         // explicitly handles OPTIONS - so we intercept it before matching the routes below.
@@ -280,6 +291,69 @@ public class ControllerDispatcher implements WebDispatcher {
         }
 
         return DispatchDecision.CONTINUE;
+    }
+
+    /**
+     * Determines the {@link AllowedOrigin} for the given URI and stores the resolved CORS origin in the request's
+     * {@link WebContext}.
+     *
+     * <p>
+     * The stored origin can then be used in order to set the {@code Access-Control-Allow-Origin} header in the
+     * response in the relevant places by using
+     * {@link CorsAllowOriginHelper#applyHeaderFromWebContext(WebContext, Response)}.
+     * </p>
+     *
+     * <p>
+     * <b>Note:</b> If {@code WebContext.isCorsAllowAll()} is set, a {@link AllowedOrigin.MatchRequest} is used.
+     * Otherwise, the interceptors decide based on the routes matching the URI.
+     * </p>
+     *
+     * @param webContext the current request
+     * @param uri        the effective request URI
+     */
+    private void determineAndStoreAllowedOriginForUri(WebContext webContext, String uri) {
+        Collection<Route> routes = collectMatchingRoutes(webContext, uri);
+
+        Optional<AllowedOrigin> origin = WebContext.isCorsAllowAll() ?
+                                         Optional.of(new AllowedOrigin.MatchRequest()) :
+                                         determineAllowedOriginForRoutes(webContext, routes);
+
+        origin.ifPresentOrElse(actual -> corsOriginHelper.tryResolveOriginAndStoreInWebContext(webContext, actual),
+                               () -> LOG.WARN("CORS: No allowed origin could be determined for '%s'!", uri, routes));
+    }
+
+    /**
+     * Asks the interceptors to determine the {@link AllowedOrigin} for the given routes.
+     * <p>
+     * The first interceptor returning a non-empty result via
+     * {@link Interceptor#determineAllowedCorsOrigin(WebContext, Collection)} wins.
+     * </p>
+     *
+     * @param webContext the current request
+     * @param routes     the routes matching the requested URI
+     * @return the preferred CORS allowed origin, or an empty optional if none could be determined
+     */
+    private Optional<AllowedOrigin> determineAllowedOriginForRoutes(WebContext webContext, Collection<Route> routes) {
+        for (var interceptor : interceptors) {
+            Optional<AllowedOrigin> origin = interceptor.determineAllowedCorsOrigin(webContext, routes);
+            if (origin.isPresent()) {
+                return origin;
+            }
+        }
+        return Optional.empty();
+    }
+
+    /**
+     * Collects all routes matching the given URI, regardless of their HTTP method.
+     *
+     * @param webContext the current request
+     * @param uri        the effective request URI
+     * @return the routes matching the URI
+     */
+    private Collection<Route> collectMatchingRoutes(WebContext webContext, String uri) {
+        return getRoutes().stream()
+                          .filter(route -> route.matches(webContext, uri, route.isPreDispatchable()) != Route.NO_MATCH)
+                          .collect(Collectors.toSet());
     }
 
     /**
@@ -365,6 +439,7 @@ public class ControllerDispatcher implements WebDispatcher {
             response.setHeader(HttpHeaderNames.ACCESS_CONTROL_ALLOW_METHODS, allowedMethods)
                     .setHeader(HttpHeaderNames.ACCESS_CONTROL_ALLOW_HEADERS,
                                requestedHeaders == null ? "" : requestedHeaders);
+            corsOriginHelper.applyHeaderFromWebContext(webContext, response);
         }
 
         response.status(HttpResponseStatus.OK);
--- a/src/main/java/sirius/web/controller/Interceptor.java
+++ b/src/main/java/sirius/web/controller/Interceptor.java
@@ -114,8 +114,8 @@ public interface Interceptor extends Priorized {
      * </p>
      *
      * <p>
-     * Please note that {@code corsAllowAll} acts as a master switch: this method is only consulted when
-     * {@link WebContext#isCorsAllowAll()} returns {@code true} for the request's scope. If CORS handling is disabled,
+     * Please note that {@code enableCors} acts as a master switch: this method is only consulted when
+     * {@link WebContext#isCorsEnabled()} returns {@code true} for the request's scope. If CORS handling is disabled,
      * no origin is resolved at all and this method is not invoked.
      * </p>
      *
--- a/src/main/java/sirius/web/controller/Interceptor.java
+++ b/src/main/java/sirius/web/controller/Interceptor.java
@@ -108,20 +108,23 @@ public interface Interceptor extends Priorized {
      * </p>
      *
      * <p>
-     * By <b>default</b>, this method returns an {@link AllowedOrigin.MatchRequest} which will resolve to the origin of
-     * the current request.
+     * By <b>default</b>, this method returns an empty optional, signalling that this interceptor does not make a
+     * decision. If no interceptor decides on a strategy, the {@link ControllerDispatcher} falls back to an
+     * {@link AllowedOrigin.MatchRequest}, which resolves to the origin of the current request.
      * </p>
      *
      * <p>
-     * Please note that the {@link ControllerDispatcher} will ignore this choice if {@link WebContext#isCorsAllowAll()}
-     * returns {@code true}!
+     * Please note that {@code corsAllowAll} acts as a master switch: this method is only consulted when
+     * {@link WebContext#isCorsAllowAll()} returns {@code true} for the request's scope. If CORS handling is disabled,
+     * no origin is resolved at all and this method is not invoked.
      * </p>
      *
      * @param webContext the current request
      * @param routes     the routes matching the uri that has been called
-     * @return the preferred CORS allowed origin strategy
+     * @return the preferred CORS allowed origin strategy, or an empty optional if this interceptor does not make a
+     * decision
      */
     default Optional<AllowedOrigin> determineAllowedCorsOrigin(WebContext webContext, Collection<Route> routes) {
-        return Optional.of(new AllowedOrigin.MatchRequest());
+        return Optional.empty();
     }
 }
--- a/src/main/java/sirius/web/controller/Interceptor.java
+++ b/src/main/java/sirius/web/controller/Interceptor.java
@@ -9,7 +9,11 @@
 package sirius.web.controller;
 
 import sirius.kernel.di.std.Priorized;
+import sirius.web.cors.AllowedOrigin;
 import sirius.web.http.WebContext;
+
+import java.util.Collection;
+import java.util.Optional;
 
 /**
  * Can be used to intercept calls to controllers ({@link Controller})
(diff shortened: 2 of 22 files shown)


src/main/java/sirius/web/cors/CorsAllowOriginHelper.java
```diff
@@ -0,0 +1,104 @@
...
+    private Optional<String> resolveMultipleAllowedOrigins(Collection<String> origins, String requestOrigin) {
+        return origins.contains(requestOrigin) ? Optional.of(requestOrigin) : Optional.empty();
+    }
+
+    private void storeOriginInWebContext(WebContext webContext, String origin) {
+        webContext.setAttribute(ATTRIBUTE_CORS_ORIGIN, origin);
+    }
+
+    private Optional<String> getOriginFromWebContext(WebContext webContext) {
+        return webContext.safeGet(ATTRIBUTE_CORS_ORIGIN).asOptionalString();
```
jakobvogel: 🤖 _Automated review by Claude (Opus 5), posted under Jakob's account — not written by Jakob personally._

`WebContext.get(String)` checks the attribute map first, but then falls through to the query string and the posted content. So whenever no attribute was stored — which is exactly the "origin was rejected" case — this reads attacker-supplied request data.

I verified this against the PR head with the interceptor returning `Specific(["https://a.example.com", "https://b.example.com"])` and `corsAllowAll` disabled:

```
GET /system/ok?sirius_corsAllowOrigin=https://evil.example.com
Origin: https://evil.example.com

→ Access-Control-Allow-Origin: https://evil.example.com
→ Access-Control-Allow-Credentials: true
```

The allow-list is bypassed, with credentials. It also works with `?sirius_corsAllowOrigin=*`, on any path, and preflights carry the same query string so non-simple requests pass too.

Suggestion: use a channel that can't collide with user input — a typed field on `WebContext` or `CallContext` — rather than the attribute namespace.

src/main/java/sirius/web/controller/Interceptor.java
```diff
@@ -83,4 +87,41 @@ default int getPriority() {
...
+     * Please note that the {@link ControllerDispatcher} will ignore this choice if {@link WebContext#isCorsAllowAll()}
+     * returns {@code true}!
+     * </p>
+     *
+     * @param webContext the current request
+     * @param routes     the routes matching the uri that has been called
+     * @return the preferred CORS allowed origin strategy
+     */
+    default Optional<AllowedOrigin> determineAllowedCorsOrigin(WebContext webContext, Collection<Route> routes) {
+        return Optional.of(new AllowedOrigin.MatchRequest());
```
jakobvogel: 🤖 _Automated review by Claude (Opus 5), posted under Jakob's account — not written by Jakob personally._

Two things follow from this default, since `determineAllowedOriginForRoutes` takes the first non-empty answer:

1. Every existing interceptor inherits it, so the first one in the chain always decides `MatchRequest`. `sirius-biz` ships `sirius.biz.tenants.BizInterceptor` (no `getPriority` override → `DEFAULT_PRIORITY`), so in a sirius-biz app the chain resolves to "mirror the request origin" before a narrowing interceptor is ever consulted.
2. Combined with the change in `Response.setupCors`, `corsAllowAll = false` no longer means "no `Access-Control-Allow-Origin`". It now mirrors the request origin with `Allow-Credentials: true` — i.e. installations that deliberately had CORS off would get it on after upgrading.

The Javadoc says an empty optional means "not able to make a decision", which reads right — I think `Optional.empty()` is the default you want here, so abstaining is the default and only an interceptor that actually opts in decides.

src/main/java/sirius/web/http/Response.java
```diff
@@ -279,18 +283,13 @@ private void setupResponse(HttpResponseStatus status, boolean keepalive, Default
...
-        String requestedOrigin = webContext.getHeader(HttpHeaderNames.ORIGIN);
-        if (Strings.isFilled(requestedOrigin)) {
-            response.headers().set(HttpHeaderNames.ACCESS_CONTROL_ALLOW_ORIGIN, requestedOrigin);
-            if (!response.headers().contains(HttpHeaderNames.ACCESS_CONTROL_ALLOW_CREDENTIALS)) {
-                response.headers().set(HttpHeaderNames.ACCESS_CONTROL_ALLOW_CREDENTIALS, "true");
-            }
+
+        if (!response.headers().contains(HttpHeaderNames.ACCESS_CONTROL_ALLOW_CREDENTIALS)) {
+            response.headers().set(HttpHeaderNames.ACCESS_CONTROL_ALLOW_CREDENTIALS, "true");
         }
```
jakobvogel: 🤖 _Automated review by Claude (Opus 5), posted under Jakob's account — not written by Jakob personally._

`Access-Control-Allow-Credentials` is now decoupled from whether an origin header is actually emitted, so it goes out on every response. Comparing a plain request with no `Origin` header at all:

| | base (`17dc4c7d`) | this PR |
|---|---|---|
| `Access-Control-Allow-Credentials` | *(absent)* | `true` |

Beyond being noise on most responses, it makes `AllowedOrigin.Wildcard` unusable — the observed output is `Access-Control-Allow-Origin: *` together with `Allow-Credentials: true`, which browsers reject for credentialed requests. Coupling `ACAC` to a resolved, non-`*` origin would fix both.

Small related note: the old `response.headers().contains(ACCESS_CONTROL_ALLOW_ORIGIN)` guard is gone, and `setupCors` runs last in `setupResponse` (after `headers().add(headers)`), so a controller that sets the header explicitly now has it overwritten.

src/main/java/sirius/web/controller/ControllerDispatcher.java
```diff
@@ -228,6 +233,12 @@ private void preparePerformRoute(WebContext webContext,
     public DispatchDecision dispatch(WebContext webContext) throws Exception {
         String uri = determineEffectiveURI(webContext);
 
+        // In order to be able to send a CORS `Access-Control-Allow-Origin` header, we determine the
+        // preferred allowed origin and store it in the WebContext.
+        // It may then be used in the different relevant places: In this class in order to answer preflight requests
+        // and OPTIONS requests, as well as the Response class to attach it to all remaining responses.
+        determineAndStoreAllowedOriginForUri(webContext, uri);
```
jakobvogel: 🤖 _Automated review by Claude (Opus 5), posted under Jakob's account — not written by Jakob personally._

Resolving the origin here means it only happens for requests that reach `ControllerDispatcher`, which has priority `DEFAULT_PRIORITY + 10`. `AssetsDispatcher` (`- 10`) and `ServiceDispatcher` (`- 5`) run earlier and return `DONE`, so this line is never reached for them and no origin is stored.

Measured with `corsAllowAll` enabled:

| | base (`17dc4c7d`) | this PR |
|---|---|---|
| `GET /assets/test.css` + `Origin` | `ACAO: https://friendly.example.com` | *(absent)* |

`ServiceDispatcher` takes the same path, and `/service/*` is where most cross-origin API consumers sit, so that one is probably the more painful case in practice. Doing the resolution upstream of the dispatcher pipeline would cover all dispatchers uniformly.

src/main/java/sirius/web/controller/ControllerDispatcher.java
```diff
@@ -282,6 +293,69 @@ && shouldExecute(webContext, uri, route, true) != Route.NO_MATCH)) {
...
+     */
+    private void determineAndStoreAllowedOriginForUri(WebContext webContext, String uri) {
+        Collection<Route> routes = collectMatchingRoutes(webContext, uri);
+
+        Optional<AllowedOrigin> origin = WebContext.isCorsAllowAll() ?
+                                         Optional.of(new AllowedOrigin.MatchRequest()) :
+                                         determineAllowedOriginForRoutes(webContext, routes);
+
+        origin.ifPresentOrElse(actual -> corsOriginHelper.tryResolveOriginAndStoreInWebContext(webContext, actual),
+                               () -> LOG.WARN("CORS: No allowed origin could be determined for '%s'!", uri, routes));
```
jakobvogel: 🤖 _Automated review by Claude (Opus 5), posted under Jakob's account — not written by Jakob personally._

This fires whenever the chain abstains, including for requests that carry no `Origin` header at all — so once the default becomes `Optional.empty()`, a `corsAllowAll = false` deployment logs a warning for every ordinary same-origin page load. It showed up in my local test run. Gating it on the presence of an `Origin` header (or dropping it to `FINE`) would keep it useful.

Minor: `routes` is passed as an argument but the format string only has one placeholder, so it's silently dropped.

src/main/java/sirius/web/controller/ControllerDispatcher.java
```diff
@@ -282,6 +293,69 @@ && shouldExecute(webContext, uri, route, true) != Route.NO_MATCH)) {
...
+     * Collects all routes matching the given URI, regardless of their HTTP method.
+     *
+     * @param webContext the current request
+     * @param uri        the effective request URI
+     * @return the routes matching the URI
+     */
+    private Collection<Route> collectMatchingRoutes(WebContext webContext, String uri) {
+        return getRoutes().stream()
+                          .filter(route -> route.matches(webContext, uri, route.isPreDispatchable()) != Route.NO_MATCH)
+                          .collect(Collectors.toSet());
```
jakobvogel: 🤖 _Automated review by Claude (Opus 5), posted under Jakob's account — not written by Jakob personally._

This adds a full scan over all routes with no early exit, on every request, on top of the existing match loop (and a third scan for preflights).

Worth noting that `Route.matches` isn't side-effect free: it calls `CallContext.addToMDC("route", …)` and `webContext.setAttribute(...)` for `#` parameters. Since this pass doesn't stop at the first match, routes declared *after* the matching one can now write attributes into the context that the dispatch loop would never have reached.

Skipping the whole resolution when the request carries no `Origin` header would avoid the cost for the large majority of requests.

jakobvogel (commented):
🤖 **Automated review by Claude (Opus 5).** Posted under Jakob's account, but written by Claude, not by Jakob personally.

The design here is good — a sealed `AllowedOrigin` with `MatchRequest`/`Specific`/`Wildcard`, resolved once per request and applied centrally, is the right shape, and the test rewrite reads much better than before. The findings below are all in the wiring, not the idea.

I built the PR head and its base and ran probe requests against the test server, so the behavioural claims are measured rather than inferred.

**Worth addressing before merge:**

1. **The allow-origin can be forged via a query parameter.** The resolved origin is stored as a `WebContext` attribute, but `WebContext.get()` falls through to the query string and post body when no attribute is set. `?sirius_corsAllowOrigin=https://evil.example.com` bypasses a `Specific` allow-list and comes back with `Allow-Credentials: true`. (Comment in `CorsAllowOriginHelper`.)
2. **The `Interceptor` default returns `MatchRequest` rather than `empty()`.** Every existing interceptor inherits it and short-circuits the chain, and `corsAllowAll = false` installations would start mirroring the request origin after upgrading.
3. **CORS stops working for dispatchers ahead of `ControllerDispatcher`** — confirmed absent for `/assets/*`; `/service/*` follows the same path.
4. **`Allow-Credentials: true` is now unconditional**, which also makes `Wildcard` unusable (`ACAO: *` plus credentials is rejected by browsers).

**Smaller items:** the new WARN fires on ordinary non-CORS traffic; `collectMatchingRoutes` adds a full route scan per request; the manual `ACAO` override guard was dropped.

**Test coverage note:** `Allow-Credentials`, `Vary`, and the "no interceptor overrides the default" path aren't asserted anywhere — which is why points 2 and 4 pass the suite today. Also FYI, no build/test check seems to have run on this PR yet, only `label` and the Copilot reviewer.

Happy to be wrong on any of these — the probes were run locally, so please do sanity-check them on your side.

jakobvogel (dismissed):
Die Claude-Findings erscheinen mir schwerwiegend. Bitte genau prüfen, v.a. die möglichen Angriffsvektoren und entsprechend Code anpassen; ebenfalls, wie von Claude vorgeschlagen, die Test Coverage erhöhen und sicherstellen, dass die Tests durchlaufen

src/main/java/sirius/web/controller/ControllerDispatcher.java
hso-scireum: This has been resolved by deciding that `Dispatcher`s need to handle CORS themselves. `Controllerdispatcher` will continue to offer the abstraction via the `Interceptor` class. `Request` will respect already set headers to manually added CORS headers will not be overridden.

idlira (approved):

src/test/java/sirius/web/http/TestCorsInterceptor.java
```diff
@@ -0,0 +1,56 @@
+/*
+ * Made with all the love in the world
+ * by scireum in Remshalden, Germany
+ *
+ * Copyright by scireum GmbH
```
mkeckmkeck: Bitte den aktuellen Header nutzen

src/test/java/sirius/web/http/TestCorsScopeDetector.java
```diff
@@ -0,0 +1,64 @@
+/*
+ * Made with all the love in the world
+ * by scireum in Remshalden, Germany
+ *
+ * Copyright by scireum GmbH
+ * http://www.scireum.de - info@scireum.de
```
mkeckmkeck: Bitte den aktuellen Header nutzen

src/main/java/sirius/web/cors/CorsAllowOriginHelper.java
```diff
@@ -0,0 +1,156 @@
...
+/**
+ * Resolves the effective {@code Access-Control-Allow-Origin} for a request and applies it to the response.
+ * <p>
+ * An {@link AllowedOrigin} (as determined by the {@link sirius.web.controller.Interceptor interceptors}) describes the
+ * <i>strategy</i> for the header. This helper turns that strategy into a concrete origin for the current request,
+ * stores it in the request scoped {@link CorsContext}, and later writes it to the response as the actual CORS header.
+ * </p>
+ */
+@Register(classes = CorsAllowOriginHelper.class)
+public class CorsAllowOriginHelper {
```
mkeckmkeck: Naming: I do not like the Helper suffix as this class is a Part and not a scoped helper. Maybe CorsAllowOriginUtil or CorsAllowOriginResolver?

mkeckmkeck (approved):

src/test/java/sirius/web/http/TestCorsInterceptor.java
idlira: Vllcht ein mal ein search-replace in einen extra commit/PR machen? KI kopiert immer wieder aus alten Klassen. Wir haben in OX mal gemacht. Eventuell ne mini task für ein Werkstudi 😊

sabieber: In OX hieß der Werkstudent Claude oder Cursor 😁

hso-scireum: Ist für die neu hinzugefügten Klassen angepasst 👍🏻

src/main/java/sirius/web/cors/CorsAllowOriginHelper.java
hso-scireum: Good call, I decided on `CorsAllowOriginResolver` for now 👍🏻 The helper evolved quite a bit during development, so you're right: it's not really a helper anymore

src/main/java/sirius/web/cors/CorsAllowOriginResolver.java
```diff
@@ -0,0 +1,156 @@
...
+        // If `enableCors` is disabled, we explicitly do not want to handle anything CORS.
+        if (!webContext.isCorsEnabled()) {
+            return;
+        }
+
+        var corsContext = CorsContext.get();
+        var origin = allowedOriginSupplier.get();
+
+        corsContext.setConfiguredOrigin(origin);
+        tryResolveOrigin(webContext, origin).ifPresent(corsContext::setResolvedOrigin);
```
jakobvogel, line 88: 🤖 _Automated review by Claude (Opus 5), posted under Jakob's account — not written by Jakob personally._

These two lines overwrite unconditionally, so when several dispatchers run for one request the *last* one wins — and that's the least restrictive one. With an interceptor returning `Specific(false, ["https://a.example.com"])` and `Origin: https://evil.example.com`:

| request | resulting `Access-Control-Allow-Origin` |
|---|---|
| `/system/ok` (route matches) | *(absent)* — policy honoured |
| `/no-such-path-404` | `https://evil.example.com` |
| `/service/no-such-service` | `https://evil.example.com` |

`ControllerDispatcher` stores the `Specific` strategy and returns `CONTINUE`, then `ServiceDispatcher`/`DefaultDispatcher` replace it with their hardcoded `MatchRequest`.

Only error pages are affected today and no credentials ride along, so the practical impact is small — but the configured policy isn't enforced, and any dispatcher added later inherits the same behaviour. Keeping the first stored decision (or having the dispatchers only set a strategy when none is present yet) would close it.

src/main/java/sirius/web/http/Response.java
```diff
@@ -279,17 +284,40 @@ private void setupResponse(HttpResponseStatus status, boolean keepalive, Default
...
+        // interceptor determining the origin:
+
+        var shouldAllowCredentials = corsOriginResolver.getConfiguredOrigin()
+                                                       .filter(origin -> origin instanceof AllowedOrigin.Specific)
+                                                       .map(origin -> ((AllowedOrigin.Specific) origin).allowCredentials())
+                                                       .orElse(false);
+
+        if (shouldAllowCredentials && !headers.contains(HttpHeaderNames.ACCESS_CONTROL_ALLOW_CREDENTIALS)) {
+            headers.set(HttpHeaderNames.ACCESS_CONTROL_ALLOW_CREDENTIALS, "true");
+        }
```
jakobvogel, line 317: 🤖 _Automated review by Claude (Opus 5), posted under Jakob's account — not written by Jakob personally._

`shouldAllowCredentials` is derived from the *configured* strategy only, so the header is emitted even when the origin was rejected. With `Specific(true, ["https://a.example.com", "https://b.example.com"])` and `Origin: https://evil.example.com` I get:

```
Access-Control-Allow-Origin:      (absent)
Access-Control-Allow-Credentials: true
```

Harmless in the browser since both headers are required together, but it's a contradictory response and it does advertise that a credentialed allow-list exists behind the endpoint.

`applyResolvedOriginHeader` on line 299 already returns exactly the boolean needed here (`true` if an origin was applied) and the result is currently discarded — capturing it and `&&`-ing it into the condition would keep the two headers consistent.

Minor style point while you're in here: `.filter(origin -> origin instanceof AllowedOrigin.Specific).map(origin -> ((AllowedOrigin.Specific) origin).allowCredentials())` can drop the cast with pattern matching (`instanceof AllowedOrigin.Specific specific`), which SonarLint tends to prefer.

src/main/java/sirius/web/http/Response.java
```diff
@@ -279,17 +284,40 @@ private void setupResponse(HttpResponseStatus status, boolean keepalive, Default
...
+
+        // Regardless whether the CORS origin has actually been resolved/applied, we must send a `Vary` header for all
+        // resources depending on the origin. For performance reasons we do always set the header, though, in order to
+        // avoid expensive evaluation of the origin for same-origin requests without an `Origin` header.
+        // Also see `ControllerDispatcher#dispatch()`.
+
+        if (!headers.contains(HttpHeaderNames.VARY)) {
+            headers.set(HttpHeaderNames.VARY, HttpHeaderNames.ORIGIN);
         }
```
jakobvogel, line 333: 🤖 _Automated review by Claude (Opus 5), posted under Jakob's account — not written by Jakob personally._

This skips `Vary` entirely when *any* `Vary` header is already present, rather than adding `Origin` to it. `Vary` is not part of `TunnelHandler.NON_TUNNELLED_HEADERS`, so an upstream `Vary: Accept-Encoding` is forwarded and then wins here — and the response stays origin-dependent while `Origin` never enters the cache key. Behind a shared cache that shows up as intermittent CORS failures. Appending rather than skipping would be safer. (Not a data-leak: a mismatched `ACAO` fails closed in the browser.)

Separately, `Vary: Origin` is also emitted for `Wildcard` responses, which are byte-identical for every origin — I see it on `/assets/*` and `/help`. That fragments CDN caches on the highest-traffic paths for no benefit. It's pre-existing behaviour rather than something this PR introduces, but now that the strategy is known at this point it'd be cheap to skip for `Wildcard`.

src/main/resources/component-065-web.conf
```diff
@@ -116,10 +116,12 @@ http {
...
-    # Determines if all cors requests (preflight requests especially) should be accepted by the server
-    # automatically. If you need a more advanced behaviour, you probably want to implement a custom
-    # dispatcher. This can be overridden per security scope via scope config using http.corsAllowAll.
-    corsAllowAll = false
+    # Determines whether automatic CORS handling is enabled. When enabled, the appropriate
+    # `Access-Control-*` headers (including answers to preflight requests) are sent automatically;
+    # when disabled, no CORS headers are emitted at all. For fine-grained control over the allowed
+    # origins, register a custom Interceptor. This can be overridden per security scope via scope
+    # config using http.enableCors.
+    enableCors = false
```
jakobvogel, line 124: 🤖 _Automated review by Claude (Opus 5), posted under Jakob's account — not written by Jakob personally._

There are no references to `http.corsAllowAll` left anywhere in the codebase, and Typesafe Config won't complain about an unknown key — so an installation that keeps `corsAllowAll = true` in its config silently falls back to `enableCors = false` and loses CORS handling with no signal at all. The failure mode is a cross-origin integration that stops working after a deploy, with nothing in the log pointing at the cause.

The `[!CAUTION]` note in the PR description covers people who read the release notes; a runtime hint would cover the rest. Reading the legacy key for one release and logging a deprecation warning when it's present would make the migration self-diagnosing.

src/main/java/sirius/web/security/UserContext.java
```diff
@@ -580,6 +580,30 @@ public ScopeInfo getScope() {
...
+     * @param webContext the request to detect the scope from
+     * @return the currently installed scope, the detected scope or {@link ScopeInfo#DEFAULT_SCOPE}
+     */
+    public ScopeInfo peekScope(WebContext webContext) {
+        if (currentScope != null) {
+            return currentScope;
+        }
+        if (detector != null && webContext != null && webContext.isValid()) {
+            return detector.detectScope(webContext);
+        }
```
jakobvogel, line 603: 🤖 _Automated review by Claude (Opus 5), posted under Jakob's account — not written by Jakob personally._

Good catch on the premature scope binding — that was a real bug and I'd missed it in the first round.

Two small things about the peek itself:

The result isn't memoized, so `detectScope` runs more than once. I instrumented the detector and counted calls per request: **2** for a request carrying an `Origin` header (once here, once when the scope is really installed), 1 without. Caching the peeked scope in a field would bring it back to one.

More relevant than the cost: `detectScope` is now invoked speculatively and its result sometimes thrown away. Existing detectors were written against a "detected ⇒ installed" contract, so any implementation doing bookkeeping in `detectScope` would now see extra calls. Probably fine for the ones in `sirius-biz`, but worth a quick confirmation with whoever owns them, since the Javadoc here doesn't currently warn implementors about it.

jakobvogel (commented):
🤖 **Automated review by Claude (Opus 5).** Posted under Jakob's account, but written by Claude, not by Jakob personally. Second pass, following up on the review of `aaa9af1f`.

This is a big improvement — everything from the first round is genuinely fixed, and I re-measured rather than assuming. I built the new head and re-ran the earlier probes plus some new ones.

| Earlier finding | Status |
|---|---|
| Origin forgeable via query parameter | Gone — the typed `CorsContext` closed it properly |
| `Interceptor` default returned `MatchRequest` | Fixed — defaults to `empty()`, fallback moved into the dispatcher |
| CORS lost for `/assets/*` and `/service/*` | Fixed — all four dispatchers participate |
| Unconditional `Allow-Credentials` | Fixed — explicit opt-in via `Specific` |

I also checked two things the rework newly exposed: `Route#matchesUri` is genuinely side-effect free, and `CorsContext` does not bleed across keep-alive requests (probed a wildcard asset followed by a controller hit on the same connection — clean). The `peekScope` fix for the premature scope binding was a real bug that I missed first time round; nice catch.

**Remaining findings**, none critical, all inline:

1. A later dispatcher overwrites an interceptor's restrictive policy — a `Specific` allow-list is replaced by `MatchRequest` on 404 and unknown-service paths, so those reflect any origin.
2. `Allow-Credentials: true` is emitted even when the origin was rejected and no `Allow-Origin` is sent. Inert in browsers, but contradictory — and `applyResolvedOriginHeader` already returns the boolean that would gate it.
3. The `corsAllowAll` → `enableCors` rename is silent at runtime; a stale key just disables CORS with nothing in the log.
4. `Vary` is skipped rather than extended when a header is already present (relevant for tunnelled responses, where `Vary` is forwarded).
5. `peekScope` isn't memoized — measured 2 `detectScope` calls per CORS request — and detectors are now called speculatively.

**One thing for the PR description rather than the code:** the credentials change deserves a line in the breaking-change block. Measured on this head, `enableCors = true` with the default `MatchRequest` sends no `Access-Control-Allow-Credentials`, where `main` sent `true`. Every existing `credentials: 'include'` consumer breaks. It's the right call security-wise, and probably a bigger practical break than the config rename — it just isn't documented yet. (Also three typos in the `[!CAUTION]` block: *cofnguration*, *yonfig*, *ControllerDisaptcher*.)

**Test coverage** is much better with the credentials and `Vary` cases. Still uncovered: the non-controller dispatchers this PR adds CORS to, a regression test for the query-parameter forging, and the `Vary`-already-present path.

As before, all of this was probed locally, so please sanity-check anything that looks off.

jakobvogel:
🤖 _Automated summary by Claude (Opus 5), posted under Jakob's account — not written by Jakob personally._

> [!NOTE]
> **Documentation only — no action required on this comment.** It's a map of how CORS is answered after this PR, written up so reviewers (and future readers) don't have to reconstruct it from the diff. Nothing here is a finding or a change request.

## How CORS is answered after this PR

The handling is split into two phases, and that split is the key thing to hold onto: **a dispatcher decides *which* origin policy applies, `Response` writes the headers for every response, and `CorsContext` is the request-scoped hand-off between them.**

```
REQUEST
  │
  ├─ DECIDE  (during dispatch, in whichever dispatcher handles the request)
  │    each dispatcher calls corsOriginResolver.tryResolveAndStoreOrigin(...):
  │
  │      AssetsDispatcher     (prio  -10)  → Wildcard       hardcoded
  │      ServiceDispatcher    (prio   -5)  → MatchRequest   hardcoded
  │      ControllerDispatcher (prio  +10)  → asks Interceptors, else MatchRequest
  │      HelpDispatcher       (prio +100)  → Wildcard       hardcoded
  │      DefaultDispatcher    (prio  999)  → MatchRequest   hardcoded
  │
  │    inside tryResolveAndStoreOrigin:
  │      1. enableCors off?  → store nothing, done          (master switch)
  │      2. store the STRATEGY        → CorsContext.configuredOrigin
  │      3. resolve strategy → concrete origin:
  │           MatchRequest → the request's Origin header
  │           Specific     → the Origin, if it is in the allow-list
  │           Wildcard     → "*"
  │      4. store the RESULT          → CorsContext.resolvedOrigin
  │
  │    CorsContext = request-scoped SubContext on CallContext (fresh per request)
  │
  └─ WRITE  (when the response is built)
       Response.setupCors() — runs for EVERY response built through Response:
         · bail out if enableCors is off
         · Access-Control-Allow-Origin      ← CorsContext.resolvedOrigin
         · Access-Control-Allow-Credentials ← only for Specific(allowCredentials = true)
         · Vary: Origin                     ← unless a Vary header already exists

       Preflight only (OPTIONS + Access-Control-Request-Method):
         ControllerDispatcher.answerOptionsRequest() additionally writes
         Allow / Allow-Methods / Allow-Headers, and writes Allow-Origin itself.
         Preflights are answered ONLY by ControllerDispatcher.
```

**The three files that matter:**

| File | Role |
|---|---|
| `sirius/web/cors/CorsAllowOriginResolver` | the decide logic — strategy → concrete origin |
| `sirius/web/cors/CorsContext` | the request-scoped hand-off between decide and write |
| `Response#setupCors` | the write — the single place all non-preflight responses get their headers |

Everything else (the five dispatchers, `Interceptor#determineAllowedCorsOrigin`, `AllowedOrigin`) is a caller or an input to those three.

**Two behaviours that are easy to miss when reading the diff:**

- `ControllerDispatcher` only resolves an origin when the request is a preflight **or** carries an `Origin` header. Everything else skips CORS entirely. The other dispatchers resolve unconditionally, which is why e.g. `/assets/*` gets `Access-Control-Allow-Origin: *` even on a request with no `Origin` header.
- `enableCors` is a hard master switch checked in *both* phases — with it off, no CORS header is emitted anywhere, not even `Vary`.

Config: `http.enableCors` (default `false`), overridable per security scope, read via `WebContext#isCorsEnabled()`.

jakobvogel:
🤖 _Automated review by Claude (Opus 5), posted under Jakob's account — not written by Jakob personally._

## Test coverage: concrete gaps and requested cases

The suite covers the `ControllerDispatcher` path thoroughly — all three strategies, the full credentials matrix, `Vary`, preflight. Two areas have no coverage at all, and both are places this PR changed behaviour. In priority order:

### 1. A regression test for the `peekScope` fix (highest value)

Commit `d2d9d08` fixed a real bug: resolving CORS bound the scope, which discarded a manually installed user, which broke the permission check. Nothing pins that fix. If `isCorsEnabled()` is ever simplified back to `UserContext.getCurrentScope()`, **all 19 CORS tests stay green** and authentication breaks for every request carrying an `Origin` header.

`PermissionsRouteTest` would already catch it — except it sends no `Origin` header, and `ControllerDispatcher` short-circuits CORS resolution entirely without one. So it passes either way.

```kotlin
@Test
fun `a manually installed user survives the CORS origin resolution`() {
    UserContext.get().setCurrentUser(
        UserInfo.Builder.createUser("test").withPermissions(setOf("permission-system-tags")).build()
    )

    val response = TestRequest.GET("/system/tags")
        .addHeader(HttpHeaderNames.ORIGIN, "https://example.com")
        .execute()

    assertEquals(TestResponse.ResponseType.TEMPLATE, response.getType())
}
```

Fails without the fix (`ResponseType.ERROR`), passes with it.

### 2. The four non-controller dispatchers — currently zero coverage

Every test hits `/system/ok` or `/test/another-restricted-method`. This PR newly gives CORS to Assets, Help, Service and Default, each with a hardcoded strategy.

Two prerequisites: `sendGet` hardcodes `/system/ok` — it needs a `uri: String = "/system/ok"` parameter (`sendPreflight` already has one); and `connection.inputStream.close()` throws on 404, so those cases need `runCatching { }` or `errorStream`.

| # | Request | Expected |
|---|---|---|
| 2a | `GET /assets/test.css` + Origin | `ACAO: *`, no `ACAC` |
| 2b | `GET /assets/test.css`, **no** Origin | `ACAO: *` — assets resolve unconditionally |
| 2c | `GET /help` + Origin | `ACAO: *` |
| 2d | `GET /service/unknown` + Origin | `ACAO` reflects the origin |
| 2e | `GET /no-such-path` (404) + Origin | `ACAO` reflects the origin |
| 2f | **each of the above with `enableCors` disabled** | no CORS headers at all |

2f is the important one — it proves the master switch gates the non-controller dispatchers, which today is only verified for the controller path.

(The expected values above are measured against the current head, not assumed.)

### 3. Two cases that should fail until the review findings are fixed

> [!NOTE]
> 3a and 3b are written deliberately to **fail today**. They are specifications for the two fixes requested in the review comments, not descriptions of current behaviour.

**3a — credentials without an allowed origin.** The existing credentials tests only ever use *allowed* origins, which is exactly why this slipped through:

```kotlin
TestCorsInterceptor.allowedOrigin = AllowedOrigin.Specific(true, specificAllowedOrigins)
val connection = sendGet("https://evil.example.com")

assertNull(connection.getHeaderField(ACCESS_CONTROL_ALLOW_ORIGIN.toString()))
assertNull(connection.getHeaderField(ACCESS_CONTROL_ALLOW_CREDENTIALS.toString()))  // currently "true"
```

**3b — restrictive policy on an unmatched path:**

```kotlin
TestCorsInterceptor.allowedOrigin = AllowedOrigin.Specific(false, setOf("https://a.example.com"))
assertNull(requestAllowedOrigin("https://evil.example.com", uri = "/no-such-path"))
// currently reflects the origin: DefaultDispatcher overwrites the stored strategy
```

**3c — `Vary` when a header already exists.** Needs a route or tunnel that sets its own `Vary`; assert `Origin` is still listed. Lower priority, more setup.

### 4. Nice to have

- Controller request **without** an `Origin` header → no `ACAO` (documents the short-circuit, and is the counterpart to 2b).
- Two interceptors at different priorities → the higher-priority one wins. "First to decide wins" is documented but unverifiable with a single test interceptor.

### Housekeeping

- `TestCorsInterceptor.allowedOrigin` is `public static volatile` with an `@AfterEach` reset — correct sequentially, but worth a comment noting these tests can't run in parallel.
- `@AfterEach` sits above the test methods; house style is test methods first, lifecycle hooks and helpers below.

jakobvogel (approved):

src/main/resources/component-065-web.conf
hso-scireum: This is true in theory but since there are no real-world uses of the config flag, there will be no changes to this one. The config value will now primarily work as a kill-switch and is not harmful if set to `false`.

src/main/java/sirius/web/security/UserContext.java
hso-scireum: I'll keep this in mind for future optimizations

hso-scireum:
@jakobvogel I just pushed a bunch of commits addressing most of the newly introduced findigs, if you could give them a quick look and re-approve this would be great 👍 I omitted some of the smaller suggestions in order to get this merged as quickly as possible, maybe we can address them in a follow-up PR

src/main/java/sirius/web/http/Response.java
```diff
@@ -279,17 +284,52 @@ private void setupResponse(HttpResponseStatus status, boolean keepalive, Default
...
+    }
+
+    private void setOrAppendOriginToVaryHeader(HttpHeaders headers) {
+        if (!headers.contains(HttpHeaderNames.VARY)) {
+            headers.add(HttpHeaderNames.VARY, HttpHeaderNames.ORIGIN);
+        } else {
+            var current = headers.get(HttpHeaderNames.VARY);
+            headers.set(HttpHeaderNames.VARY, String.format("%s, %s", current, HttpHeaderNames.ORIGIN));
         }
     }
```
jakobvogel, line 334: 🤖 _Automated review by Claude (Opus 5), posted under Jakob's account — not written by Jakob personally._

The append fixes the overwrite, but this variant drops header values. Netty's `get()` returns only the **first** value, and `set()` removes all of them — so a response carrying two `Vary` lines loses one. Reproduced by adding a second `addHeader` to the `/test/vary` route:

```java
.addHeader(HttpHeaderNames.VARY, HttpHeaderNames.ACCEPT_ENCODING)
.addHeader(HttpHeaderNames.VARY, HttpHeaderNames.ACCEPT_LANGUAGE)
```

```
Vary: accept-encoding, origin      ← Accept-Language silently gone
```

`headers.add(HttpHeaderNames.VARY, HttpHeaderNames.ORIGIN)` on its own does the right thing here — repeated field lines are semantically equivalent to comma-joining (RFC 9110 §5.3), and Netty emits them as separate lines. That makes the whole if/else unnecessary and also avoids the `Vary: Origin, Origin` you currently get if something already varied on `Origin`.

Worth pinning with a two-value test — the current suite passes straight over this (417 tests, none catch it).

src/main/java/sirius/web/cors/CorsContext.java
```diff
@@ -0,0 +1,124 @@
...
+    /**
+     * Stores the origin resolved for the current request.
+     *
+     * @param resolvedOrigin the resolved origin to store
+     * @throws IllegalStateException if the context has already been finalized
+     */
+    public void setResolvedOrigin(@Nullable String resolvedOrigin) {
+        if (isFinalized) {
+            throw new IllegalStateException("Cannot modify finalized CORS context");
+        }
```
jakobvogel, line 92: 🤖 _Automated review by Claude (Opus 5), posted under Jakob's account — not written by Jakob personally._

A suggestion rather than a defect — the first-writer-wins mechanism itself looks right.

These setters are public on a class with a public static `get()`, and they now throw. Meanwhile `WebDispatcher`'s javadoc explicitly tells dispatcher authors to handle CORS themselves, so a custom dispatcher reaching for `CorsContext.get().setResolvedOrigin(...)` directly would get an `IllegalStateException` **thrown mid-dispatch** — turning a CORS bookkeeping mistake into a 500.

Since `CorsAllowOriginResolver` lives in the same package and already guards with `isFinalized()` before calling these, making both setters (and `markFinalized`) package-private would close that off entirely and leave `tryResolveAndStoreOrigin` as the only public way in. It costs the one test asserting the throw, which seems a fair trade.

Minor doc point either way: the class javadoc here says the setters "refuse any further modification" by throwing, while `CorsAllowOriginResolver`'s says subsequent calls are "silently ignored". Both are accurate from their own side, but read together they look contradictory.

jakobvogel (commented):
🤖 **Automated review by Claude (Opus 5).** Posted under Jakob's account, but written by Claude, not by Jakob personally.

Everything from the last round is addressed, and a couple of the fixes are better than what was asked for — `d482b26e` (only resolving for services that are actually handled) is an interaction I'd missed, and the `peekScope` regression test correctly works around the forked-task problem and asserts that the origin really was resolved, so it can't silently degrade into a duplicate of `PermissionsRouteTest`. Nice.

### The red build

Drone build 8326: `417 tests, 1 failure` — the `/help` test at `CorsTest.kt:263`. `551f590f` removed the resolution from `HelpDispatcher` but left the test asserting it.

I think **removing it was right and the test should just go.** `ControllerDispatcher` (+10) runs before `HelpDispatcher` (+100) and resolves for any request carrying an `Origin` header, so `HelpDispatcher`'s `Wildcard` could only ever win for requests *without* an `Origin` — i.e. requests that don't need CORS in the first place. It was unreachable in every case that mattered.

So: delete `given enableCors is enabled when the help system is requested without an origin then the HelpDispatcher allows any origin`, and update the comment on its sibling test, which still describes a `Wildcard` that no longer exists.

### Inline

Two items on the code, both in the comments above: the `Vary` append drops values when more than one `Vary` header is present (a real bug, with a simpler fix), and a suggestion to make the `CorsContext` setters package-private.

### Still open from earlier rounds

All small:

- `http.corsAllowAll` still has no legacy handling and no startup warning — a stale key silently disables CORS with nothing in the log.
- The credentials change isn't in the breaking-change block yet: `enableCors = true` with the default `MatchRequest` no longer sends `Access-Control-Allow-Credentials`, where `main` did. Arguably a bigger practical break for existing consumers than the config rename.
- `peekScope` isn't memoized — measured 2 `detectScope` calls per CORS request. Genuinely minor.
- None of the new dispatcher tests run with `enableCors` disabled, so the master switch is still only proven for the controller path.

### Unrelated but adjacent

`-Dsun.net.http.allowRestrictedHeaders=true` only exists in `.drone.yml`. Since `Origin` is on the JDK's restricted-header list for `HttpURLConnection`, moving it into the surefire configuration in `pom.xml` would make a local `mvn test` of `CorsTest` behave like CI. This PR is the one adding most of the tests that depend on it.

Once the build is green I don't expect anything further from my side.

jakobvogel (approved):

PR #1704, opened 2026-08-06
Repairs the Wondergem image upload (MIO-6562)

jakobvogel:
### Description

Repairs the Wondergem `<w:imageUpload>` tag, which has been unable to upload anything at all. Three defects, one commit each:

**1. No CSRF token was sent.** `imageUpload()` passed no parameters to the underlying `qq.FileUploader`, so its POST reached the server without a token. Since CSRF validation became the default for every POST (SIRI-1216), every `<w:imageUpload>` was answered with a `403`. The helper now takes an optional `params` object — exactly as `fileUpload()` already did — and the taglib supplies the token through it. This is the defect behind the reported customer issue.

**2. The default extension filter rejected every file.** `allowedExtensions` defaulted to the four character string `'[]'`, so the uploader received the JavaScript *string* `"[]"` rather than an empty array. Its extension check then matched each file against the two characters `[` and `]`, refused all of them, and threw `allowed.join is not a function` while assembling the error message. Dropping the inner quotes yields an empty array literal, matching what `<w:fileUpload>` has always used. Only callers that omit the argument were affected — every current consumer passes it explicitly.

**3. The upload URL was escaped for the wrong context.** Both upload taglibs emit the URL into a JavaScript string literal, but the value was escaped by the XML escaper an HTML template renders with. Every character that escaper touches is destroyed on the way to the server — `?name=O'Brien` arrives as a lone `name=O`, because the `&` opening the entity restarts query string parsing. An `&` of our own currently survives, but only by accident: netty's `QueryStringDecoder` splits on `;` as well, so `&amp;a=b` degrades into a junk empty `amp` parameter next to an intact `a=b`. That is non-standard and being phased out, so it is nothing to rely upon. Escaping for the surrounding JavaScript context keeps the URL intact either way, and additionally guards against a stray quote or a `</script>` breaking out of the literal.

`previewUrl` and `imageUrl` are deliberately left untouched in (3): those end up inside an HTML attribute assembled by the uploader, where the entity is what the browser expects and decodes for itself.

No breaking changes. `imageUpload()` gained a trailing optional argument, so existing six-argument callers keep working unchanged. No consumer in our repositories writes `&amp;` into an `uploadUrl`, and apart from one memoio template every `w:` consumer has zero or one query parameter, where (3) is a no-op.

### Verification

`WondergemUploadTest` pins all three behaviours for both upload taglibs; every assertion was confirmed to fail when its own fix is reverted. The URL assertions read the emitted value back and undo the JavaScript escapes, so they pin the URL that actually reaches the uploader rather than a particular escaping scheme. Full suite green.

The emitted JavaScript was also evaluated in a JS engine to confirm the literal behaves as intended: `'\/test\/upload?a=1&b=2'` evaluates to `/test/upload?a=1&b=2`, an embedded apostrophe round-trips, and `</script>` never appears unescaped in the source.

Beyond that, the change was verified end to end against a locally running memoio, using an artifact built from tag `107.0.5` plus these three commits, so the delta under test was exactly this PR:

| Upload | Without the token | With the token the fixed taglib emits |
| --- | --- | --- |
| `/account/avatar` (user picture) | `403` | `200`, avatar stored |
| `/company/:1/avatar` (company picture) | `403` | `200`, avatar stored |
| `/company/mail-config/header-image` | `403` | `200`, image stored |

The `403` reproduces the reported customer symptom. `<w:fileUpload>` pages were re-checked for regressions and still emit their token and their full URL unchanged.

### Additional Notes

- This PR fixes or works on following ticket(s): [MIO-6562](https://scireum.myjetbrains.com/youtrack/issue/MIO-6562)
- Follow-up for consumers: memoio needs the resulting release to pick the fix up; its three `<w:imageUpload>` call sites need no change of their own.

### Checklist

### DIFF
--- a/src/main/resources/default/taglib/w/imageUpload.html.pasta
+++ b/src/main/resources/default/taglib/w/imageUpload.html.pasta
@@ -17,7 +17,7 @@
                 '@imageUrl',
                 <i:raw>@allowedExtensions</i:raw>,
                 @singleFile,
-                {CSRFToken: '@part(sirius.web.http.CSRFHelper.class).getCSRFToken()'});
+                {CSRFToken: '@part(CSRFHelper.class).getCSRFToken()'});
         });
     });
 </script>
--- a/src/main/resources/default/taglib/w/imageUpload.html.pasta
+++ b/src/main/resources/default/taglib/w/imageUpload.html.pasta
@@ -15,9 +15,9 @@
                 this,
                 '@previewUrl',
                 '@imageUrl',
-                 <i:raw>@allowedExtensions</i:raw>,
-                @singleFile
-        );
+                <i:raw>@allowedExtensions</i:raw>,
+                @singleFile,
+                {CSRFToken: '@part(sirius.web.http.CSRFHelper.class).getCSRFToken()'});
         });
     });
 </script>
--- a/src/main/resources/default/taglib/w/imageUpload.html.pasta
+++ b/src/main/resources/default/taglib/w/imageUpload.html.pasta
@@ -11,7 +11,7 @@
 <script type="text/javascript">
     $(document).ready(function() {
         $('.@name').each(function (index) {
-            imageUpload('@uploadUrl',
+            imageUpload('<i:raw>@escapeJS(uploadUrl)</i:raw>',
                 this,
                 '@previewUrl',
                 '@imageUrl',
--- a/src/main/resources/default/taglib/w/imageUpload.html.pasta
+++ b/src/main/resources/default/taglib/w/imageUpload.html.pasta
@@ -2,7 +2,7 @@
 <i:arg type="String" name="uploadUrl" />
 <i:arg type="String" name="imageUrl" default="" />
 <i:arg type="String" name="previewUrl" default="@imageUrl" />
-<i:arg type="String" name="allowedExtensions" default="'[]'" />
+<i:arg type="String" name="allowedExtensions" default="[]" />
 <i:arg type="boolean" name="singleFile" default="false" />
 
 <i:pragma name="description" value="Renders an image upload within a Wondergem template" />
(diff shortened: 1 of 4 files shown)


src/main/resources/default/taglib/w/imageUpload.html.pasta
```diff
@@ -15,9 +15,9 @@
                 this,
                 '@previewUrl',
                 '@imageUrl',
-                 <i:raw>@allowedExtensions</i:raw>,
-                @singleFile
-        );
+                <i:raw>@allowedExtensions</i:raw>,
+                @singleFile,
+                {CSRFToken: '@part(sirius.web.http.CSRFHelper.class).getCSRFToken()'});
```
idlira: I think this part can be initialized without full package signature (or just in Tycho)?

idlira (approved):

sabieber (approved):

hso-scireum (approved):

src/main/resources/default/taglib/w/imageUpload.html.pasta
jakobvogel: You are right, and it is not Tycho-only: `SiriusClassAliasProvider` registers `CSRFHelper` as a global alias right here in sirius-web, so the short form resolves everywhere. The sibling taglibs `w:editForm`, `w:deleteLink` and `w:deleteButton` have always used it, which makes it the convention within this very directory.

Verified rather than assumed: with the short form the tests still assert the *actual* token value, so it resolves to the same part and not just to something that compiles. As a control I swapped in a bogus `NoSuchHelperXyz.class`, which fails the template compilation outright — so a non-resolving alias would not have slipped through silently.

Fixed in 4d7345e1. I also normalised the reference in `w:fileUpload`, which had carried the long form since it was written, so both upload taglibs read alike.

PR #1705, opened 2026-08-06
Extends WebContext session handling to enable hook-ins of server-side sessions

mkeckmkeck:
### Description
Adds an optional, application-supplied hook which permits storing the client session of selected
requests **on the server** instead of transporting it via the session cookie. Token authenticated
clients (mobile apps on a JSON API) have no cookie jar and currently lose all session state. The
framework only defines the seam - the actual store lives in the application, so browser and token
requests run side by side, each with its own session, without touching the ~60 call sites of
`setSessionValue`.

**`ServerSessionStorage`** (new, `sirius.web.http`) is consulted as an optional `@Part` (same shape
as `SessionSecretComputer`):

```java
boolean isResponsibleFor(WebContext webContext);            // once per request, lazily
Map<String, String> loadSession(WebContext webContext);      // empty map = fresh session
void persistSession(WebContext webContext, Map<String, String> session);  // empty map = delete
```

If a storage claims a request, `WebContext` neither reads nor writes the session cookie (nor the pin
cookie); all session accessors work unchanged on the loaded map. The cookie contract is mirrored:
only modified sessions are persisted at response time, cacheable responses discard changes, and the
response commit is never broken. Claiming and loading degrade instead of failing - a failed load
additionally suppresses persisting, so a transient error cannot overwrite the stored session.

Three smaller changes come with it:

* **`UserInfo.PERMISSION_TOKEN_AUTHENTICATED`** (following the existing `flag-logged-in` pattern) so
  a user manager can mark token-authenticated users.
* **CSRF skip for server session requests**: CSRF protects ambient cookie authority, which these
  requests do not have. Evaluated after the cheap method checks, so GETs do not force session init.
* **`TestRequest` uri handling** (test-jar): reading a parameter before `execute()` pinned the
  parameterless uri, hiding the parameters `build()` appends. Independent pre-existing bug which the
  new hook merely surfaced.

**No behavioural change without an implementation** - the state of every existing product. The test
storage is gated behind a framework flag disabled in the published test-jar.

### Additional Notes
- This PR fixes or works on following ticket(s): [SE-15422](https://scireum.myjetbrains.com/youtrack/issue/SE-15422)

### Checklist

### DIFF
--- a/src/main/java/sirius/web/http/ServerSessionStorage.java
+++ b/src/main/java/sirius/web/http/ServerSessionStorage.java
@@ -8,6 +8,8 @@
 
 package sirius.web.http;
 
+import sirius.kernel.di.std.AutoRegister;
+
 import java.util.Map;
 
 /**
@@ -25,6 +27,7 @@ import java.util.Map;
  * (which can carry a session id) instead of a session cookie. Implementations are registered via
  * {@link sirius.kernel.di.std.Register} against this interface.
  */
+@AutoRegister
 public interface ServerSessionStorage {
 
     /**
--- a/src/main/java/sirius/web/http/ServerSessionStorage.java
+++ b/src/main/java/sirius/web/http/ServerSessionStorage.java
@@ -1,0 +1,64 @@
+/*
+ * Made with all the love in the world
+ * by scireum in Stuttgart, Germany
+ *
+ * Copyright by scireum GmbH
+ * https://www.scireum.de - info@scireum.de
+ */
+
+package sirius.web.http;
+
+import java.util.Map;
+
+/**
+ * Permits an application to store the client session of selected requests on the server instead of transporting
+ * it via the session cookie.
+ * <p>
+ * If an implementation is registered and claims responsibility for a request (via
+ * {@link #isResponsibleFor(WebContext)}), the {@link WebContext} will neither read nor write the session cookie
+ * (nor the session pinning cookie) for that request. Instead, the session values are loaded once (lazily, on the
+ * first session access) and persisted back at response time if they were modified. If no implementation is
+ * registered, or it does not claim responsibility, the built-in cookie based client session remains fully in
+ * charge.
+ * <p>
+ * This is commonly used to provide a server side session for requests which are authenticated via a bearer token
+ * (which can carry a session id) instead of a session cookie. Implementations are registered via
+ * {@link sirius.kernel.di.std.Register} against this interface.
+ */
+public interface ServerSessionStorage {
+
+    /**
+     * Determines if this storage manages the session of the given request.
+     * <p>
+     * This is invoked at most once per request, lazily, when the session is first accessed. Implementations should
+     * be cheap for non-matching requests (e.g. check for the presence of a header first) and may cache expensive
+     * intermediate results (e.g. a validated token) as request attributes.
+     *
+     * @param webContext the request to check
+     * @return <tt>true</tt> if the session of this request is managed by this storage, <tt>false</tt> otherwise
+     */
+    boolean isResponsibleFor(WebContext webContext);
+
+    /**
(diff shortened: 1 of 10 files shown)


src/main/java/sirius/web/http/ServerSessionStorage.java
```diff
@@ -0,0 +1,64 @@
...
+ * (nor the session pinning cookie) for that request. Instead, the session values are loaded once (lazily, on the
+ * first session access) and persisted back at response time if they were modified. If no implementation is
+ * registered, or it does not claim responsibility, the built-in cookie based client session remains fully in
+ * charge.
+ * <p>
+ * This is commonly used to provide a server side session for requests which are authenticated via a bearer token
+ * (which can carry a session id) instead of a session cookie. Implementations are registered via
+ * {@link sirius.kernel.di.std.Register} against this interface.
+ */
+public interface ServerSessionStorage {
```
sabieber, line 31: **Optional:** `@AutoRegister`?

src/test/kotlin/sirius/web/http/WebContextTest.kt
```diff
@@ -165,6 +165,109 @@ class WebContextTest {
 
     }
 
+    private fun openConnection(path: String, sessionId: String? = null): HttpURLConnection {
```
sabieber: **Optional:** Put test harness boilerplate below the actual tests so it's immediately visible what gets tested.

src/test/kotlin/sirius/web/http/WebContextTest.kt
```diff
@@ -165,6 +165,109 @@ class WebContextTest {
...
+        // ...and the written value is NOT persisted, so a transient error cannot wipe the stored session.
+        assertEquals(persistCallsBefore, TestServerSessionStorage.getPersistCalls())
+        assertFalse { TestServerSessionStorage.hasStoredSession(TestServerSessionStorage.FAILING_SESSION_ID) }
+    }
+
+    @Test
+    fun `clearing a server session deletes the stored session`() {
+
+        // Seed a stored session...
+        val writeConnection = openConnection("/test/session-test", "to-clear")
```
sabieber: **Very optional:** Could utilize `.apply` in some of these tests to get nicer syntax

sabieber (approved):

src/main/java/sirius/web/http/ServerSessionStorage.java
mkeckmkeck: done

src/test/kotlin/sirius/web/http/WebContextTest.kt
mkeckmkeck: done

src/test/kotlin/sirius/web/http/WebContextTest.kt
mkeckmkeck: done

sabieber (approved):

src/test/kotlin/sirius/web/http/WebContextTest.kt
```diff
@@ -192,4 +344,10 @@ class WebContextTest {
         assertFalse { rewrittenCookie.contains("test1=test") }
 
     }
+
+    private fun openConnection(path: String, sessionId: String? = null): HttpURLConnection {
```
jmuscireum: Rename this method, I first couldn't find where the Header is set

fhaScireum (approved):

src/test/kotlin/sirius/web/http/WebContextTest.kt
mkeckmkeck: done

sabieber (approved):

jmuscireum (approved):

PR #1713, opened 2026-08-12
Add an `AllowedOrigin.Denied` for refusing CORS explicitly

jakobvogel:
### Description

Adds `AllowedOrigin.Denied`, so that an `Interceptor` can state that a route must stay unreachable cross-origin.

Since #1698 an interceptor can answer `determineAllowedCorsOrigin` in two ways: name a strategy, or return an empty optional to abstain. There is no way to say *"no origin at all"* — and abstaining is not that, because `ControllerDispatcher` falls back to an `AllowedOrigin.MatchRequest`, which reflects the origin of the request.

That is easy to get wrong, because the two look alike at the call site while the consequences are opposite. It bit us in scireum/memoio#2989: MEMOIO enables `http.enableCors` globally and grants a wildcard to its external API below `/api/`. Returning an empty optional for everything else — the obvious reading of "no CORS here" — silently made the whole backend UI, `/login/*` and `/csrf-token` reflect arbitrary origins. The only way to express the refusal today is an `AllowedOrigin.Specific` with an empty allow-list, which happens to resolve to no origin but states the intent nowhere and reads like an oversight to the next person.

`Denied` gives that case a name. It resolves to no origin, so no `Access-Control-Allow-Origin` header is emitted; the strategy is still recorded, so the `Vary` handling is unchanged, and `Access-Control-Allow-Credentials` is untouched (only `Specific` can ever request it). Also spells out in the javadoc of `determineAllowedCorsOrigin` that abstaining and refusing differ.

**Notes for review:**

- *Naming.* `Denied` over `None`, because the confusion this fixes is precisely between "no decision" and "no origin", and `None` reads like the former. Happy to rename if you prefer the symmetry with the other records.
- *Compatibility.* `AllowedOrigin` is `sealed`, so a downstream `switch` over it that was exhaustive stops compiling. Nothing in sirius-web is affected (the only switch is in `CorsAllowOriginResolver`), but it may influence how you want to version this.

### Additional Notes

- This PR is related to PR: #1698

### Checklist

### DIFF
--- a/src/main/java/sirius/web/controller/Interceptor.java
+++ b/src/main/java/sirius/web/controller/Interceptor.java
@@ -114,6 +114,12 @@ public interface Interceptor extends Priorized {
      * </p>
      *
      * <p>
+     * <b>Abstaining is therefore not the same as refusing.</b> As the fallback reflects the origin of the request,
+     * an empty optional leaves a route readable cross-origin. An interceptor which wants a route to stay unreachable
+     * for foreign origins must say so by returning an {@link AllowedOrigin.Denied}.
+     * </p>
+     *
+     * <p>
      * Please note that {@code enableCors} acts as a master switch: this method is only consulted when
      * {@link WebContext#isCorsEnabled()} returns {@code true} for the request's scope. If CORS handling is disabled,
      * no origin is resolved at all and this method is not invoked.
--- a/src/main/java/sirius/web/controller/Interceptor.java
+++ b/src/main/java/sirius/web/controller/Interceptor.java
@@ -115,7 +115,7 @@ public interface Interceptor extends Priorized {
      *
      * <p>
      * <b>Abstaining is therefore not the same as refusing.</b> As the fallback reflects the origin of the request,
-     * an empty optional leaves a route readable cross-origin. An interceptor which wants a route to stay unreachable
+     * an empty optional leaves a route readable cross-origin. An interceptor which wants a route to stay unreadable
      * for foreign origins must say so by returning an {@link AllowedOrigin.Denied}.
      * </p>
      *
--- a/src/main/java/sirius/web/cors/AllowedOrigin.java
+++ b/src/main/java/sirius/web/cors/AllowedOrigin.java
@@ -33,8 +33,13 @@ public sealed interface AllowedOrigin
      * This is <b>not</b> the same as an interceptor abstaining: an interceptor which returns an empty optional from
      * {@link Interceptor#determineAllowedCorsOrigin(WebContext, Collection)} states that it does not decide, and
      * {@link sirius.web.controller.ControllerDispatcher} then falls back to a {@link MatchRequest} - which reflects
-     * the origin of the request. An interceptor that wants a route to remain unreachable cross-origin therefore has
-     * to say so by returning this record.
+     * the origin of the request. An interceptor that wants the response of a route to stay unreadable cross-origin
+     * therefore has to say so by returning this record.
+     * </p>
+     * <p>
+     * Note that this withholds the <i>response</i> from the calling script rather than keeping the request away from
+     * the route: a simple request is still sent and still executed, and only a failing preflight stops one from
+     * arriving at all. Refusing an origin is therefore no substitute for CSRF protection.
      * </p>
      */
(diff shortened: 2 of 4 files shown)


src/main/java/sirius/web/controller/Interceptor.java
```diff
@@ -114,6 +114,12 @@ default int getPriority() {
      * </p>
      *
      * <p>
+     * <b>Abstaining is therefore not the same as refusing.</b> As the fallback reflects the origin of the request,
+     * an empty optional leaves a route readable cross-origin. An interceptor which wants a route to stay unreachable
+     * for foreign origins must say so by returning an {@link AllowedOrigin.Denied}.
```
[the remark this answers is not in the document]
jakobvogel: Correct, changed to "unreadable" in 72e3761a.

src/main/java/sirius/web/cors/AllowedOrigin.java
```diff
@@ -24,7 +24,21 @@
...
+
+    /**
+     * No origin is allowed at all, so no {@code Access-Control-Allow-Origin} header is set and the response stays
+     * unreadable for every foreign origin.
+     * <p>
+     * This is <b>not</b> the same as an interceptor abstaining: an interceptor which returns an empty optional from
+     * {@link Interceptor#determineAllowedCorsOrigin(WebContext, Collection)} states that it does not decide, and
+     * {@link sirius.web.controller.ControllerDispatcher} then falls back to a {@link MatchRequest} - which reflects
+     * the origin of the request. An interceptor that wants a route to remain unreachable cross-origin therefore has
+     * to say so by returning this record.
```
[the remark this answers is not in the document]
jakobvogel: Correct, changed in 3d47d9ad. Also added a sentence stating that a refused origin withholds the response rather than keeping the request away from the route — a simple request is still sent and executed, only a failing preflight stops one from arriving — so this is no substitute for CSRF protection. That is the conclusion "unreachable" invited, and it is worth ruling out explicitly on a record whose purpose is to separate two easily confused CORS concepts.

src/main/java/sirius/web/cors/AllowedOrigin.java
```diff
@@ -24,7 +24,26 @@
...
+     * the origin of the request. An interceptor that wants the response of a route to stay unreadable cross-origin
+     * therefore has to say so by returning this record.
+     * </p>
+     * <p>
+     * Note that this withholds the <i>response</i> from the calling script rather than keeping the request away from
+     * the route: a simple request is still sent and still executed, and only a failing preflight stops one from
+     * arriving at all. Refusing an origin is therefore no substitute for CSRF protection.
+     * </p>
+     */
+    record Denied() implements AllowedOrigin {
```
sabieber, line 52: **Question:** Wouldn't a `OwnOnly` be better/clearer, that only allows the products base URL as origin?

sabieber (approved):

idlira (approved):

src/main/java/sirius/web/cors/AllowedOrigin.java
jakobvogel: `OwnOnly` would emit `Access-Control-Allow-Origin: <baseUrl>`, but a request from that origin is same-origin, which the same-origin policy permits without any CORS header. It would therefore behave like `Denied` for foreign origins and do nothing for our own — a header that does no work.

It also cannot be derived reliably in the framework, because which origins make up "the product" is application knowledge. In MEMOIO, `product.baseUrl` carries no scheme and is thus not a valid `Origin` value, while the actual hosts live in `apiHost`, `mediaHost` and `channelHosts`. A value that does not match would emit a header matching nothing, which is indistinguishable from `Denied` until a legitimate cross-host call breaks.

The case where own-origin CORS is meaningful — one host of a multi-host deployment calling another — is genuinely cross-origin and already covered by `Specific`.

Added a paragraph to the `Denied` javadoc stating both points in c06f518e, since that is the question the record invites.

---
🤖 This reply was drafted by Claude (Opus 5) and reviewed before posting.

sabieber: If every browser behaves like that I am fine with it :)

PR #1729, opened 2026-09-09
Adds XLS and XLSX parity tests and fixes missing sheets in XLS imports

jakobvogel:
### Description

Follow-up to #1728. Bumping `excel-streaming-reader` to 5.3.0 exposed how thin the coverage of our Excel import path was: a single 3x3 fixture with strings, integers and formulas.

**Parity tests.** `XLSProcessorTest` asserts every expectation against both formats, because `XLSProcessor` and `XLSXProcessor` share their cell extraction but obtain the `Workbook` from entirely different libraries - POI's `HSSFWorkbook` for XLS, the streaming reader for XLSX. A parity suite is what keeps the two backends from drifting apart, across future library bumps as well as across changes to the extraction itself.

Newly pinned: date and date-time cells, boolean cells, error cells (`1/0` reads as `null`), the integer/double normalization, whitespace trimming, interior gaps versus trailing empty cells, a whitespace-only row, a row without a single cell, column anchoring, the `errorHandler` contract in both directions, `importAllSheets`, and `runForSheets` including its missing-sheet report - which had no test whatsoever.

Two behaviours worth knowing, both now pinned:

- Line numbers count **processed** rows, not spreadsheet rows: the fixture's row in Excel row 11 arrives as line 9, because rows 9 and 10 hold no cells. Importer messages can therefore name a different row than the user sees in Excel.
- Every row is anchored at column A rather than at its own first cell, so a row starting in column C arrives with two leading `null`s instead of shifting left.

**Bug fix.** `runForSheets` only handled the way the streaming reader reports an unknown sheet (`MissingSheetException`, caught in order to log and continue). `HSSFWorkbook` returns `null` instead, so the same import of an **XLS** file died with a `NullPointerException` and took the sheets behind it down with it. Reachable in production: scireum-core's FabDis import hands the same sheet processors to either processor, depending only on the extension of the uploaded file.

### Known gap, deliberately not fixed here

An absent sheet is reported to the `TaskContext` as an *info* message and nothing marks the import erroneous. For a **mandatory** sheet that is worse than it sounds: in FabDis, `importTransactionHelper.start(...)` runs inside the `00_CARTOUCHE` row processor, and `close()` gates the assortment delete/mark pass on `importTransactionHelper.isActive()`. So a renamed Cartouche sheet now yields a *successful* import that silently skipped both the cartouche and the entire sync-source pass. That hazard already existed on the XLSX path; this PR extends it to XLS, where the NPE used to make the problem visible by accident.

Telling a mandatory sheet from an optional one needs something `SheetBasedRowProcessor` does not have: a way to declare itself required. Adding it touches this record's API and needs a decision in scireum-core, so it belongs in its own change rather than riding along here. The limitation is now stated in the `runForSheets` Javadoc. **Happy to raise a ticket for it.**

Also worth a ticket, unrelated to this change: because we trust `getCachedFormulaResultType()`, a `.xls` file whose cached formula results were dropped by the writing application imports text- and boolean-valued formula cells as `0`/`1`.

### Where the fixtures come from

POI writes the workbooks when a test runs, as `ExcelExportTest` already does - no binaries and no tooling, so the contents being asserted sit next to the assertions and adding a cell is an edit rather than a regeneration.

An earlier draft committed the workbooks as binaries, passed the XLSX ones through LibreOffice for a more realistic serialization and carried a generator to reproduce that pipeline. Measuring it deflated the idea: the only structural difference to POI's own output is `xl/theme/theme1.xml`, which no reader touches - POI's XSSF already writes shared strings, styles and doc properties. Meanwhile the conversion turned boolean literals into `TRUE()`/`FALSE()` formulas, normalized the empty-string cell to a blank and dropped the cell-less row, so it cost three workarounds and found no reader defect.

Coverage of files as a real application writes them stays where it already was: `LineBasedProcessorTest` reads `test.xlsx`, which was saved by Excel.

### Verification

- Full suite locally: 451 tests, 0 failures. `XLSProcessorTest` contributes 32.
- Date assertions verified timezone-independent under UTC, Europe/Berlin, Pacific/Auckland, America/Santiago and Asia/Kolkata; the date test pins the zone itself, since on a UTC JVM the conversion via `java.util.Date` cancels out and the assertion would otherwise hold even if the extraction stopped honouring the default zone.
- Reviewed adversarially from three angles, and the resulting findings are worked in. Mutation-verified in a throwaway worktree that the tests actually catch: hardcoding the date zone to UTC, anchoring rows at their own first cell, ignoring the `errorHandler`, dropping either of the two missing-sheet reports, and skipping the trailing-empty-cell trim. The first five all survived the initial draft of the suite.

### Additional Notes

- This PR fixes or works on following ticket(s): none - follow-up to #1728
- This PR is related to PR: #1728

### Checklist

- [x] Code was formatted via IntelliJ and follows [best practices](https://scireum.myjetbrains.com/youtrack/articles/MISC-A-16/CodeStyle-JavaDoc) - **SonarLint was not run**, as it cannot be executed headlessly; please run *Analyze VCS Changed Files* in the IDE
- [x] Patch Tasks: not necessary

### DIFF
--- a/src/main/java/sirius/web/data/XLSProcessor.java
+++ b/src/main/java/sirius/web/data/XLSProcessor.java
@@ -67,28 +67,51 @@ public class XLSProcessor extends LineBasedProcessor {
     }
 
     /**
-     * Processes the XLS (MS Excel) file using the given {@link SheetBasedRowProcessor sheetProcessors}.
+     * Processes the workbook using the given {@link SheetBasedRowProcessor sheetProcessors}.
+     * <p>
+     * Note that {@link XLSXProcessor} inherits this, so the workbook is an XLS or an XLSX file.
+     * <p>
+     * A sheet which is not present in the workbook is reported to the {@link TaskContext} as an info message
+     * and skipped, and the sheet processors behind it are still run. Note that this is all a caller learns
+     * about an absent sheet: the import is not marked as erroneous, as a sheet processor cannot state whether
+     * its sheet is mandatory.
+     * <p>
+     * A row which fails and whose error handler declines to handle it still aborts the whole run, and the
+     * {@code importAllSheets} flag given to the constructor has no effect here, as the sheets to process are
+     * named by the given processors.
      *
      * @param sheetProcessors one or more sheet processors to use
-     * @throws Exception                             in case an error occurred while processing.
-     * @throws IOException                           if the stream providing the given workbook cannot be read.
-     * @throws sirius.kernel.health.HandledException if workbook does not contain a sheet with the given name.
+     * @throws Exception   in case an error occurred while processing.
+     * @throws IOException if the stream providing the given workbook cannot be read.
      */
     public void runForSheets(SheetBasedRowProcessor... sheetProcessors) throws Exception {
         try (Workbook workbook = openWorkbook()) {
             for (SheetBasedRowProcessor processor : sheetProcessors) {
                 try {
                     Sheet sheet = workbook.getSheet(processor.sheetName());
+                    if (sheet == null) {
+                        // HSSFWorkbook reports an unknown sheet by returning null, whereas the streaming
+                        // reader used for XLSX raises a MissingSheetException
+                        reportMissingSheet(processor.sheetName());
+                        continue;
+                    }
                     importSheet(processor.rowProcessor(), processor.errorHandler(), sheet);
                 } catch (MissingSheetException missingSheetException) {
-                    TaskContext.get()
-                               .log(NLS.fmtr("XLSProcessor.info.missingSheet")
-                                       .set("sheet", processor.sheetName())
-                                       .format());
+                    reportMissingSheet(processor.sheetName());
                     Exceptions.ignore(missingSheetException);
                 }
             }
         }
+    }
+
+    /**
+     * Reports a sheet requested via {@link #runForSheets(SheetBasedRowProcessor...)} which the workbook does
+     * not contain.
+     *
+     * @param sheetName the name of the absent sheet
+     */
+    private void reportMissingSheet(String sheetName) {
+        TaskContext.get().log(NLS.fmtr("XLSProcessor.info.missingSheet").set("sheet", sheetName).format());
     }
 
     protected Workbook openWorkbook() throws IOException {
--- a/src/test/kotlin/sirius/web/data/XLSProcessorTest.kt
+++ b/src/test/kotlin/sirius/web/data/XLSProcessorTest.kt
@@ -20,6 +20,8 @@ import org.junit.jupiter.api.extension.ExtendWith
 import org.junit.jupiter.params.ParameterizedTest
 import org.junit.jupiter.params.provider.ValueSource
 import sirius.kernel.SiriusExtension
+import sirius.kernel.async.BasicTaskContextAdapter
+import sirius.kernel.async.TaskContext
 import sirius.kernel.commons.Values
 import java.io.ByteArrayInputStream
 import java.io.ByteArrayOutputStream
@@ -30,6 +32,7 @@ import java.util.TimeZone
 import java.util.function.Predicate
 import kotlin.test.assertEquals
 import kotlin.test.assertFailsWith
+import kotlin.test.assertTrue
 
 /**
  * Verifies that [XLSProcessor] and [XLSXProcessor] read the same file contents in the same way.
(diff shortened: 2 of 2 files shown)


src/test/kotlin/sirius/web/data/XLSProcessorTest.kt
```diff
@@ -0,0 +1,400 @@
...
+
+    private fun process(
+        format: String,
+        fill: (Workbook) -> Unit,
+        importAllSheets: Boolean,
+        errorHandler: (Exception) -> Boolean,
+        rowProcessor: (Int, Values) -> Unit
+    ) {
+        LineBasedProcessor.create("fixture.$format", workbookStream(format, fill), importAllSheets)
+            .run({ lineNumber, row -> rowProcessor(lineNumber, row) }, errorHandler)
```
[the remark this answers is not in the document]
jakobvogel: This one is a false positive - the code does compile, and this very commit proves it.

SAM conversion being limited to lambda literals was true up to Kotlin 1.3; Kotlin 1.4 extended it to function-typed values as part of the new inference. This project builds with Kotlin 2.4.10, so passing the `(Exception) -> Boolean` parameter where the Java signature wants `Predicate<Exception>` converts fine.

The evidence is the pipeline on the reviewed commit `d6dab1f4`: Drone runs `mvn clean test`, which reported 451 tests passing, 32 of them from this file. A file that did not compile would have produced no test run at all. Locally the same, under JDK 26.

Leaving the code as it is. Flagging it here so that nobody wraps it in a `Predicate` to satisfy the suggestion.

(Analysis by Claude Opus 5 on behalf of @jakobvogel.)

src/main/java/sirius/web/data/XLSProcessor.java
```diff
@@ -68,29 +68,50 @@ public void run(RowProcessor rowProcessor, Predicate<Exception> errorHandler) th
 
     /**
      * Processes the XLS (MS Excel) file using the given {@link SheetBasedRowProcessor sheetProcessors}.
```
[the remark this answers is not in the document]
jakobvogel: Fair point, fixed in `42aa294f`. The first line now reads "Processes the workbook using the given sheetProcessors", followed by a note that `XLSXProcessor` inherits the method, so the workbook may be an XLS or an XLSX file - which is exactly how the only production caller uses it, picking the processor by file extension and handing the same sheet processors to either.

The wording was inherited from before this PR, but since the block was being rewritten anyway it may as well be accurate.

(Analysis by Claude Opus 5 on behalf of @jakobvogel.)

src/test/kotlin/sirius/web/data/XLSProcessorTest.kt
```diff
@@ -0,0 +1,400 @@
+/*
+ * Made with all the love in the world
+ * by scireum in Remshalden, Germany
```
ObaiAlbek-dev: soll hier nicht stuttgart stehen? 
/*
 * Made with all the love in the world
 * by scireum in Stuttgart, Germany
 *
 * Copyright by scireum GmbH
 * https://www.scireum.de - info@scireum.de
 */

jakobvogel: Stimmt, danke - in `9fb3667d` korrigiert, inklusive `https` statt `http`.

Der Header war aus dem benachbarten `XLSProcessor` übernommen, und der nennt noch Remshalden. Im Repository stehen 439 Dateien mit dem alten Header und 23 mit dem aktuellen - die neu hinzugefügten Dateien nutzen aber durchweg Stuttgart über `https`, insofern war das schlicht die falsche Vorlage.

Geändert ist nur die Datei, die dieser Pull Request neu anlegt. Die übrigen Dateien mit dem alten Header wären eine eigene Migration und haben in einem Pull Request über Excel-Importe nichts zu suchen - falls das gewünscht ist, gerne separat.

(Bearbeitet von Claude Opus 5 im Auftrag von @jakobvogel.)

ObaiAlbek-dev (approved):

src/test/kotlin/sirius/web/data/XLSProcessorTest.kt
jakobvogel: mensch, claude hat heute wieder nen aggro ton drauf 😅 – PR mit Anpassung in den anderen Dateien kommt

jakobvogel: #1731

mkeckmkeck (approved):

idlira (approved):

PR #1611, opened 2026-04-23
Add tunnel backpressure handling

mkeckmkeck:
### Description
Add tunnel backpressure handling: 
- to fix the problem that our tunnel handling uses too much memory when serving slowly reading clients from fast sources
- this is reached by mirroring the client writability to upstream channel autoRead
- also adds some regression tests for bounded buffering and keep-alive cleanup, note: "Back-pressure keeps tunnel outbound buffering bounded for a stalled consumer" actually proves that the backpressure feature is working. the other tests ensure general tunnel reading work (still) properly and internal backpressure-netty validity

### Additional Notes
- This PR fixes or works on following ticket(s): [SE-15103](https://scireum.myjetbrains.com/youtrack/issue/SE-15103)

### Checklist

- [x] Code change has been tested and works locally
- [x] Code was formatted via IntelliJ and follows SonarLint & [best practices](https://scireum.myjetbrains.com/youtrack/articles/MISC-A-16/CodeStyle-JavaDoc)
- [ ] Patch Tasks: Is local execution of Patch Tasks necessary? If so, please also mark the PR with the tag.

PR #1730, opened 2026-09-09
Prevents tunnel back-pressure from stranding pooled connections

idlira:
### Description

Back-pressure for tunnelled responses (SE-15103, #1611) slows the upstream down by disabling `autoRead` on its channel. That channel belongs to AsyncHttpClient's connection pool and therefore outlives the request which paused it, while `mirrorWritabilityToUpstream` only verified that the channel was still open — not that it still belonged to the request. A pause could consequently survive on a connection that was already back in the pool, and such a connection can never answer: the code which would re-apply the current writability is only reached from `onHeadersReceived`, and no headers can arrive while reading is disabled. The request hangs until AsyncHttpClient's read timeout closes the connection, which is one minute by default.

Ownership of an upstream connection is now explicit, so a channel that gets replaced by a retry or given up at the end of a request is always left readable. The ownership check happens at the moment the change is applied rather than when it is issued, because pausing arrives on the client's event loop while releasing arrives on the upstream's — an already queued pause could otherwise still take effect afterwards. `onConnectionPooled` additionally restores reading on a connection that arrives paused and logs a warning; that is a deliberate second line of defence, and the warning makes any remaining gap visible.

The final commit is a framework enhancement rather than part of the fix. `http.tunnel.backpressure.minResponseSize` keeps small responses out of the mechanism altogether — they fit into a buffer or two, so pausing the upstream gains nothing for them — while responses without a `Content-Length` continue to get back-pressure, as those are the ones which can grow without bound. `http.tunnel.backpressure.enabled` exists so the mechanism can be ruled in or out on an installation without a custom build.

Two things worth knowing for review:

- The first commit is intentionally red. It adds the tests before the fix, so anyone bisecting through this range will see three failures there.
- The pause-versus-release ordering across the two event loops is not covered by a test. That interleaving cannot be provoked from a test, and the commit body states so rather than implying full coverage.

### Additional Notes

- This PR fixes or works on following ticket(s): [SIRI-1271](https://scireum.myjetbrains.com/youtrack/issue/SIRI-1271)
- This PR is related to PR: https://github.com/scireum/sirius-web/pull/1611

### Checklist

🤖 Generated with Claude Opus 5 (1M context) via [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/main/java/sirius/web/http/TunnelHandler.java
+++ b/src/main/java/sirius/web/http/TunnelHandler.java
@@ -99,6 +99,7 @@ public class TunnelHandler implements AsyncHandler<String> {
              + "required cross-thread visibility.")
     private volatile Channel upstreamChannel;
 
+
     private volatile boolean failed;
 
     TunnelHandler(Response response,
@@ -129,7 +130,7 @@ public class TunnelHandler implements AsyncHandler<String> {
     @Override
     public void onTcpConnectSuccess(InetSocketAddress remoteAddress, Channel connection) {
         this.timeToConnect = watch.elapsedMillis();
-        this.upstreamChannel = connection;
+        adoptUpstream(connection);
     }
 
     @Override
@@ -137,7 +138,21 @@ public class TunnelHandler implements AsyncHandler<String> {
         // Pooled connections never trigger onTcpConnectSuccess, but we still need a reference
         // to the upstream channel to apply back-pressure (toggling autoRead) once the response
         // headers have been received.
-        this.upstreamChannel = connection;
+        adoptUpstream(connection);
+
+        // A pause applied by a previous user of this connection can outlive that request:
+        // mirrorWritabilityToUpstream only verifies that the channel is open, and a connection
+        // sitting in the pool is open. Such a connection can never deliver a response, because
+        // installBackpressureBridge - which is what would re-apply the current writability - is
+        // only reached from onHeadersReceived, and no headers can arrive while reading is off.
+        // Reading is therefore restored here, at the one point where a pooled connection is handed
+        // to a new request.
+        if (!connection.config().isAutoRead()) {
+            WebServer.LOG.WARN("Tunnel: re-enabled reading on a pooled upstream connection which was"
+                               + " left paused by a previous request. Target: %s",
+                               webContext.getRequestedURI());
+            mirrorWritabilityToUpstream(true);
+        }
     }
 
     @Override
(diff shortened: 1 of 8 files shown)


src/test/java/sirius/web/http/TunnelPoolProbe.java
```diff
@@ -0,0 +1,306 @@
...
+     * and would otherwise leak into unrelated tests.
+     */
+    public static void restore() {
+        Response.asyncClient = previousClient;
+        previousClient = null;
+
+        if (probeClient != null) {
+            try {
+                probeClient.close();
+            } catch (Exception _) {
```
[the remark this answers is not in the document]
idlira: BS. Wir nutzen doch Java 26 😤

src/test/java/sirius/web/http/TunnelPoolProbe.java
```diff
@@ -0,0 +1,306 @@
+/*
```
ObaiAlbek-dev, line 1: sollte hier nicht die neue header stehen
/*
 * Made with all the love in the world
 * by scireum in Stuttgart, Germany
 *
 * Copyright by scireum GmbH
 * https://www.scireum.de - info@scireum.de
 */

src/main/java/sirius/web/http/TunnelHandler.java
```diff
@@ -99,6 +100,7 @@ public class TunnelHandler implements AsyncHandler<String> {
              + "required cross-thread visibility.")
     private volatile Channel upstreamChannel;
 
+
```
jakobvogel: Format, please 😇

jakobvogel (approved):

src/test/java/sirius/web/http/TunnelPoolProbe.java
idlira: Hier ist es schon in andere commit, aber andere Klasse hats noch gefailed.

mkeckmkeck (approved):

src/test/java/sirius/web/http/TunnelHandlerOwnershipTest.java
```diff
@@ -0,0 +1,83 @@
+/*
+ * Made with all the love in the world
+ * by scireum in Stuttgart, Germany
+ *
+ * Copyright by scireum GmbH
+ * http://www.scireum.de - info@scireum.de
```
jakobvogel:
```suggestion
 * https://www.scireum.de - info@scireum.de
```

src/test/java/sirius/web/http/TunnelPoolProbe.java
```diff
@@ -0,0 +1,306 @@
...
+ * Made with all the love in the world
+ * by scireum in Stuttgart, Germany
+ *
+ * Copyright by scireum GmbH
+ * http://www.scireum.de - info@scireum.de
```
jakobvogel:
```suggestion
 * https://www.scireum.de - info@scireum.de
```

src/test/kotlin/sirius/web/http/TunnelUpstreamPoolTest.kt
```diff
@@ -0,0 +1,206 @@
+/*
+ * Made with all the love in the world
+ * by scireum in Stuttgart, Germany
+ *
+ * Copyright by scireum GmbH
+ * http://www.scireum.de - info@scireum.de
```
jakobvogel:
```suggestion
 * https://www.scireum.de - info@scireum.de
```

PR #1735, opened 2026-09-11
Exposes the OAuth token expiry as an instant

jakobvogel:
### Description

> [!WARNING]
> **Breaking:** both expiry fields of `ReceivedTokens` change from `LocalDateTime` to `Instant`. Consumers storing them in a date column or rendering them have to convert explicitly.

Two fixes in `ReceivedTokens.fromJson`, found while writing the tests in #1734:

1. **Crash on JWT refresh tokens without an `exp` claim.** The claim is optional and `getExpiresAtAsInstant()` returns `null`, so the conversion threw a `NullPointerException`. Being no `JWTDecodeException`, it escaped the surrounding `catch` and aborted the token exchange instead of falling back to the usual estimate.

2. **Expiries are instants now.** An expiry is a point in time; turning it into a `LocalDateTime` discards the offset and leaves consumers guessing which zone to read it in. The JWT claim is an `Instant` to begin with, so the conversion only ever existed to satisfy the type of this record. It was also inconsistent: the JWT expiry used UTC while the other date came from the system clock, so a refresh token could look expired hours early or stay valid hours too long, depending on the host.

An earlier revision normalised everything to the system zone instead. That only moved the ambiguity and has been dropped.

### Migration

```java
// before
credentials.setRefreshTokenExpiresAt(tokens.refreshTokenExpiresAt());

// after
credentials.setRefreshTokenExpiresAt(tokens.refreshTokenExpiresAt() == null ?
                                     null :
                                     LocalDateTime.ofInstant(tokens.refreshTokenExpiresAt(), ZoneId.systemDefault()));
```

`ZoneId.systemDefault()` reproduces the previous behaviour. Existing rows need no migration, they are overwritten on the next refresh. Nothing inside sirius-web uses these fields.

### Additional Notes

- This PR fixes or works on following ticket(s): none, found while reviewing #1733.
- This PR is related to PR: #1734, already merged.

### Checklist

- [x] Patch Tasks: Is local execution of Patch Tasks necessary? If so, please also mark the PR with the tag. — not necessary

IntelliJ formatter reports *Formatted well*; SonarLint cannot run headlessly, so please run *Analyze VCS Changed Files* before merging.

### Verification

Both fixes have regression tests confirmed to fail without them. The suite is green under `TZ=UTC`, `Europe/Berlin`, `America/Anchorage` and `Pacific/Kiritimati`, with one test pinning the result against a forced non-UTC default zone. Fixtures use fixed instants, so they cannot flake on a DST transition.

### DIFF
--- a/src/main/java/sirius/web/security/oauth/ReceivedTokens.java
+++ b/src/main/java/sirius/web/security/oauth/ReceivedTokens.java
@@ -13,6 +13,7 @@ import com.auth0.jwt.exceptions.JWTDecodeException;
 import tools.jackson.databind.node.ObjectNode;
 import sirius.kernel.commons.Strings;
 
+import java.time.Instant;
 import java.time.LocalDateTime;
 import java.time.ZoneOffset;
 
@@ -56,9 +57,15 @@ public record ReceivedTokens(String accessToken, String refreshToken, String typ
         if (OAuth.TOKEN_TYPE_BEARER.equalsIgnoreCase(type)) {
             try {
                 // Try to read the exact refresh token expiration date from the JWT token itself
-                LocalDateTime refreshTokenExpiresAt =
-                        JWT.decode(refreshToken).getExpiresAtAsInstant().atZone(ZoneOffset.UTC).toLocalDateTime();
-                return new ReceivedTokens(accessToken, refreshToken, type, accessTokenExpiresAt, refreshTokenExpiresAt);
+                Instant expiry = JWT.decode(refreshToken).getExpiresAtAsInstant();
+                if (expiry != null) {
+                    return new ReceivedTokens(accessToken,
+                                              refreshToken,
+                                              type,
+                                              accessTokenExpiresAt,
+                                              expiry.atZone(ZoneOffset.UTC).toLocalDateTime());
+                }
+                // The JWT carries no expiration date, fall back to the estimates below
             } catch (JWTDecodeException _) {
                 // No valid JWT, fall back to implementation from OAuth expires_in or the default value
             }
--- a/src/main/java/sirius/web/security/oauth/ReceivedTokens.java
+++ b/src/main/java/sirius/web/security/oauth/ReceivedTokens.java
@@ -14,8 +14,7 @@ import tools.jackson.databind.node.ObjectNode;
 import sirius.kernel.commons.Strings;
 
 import java.time.Instant;
-import java.time.LocalDateTime;
-import java.time.ZoneOffset;
+import java.time.temporal.ChronoUnit;
 
 /**
  * Represents the tokens received from an OAuth authorization server. This is used to read the token in a standard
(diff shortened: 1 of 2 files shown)


src/main/java/sirius/web/security/oauth/ReceivedTokens.java
```diff
@@ -56,9 +60,15 @@ public static ReceivedTokens fromJson(ObjectNode response) {
...
-                LocalDateTime refreshTokenExpiresAt =
-                        JWT.decode(refreshToken).getExpiresAtAsInstant().atZone(ZoneOffset.UTC).toLocalDateTime();
-                return new ReceivedTokens(accessToken, refreshToken, type, accessTokenExpiresAt, refreshTokenExpiresAt);
+                Instant expiry = JWT.decode(refreshToken).getExpiresAtAsInstant();
+                if (expiry != null) {
+                    return new ReceivedTokens(accessToken,
+                                              refreshToken,
+                                              type,
+                                              accessTokenExpiresAt,
+                                              LocalDateTime.ofInstant(expiry, ZoneId.systemDefault()));
```
jakobvogel: This line is a stopgap and I would like your opinion on the proper fix.

The `exp` claim of a JWT is an **instant**. We turn it into a `LocalDateTime`, i.e. we drop the offset and keep a bare wall clock. Until now that conversion used UTC, while every other date in this record comes from `LocalDateTime.now()`, i.e. the system zone — so two fields of the same record were expressed in different zones, and `prefersLongAccessTokenExpiryOverTheDefaultOfTomorrow` even returns the same value for both. Switching to `ZoneId.systemDefault()` makes them consistent, which is all this line does.

Why it matters in practice — the consumers live outside this repo and all compare against `LocalDateTime.now()`:

- **sellsite**, `customizations/qpunchout`: `QPunchoutCustomerTraderPartnership` persists both values in `LocalDateTime` columns, `QPunchoutOAuth` decides whether to refresh via `getRefreshTokenExpiresAt().isBefore(LocalDateTime.now())` and resets the tokens when it is past, `QPunchoutTraderAuthRefresh` queries `lte(REFRESH_TOKEN_EXPIRES_AT, now+7d)`, and several `.pasta` templates render the value to admins.
- **scireum-core**: `DataBaseOAuthTokenProvider.isNotYetExpired` uses `isAfter(LocalDateTime.now())`, `OAuthWebUtils` persists both.

With the old UTC conversion, on a host east of UTC the refresh token looked expired up to two hours early, and west of UTC it looked valid for hours after it had actually expired. If all our hosts run UTC this was latent rather than active — but it is a trap, and the admin-facing display was wrong regardless.

What I think we should really do, and deliberately did **not** do here: carry the value as an `Instant` (or an `OffsetDateTime`/`ZonedDateTime`) in this record and let each consumer decide how to compare or render it. A point in time should not be stored zone-less. I left it out because it changes the signature of a public record and therefore the sellsite and scireum-core entities and their DB columns, so it wants a coordinated change across the repos rather than a drive-by in a test PR.

Worth a ticket? Happy to take it if you agree on the direction.

idlira (approved):

PR #1743, opened 2026-09-22
Releases the tunnel upstream before it enters the pool

idlira:
### Description

`TunnelHandler` applies back-pressure by disabling `autoRead` on the upstream connection and restores it when the request ends. That restore happens in `onCompleted()`, which is too late: AsyncHttpClient offers the connection to its pool *before* it reports the request as completed (`AsyncHttpClientHandler.finishUpdate` calls `tryToOfferChannelToPool(...)` and only then `future.done()`). Between those two calls the connection is visible in the pool while still paused, and a request that polls it there has to repair it before it can be used. That is not an edge case — it is what happens whenever a tunnelled response completes while back-pressure is engaged, which means a slow consumer and a response large enough to have engaged it.

This overrides `AsyncHandler.onConnectionOffer(Channel)`, which AsyncHttpClient invokes synchronously inside `tryToOfferChannelToPool` immediately before `channelPool.offer(...)` — on the connection's own event loop, and before it becomes visible to anyone else. Releasing the connection there closes the window rather than repairing it afterwards. Checked on both code paths: the callback only fires once the exchange has finished, HTTP/2 included, so it cannot disarm back-pressure mid-response.

The repair in `onConnectionPooled` stays. Correctness no longer depends on which of the two happens first, and with the window closed its warning goes back to reporting an anomaly instead of routine operation.

The first commit is unrelated to the behaviour change and only concerns the test harness. `TunnelUpstreamPoolTest` pins the pool to a single connection, and AsyncHttpClient releases the per-host slot on the upstream event loop — possibly after the response the test thread has already read. A request started immediately afterwards gets no slot and is aborted once the connect timeout expires, which made the test fail under CI load but never locally. Both tests now wait for the connection to be back in the pool instead of for a fixed duration, and the fault injection checks that the channel is still pooled when it fires.

No behavioural change for callers, no configuration change, nothing breaking.

### Additional Notes

- This PR fixes or works on following ticket(s): [SIRI-1276](https://scireum.myjetbrains.com/youtrack/issue/SIRI-1276)
- This PR is related to PR: https://github.com/scireum/sirius-web/pull/1730

### Checklist

---

`TunnelHandlerOwnershipTest.pausedConnectionIsRestoredBeforeItIsOfferedToThePool` covers the contract and is red without the change, with the two existing ownership tests staying green. A test driving this through a real request is not possible, for the reason already recorded in that class's flowerbox: a response small enough to complete in one burst disappears into the loopback socket buffers and makes the consumer writable again, while one large enough to keep the consumer blocked pauses the upstream before it finishes, so its connection never reaches the pool. `TunnelUpstreamPoolTest` already asserts `pausedOnOffer()` and passes either way for that reason.

Verified that each commit is green on its own: 4 tests at the first, 5 at the tip.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/test/java/sirius/web/http/TunnelPoolProbe.java
```diff
@@ -143,6 +144,55 @@ public static void poisonPooledConnections(boolean poison) {
...
+    private static boolean await(Duration timeout, Predicate<Channel> predicate) {
+        long deadline = System.currentTimeMillis() + timeout.toMillis();
+        while (System.currentTimeMillis() < deadline) {
+            if (pool != null && pool.hasPooledConnection(predicate)) {
+                return true;
+            }
+            Wait.millis(POLL_INTERVAL_MILLIS);
+        }
+        return pool != null && pool.hasPooledConnection(predicate);
+    }
```
sabieber, line 194: **Question:** This is code that it only used in the tests but lives in the production code? I don't really like that. Would prefer an override of the class in the tests package and moving the methods there or something like that.

sabieber (approved):

src/test/java/sirius/web/http/TunnelPoolProbe.java
idlira: Nope. TestTunnelProbe lives in src/**test**/java...

mkeckmkeck (approved):

